PR #4, opened 2018-08-15
Feature/loading indicator

HendrikThePendric:
Please review my choice of props as well as my code.


src/packages/CircularProgress/CircularProgress.js
```diff
@@ -0,0 +1,55 @@
...
+import PropTypes from 'prop-types';
+import './circular-progress.css';
+
+const CircularProgress = ({ color, size, strokeWidth, overlay, overlayFill }) => {
+    const wrapperStyle = { width: size, height: size };
+    const spinnerLayerStyle = { borderColor: color };
+    const circleStyle = { borderWidth: strokeWidth };
+
+    const circularProgress = (
+        <div className="d2ui-circular-progress" style={wrapperStyle}>
```
varl: Avoid `style`.

src/packages/CircularProgress/CircularProgress.js
```diff
@@ -0,0 +1,55 @@
...
+                )}
+                {circularProgress}
+            </div>
+        );
+    }
+
+    return circularProgress;
+};
+
+CircularProgress.propTypes = {
```
varl: I do not think we need the ability to customize `color`, `size`, `strokeWidth`, `overlay`, and `overlayFill` separately.

We can instead use two different classes, one for large spinners and one for small spinners.

Large spinners use the overlay to block out big chunks of content (e.g. a Table loading data, or the entire app).

A small spinner is used for inline spinning needs, e.g. inside a "save" button after it's been clicked.

src/packages/LinearProgress/LinearProgress.js
```diff
@@ -0,0 +1,45 @@
...
+        <div
+            className={classNames('d2ui-linear-progress', { minimal: minimal })}
+            style={outerStyle}
+        >
+            <div className={innerClass} style={innerStyle} />
+        </div>
+    );
+};
+
+LinearProgress.propTypes = {
```
varl: I don't think we need `progress`, `backgroundColor`, `progressColor`, `height`, or `minimal`. We can just have a default indeterminate linear loading indicator.

If we need a separate look and feel for the loading indicators I would rather enable that by different theme classes than let the developer override it manually.

HendrikThePendric:
Thanks for the input. In the case of progress indicators, I hope we will not be rendering that many on the page at once, so performance is not really an issue I think. But anyway, I agree with making things even simpler, and this will enable me to get rid of those inline styles....

src/packages/CircularProgress/CircularProgress.js
HendrikThePendric: I agree but would suggest we keep some ability to either include or omit a backdrop. Not offer an option to style it, but do offer an option toggle it on or off.

src/packages/LinearProgress/LinearProgress.js
HendrikThePendric: So just to check, I will end up with these props:
- progress (float 0-1)
- padded (bool) (i.e. minimal is default and by adding padding you get the margin top/bottom)

src/packages/CircularProgress/CircularProgress.js
varl: When would you want to overlay without a backdrop?

HendrikThePendric: > When would you want to overlay without a backdrop?
Some examples:
Without backdrop: for example when a form is loading and we are just showing an empty space, I would want to show the overlay spinner which is centered nicely without a backdrop
With backdrop: When a full UI is already on the page but for whatever reason you want to block a section or the whole page....

src/packages/LinearProgress/LinearProgress.js
HendrikThePendric: I wasn't sure. In my personal experience progress is often a result of doing something like:
```javascript
progress = finishedItems / totalItems;
```
Which then makes it mildly easier to work with a 0-1 range

HendrikThePendric: Actually, forget about this. I will change to 0-100 to be more consistent with other libraries (i.e. Material UI, etc) out there.

varl: >Which then makes it mildly easier to work with a 0-1 range

- The way it's implemented now that means that we just convert the float to a decimal percentage, so we can attempt to render `33.333333333335%`. We should at least trim it to `33%`.

- Without looking at the docs, I would assume that progress is measured in percentages. I find it more intuitive.

src/packages/CircularProgress/CircularProgress.js
varl: >Without backdrop: for example when a form is loading and we are just showing an empty space, I would want to show the overlay spinner which is centered nicely without a backdrop

If we are showing an empty space it's the same with or without the backdrop, given that the backdrop color is: `rgba(255,255,255,0.7)`.

HendrikThePendric: After discussion on Slack, we decided to completely ditch the overlay and backdrop. The current components only render a ProgressIndicator. If we want them in a modal with or without backdrop this should be done via a Modal component.

PR #7, opened 2018-08-20
Implement bem methodology

HendrikThePendric:
When reviewing, I think the most important file would be `src/utils/bemClassNames.js` which contains the actual implementation of the `bemClassNames` utility function. The idea is like this:
- `bemClassNames` is a closure. You call it outside of the component with the block/component name and store the returned function in a variable (called bem for convention).
- Then, in the component call the returned function with as many parameters as you like.
- The first parameter is used as an `element` (as in block__element). This is also why you must call `bem(null, 'modifier');` if you want to add a modifier to the block class.
- The subsequent parameters are used to build modifiers. The HTML element will end up with `nrOfModifiers + 1` classes attached to it: 1 for the base block_element class and however many modifier classes. I.e. 
```javascript
const bem = bemClassNames('module-name')
bem('inner', { test: 1===1})
// returns "module-name__inner module-name__inner--test"
```

Perhaps another point of interest is `src/packages/Button/button.css`, lines 110-113 & 139-142. Apart from obviously creating far better CSS scoping, I think that in general the CSS structure and readability has improved by switching to BEM. These two examples in `button.css` are cases where I think we see a deterioration. Personally I don't think it's a major issue, but I just wanted to point out to you that we may bump into an ugly line of CSS every once in a while because of this switch.

The rest of what you see is just how the component JS and CSS files have been updated to work with BEM. And obviously you will see the removal of `classnames` from the `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
         "babel-runtime": "6.26.0",
         "case-sensitive-paths-webpack-plugin": "2.1.1",
         "chalk": "1.1.3",
-        "classnames": "^2.2.6",
         "css-loader": "0.28.7",
         "dotenv": "4.0.0",
         "dotenv-expand": "4.2.0",
--- a/src/App.js
+++ b/src/App.js
@@ -16,21 +16,9 @@ import CircularProgress from './packages/CircularProgress';
 import LinearProgress from './packages/LinearProgress';
 import Logo from './packages/Logo/Logo';
 
-import { bemClassNames } from './utils';
-
 const clickTest = event => {
     console.log('Button click', event.target);
 };
-
-const bem = bemClassNames('d2ui-test');
-const color = 'red';
-console.log(bem()); // Return block name
-console.log(bem('inner-el')); // Return block__element
-console.log(bem(null, 'active')); // Return block--modifier
-console.log(bem('inner-el', 'some-state')); // Return block__element & block__element--modifier
-console.log(
-    bem('second-inner-el', color, { active: color === 'red', 'not-red': color !== 'red' })
-);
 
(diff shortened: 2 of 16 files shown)


varl:
Opened a PR to to your branch with: 

- tests for the `bemClassNames` function and 
- a suggestion about using [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) over [`.forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach). Primarily to reduce the use of callbacks.

As a note, we should get some performance timings on how this function performs as it will execute on _every_ (re-)render of _every_ component.

varl:
Also, yay 🎉 on fewer deps.

HendrikThePendric:
I've merged you PR. Thanks for the improvements. Reg. the null parameter: I did consider working with an object with named keys, i.e. 
```javascript
bem({e: 'element', m: ['string-modifier', {'object-modifier': 1===1}])
```
As you can see, the only way I could think of to make this work is to declare the modifier as an array. Personally I think that just passing everything as a parameter is pretty neat. Having to pass null if you have a modifier on the block is a small price to pay....

Another alternative would be to return an object with a couple of methods attached to it, something like this perhaps:
```javascript
const bem = bemClassNames('d2ui-tabs');
bem.b(); // d2ui-tabs
bem.b('modifier'); // d2ui-tabs--modifier
bem.e('element'); // d2ui-tabs__element
bem.e('element', 'modifier') // d2ui-tabs__element--modifier
```
Actually, the above looks perhaps a little cleaner than what I had before. Let me know what you think, perhaps we should change it.

PR #9, opened 2018-08-21
Package structure

varl:


scripts/copy-files.js
```diff
@@ -0,0 +1,54 @@
...
+            path.join(root, 'node_modules', 'material-design-icons', 'iconfont'),
+            mdi
+        )
+
+        console.info('Copy `typeface-rubik`...')
+        await fs.ensureDir(rubik)
+        await fs.copy(
+            path.join(root, 'node_modules', 'typeface-rubik'),
+            rubik
+        )
```
HendrikThePendric: I think the try block above should be a loop through some configuration object, perhaps in `package.json`.

Suppose you have this in your `package.json`:
```javascript
{
  "assetsToCopy": [
        {
            "src": ["node_modules", "material-design-icons"],
            "target": ["theme", "fonts", "material-design-icons"]
        },
        {
            "src": ["node_modules", "typeface-rubik"],
            "target": ["theme", "fonts", "rubik"]
        }
    ]
}
```

And then you loop through it for each entry.... I think this is nicer in case we want to start copying more stuff.....

varl: Yes, agreed. I'll do that in a separate branch.

PR #23, opened 2018-09-11
Tech 97 input

HendrikThePendric:
I've introduced a whole bunch of form fields here:
- **CheckBox** and **Switch**: both a checkbox really but with a different appearance
- **RadioButton** and **RadioButtonGroup**: because there is no use for radio-buttons by themselves
- **SelectField**: a dropdown, supporting custom and native dropdown elements
- **TextField**: This has a lot of options and a huge CSS file. It can be used to display inputs of type text/search/password/number/etc. but it's also used by the SelectField

The above are to be part of the public API. But I also introduced the **ToggleField** and **FormField** components, which are only used internally by the others.

And, whilst in this branch, I noticed a LOT of things were broken in non-webkit browsers. So this branch also includes quite a few style fixes. I managed to fix almost all cross-browser issues, but for some IE11 rounded corner/ border-radius issues, I reverted to a _graceful degradation_ strategy. I'll mark the relevant files. If you can think of a solution, please let me know.


src/Button/button.css
```diff
@@ -262,12 +265,43 @@
...
             background-color: #033976;
         }
     }
 }
+/* IE11 hacks */
+@media all and (-ms-high-contrast: none) {
+    /* Graceful degradation: remove the rounded corners on all buttons apart from circle.
+       These are not rendered correctly in IE11, where they produce white pixels in the corners  */
+    .d2ui-button {
+        border-radius: 0;
```
HendrikThePendric: If you disable this line and open the example app in IE11 you will see some white pixels in the rounded corners. Maybe you know of a way to get rid of them.

src/Input/textfield.css
```diff
@@ -0,0 +1,342 @@
...
+    &--error.d2ui-text-field--with-value &__floating-label,
+    &--error &__input:focus ~ &__floating-label,
+    &--error &__icon--trailing {
+        color: #e53935;
+    }
+}
+/* IE11 hacks */
+@media all and (-ms-high-contrast: none) {
+    .d2ui-text-field {
+        border-radius: 0;
```
HendrikThePendric: Again... If you disable this line and open the example app in IE11 you will see some white pixels in the rounded corners. Maybe you know of a way to get rid of them.

example-app/src/InputDemo.js
```diff
@@ -0,0 +1,169 @@
...
+            />
+            <SelectField
+                options={selectOptions}
+                label="Choose something"
+                onChange={this.updateSelect}
+                value={this.state.selectValue}
+                variant="outlined"
+                native
+                includeEmpty
+                emptyOptionText="None"
```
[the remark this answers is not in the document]
varl:
```
const selectOptions = [
    { value: null, label: "None" },
    { value: 1, label: 'One' },
    { value: 2, label: 'Two' },
    { value: 3, label: 'Three' },
];
```

varl:
image

When writing long help texts they just force the parent to become wider, according to the MD docs the helper text should just wrap below the input:

image

src/Input/FormField.js
```diff
@@ -0,0 +1,44 @@
...
+
+const FormField = ({
+    children,
+    valid,
+    warning,
+    disabled,
+    error,
+    dense,
+    fullWidth,
+    helpText,
```
varl: I might be going overboard here, but following the MD they call the different labels:

* Label text
* Assistive text

The first is translated to `label` in property form, so I'm thinking we might do the same with the second: `assistive` (annoying to write) or `assist`. Then I note that @joeDHIS  calls it "Helper text", so maybe just `helper` is sufficient? Or `info`?

I vote for `helper`.

src/Input/FormField.js
```diff
@@ -0,0 +1,44 @@
...
+const bem = bemClassNames('d2ui-form-field');
+
+const FormField = ({
+    children,
+    valid,
+    warning,
+    disabled,
+    error,
+    dense,
+    fullWidth,
```
varl: Does `fullWidth` just mean we turn it into a `block`? Could we just call it `block`? :)

varl:
image

RTL is mostly OK, but there are some quirks for the text fields.

src/Input/FormField.js
HendrikThePendric: I'd like to keep `helpText` here because `helper` is often used to describe utility functions, so I am afraid users might start passing functions into it. At the same time, I think that `info` is not descriptive enough. It works for a default state, because then that is what it is: some text that explains what to do with the input. However, once the component goes into _error_ or _warning_ state, then actually the text is there to **help** the user correct its input.

I just think that helpText is wonderfully apt for this situation:
- It is **text**
- It **helps** the user fill in the input correctly.

example-app/src/InputDemo.js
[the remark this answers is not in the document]
HendrikThePendric: You are right that two props for this are complete overkill. However, I think your last suggestion of just including it in the options is not the best way to go. 

First of all because I am making some exceptions for the empty option in the component, so it would be good to keep it separated.

But perhaps more importantly, I think keeping it separate will make for a nicer API. This ways you can use option lists that you receive (from the API) directly. Otherwise, you will continuously have to do stuff like 
```javascript
const options = [ { value: null, label: 'None' }, ...someCollectionAsArray ]
```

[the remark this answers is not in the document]
HendrikThePendric: Now implemented as one property: `emptyOption` (string)

src/Input/FormField.js
HendrikThePendric: implemented as `helperText`

HendrikThePendric:
Below is a summary of how I implemented the PR feedback:
- Updated props: fullWidth => block & includeEmpty + emptyOptionText => emptyOption (kept helpText)
- Implemented dense layout on CheckBox, RadioButton & Switch
- Extracted Field from TextField
- TextField and SelectField now just supply an inputComponent to Field for their styles
- Implemented TextField multiline property, which will render a <textarea />
- Renamed FormField to FieldWrap to avoid confusion with Field component
- Introduced required property and helper function to append asterisks to label texts

varl:
Changes approved.

PR #28, opened 2018-09-17
Tech 113 dialog

HendrikThePendric:
This PR introduces:
- The `Dialog` component
- The `withCloseAnimationHOC` that enables the `Dialog` and `Popover` to have both an appear- and a hide-animation (using CSS keyframe animations + `onAnimationEnd` callback)
- Some edits to the `Popover` to use the new HOC
- A fix on the `Tabs` and the `animatedScrollTo` helper to prevent a performance violation


varl:
Since we merged the new structure for App.js, could you pull out the examples to a Demo component here too?

PR #29, opened 2018-09-20
Tech 104 chip

HendrikThePendric:
After reviewing the MD specs and [the specs that Joe provided](https://projects.invisionapp.com/d/main#/console/14549267/302561749/inspect) I decided to go for a very closed API.

However, whilst writing this PR description, I suddenly remembered [another design document](https://projects.invisionapp.com/share/JFEO0D77T#/screens/266159911) I saw a while ago. This screen seems to feature quite a few chip-like components.... However, these type scenarios would call for a much more open API. I guess props like this:
- leadingIcon => a react element
- trailingIcon => a react element 

Come to think of it.... We could create this component too and call it ChipBase. And then Chip could be a convenience layer over the top of the ChipBase....


varl:
LGTM

adeelshahid:
@HendrikThePendric there are too many changes old/new changes in this PR.

Chip would only need 3 files, you can add them to a new branch and PR.
- index.js
- styles.css
- styles.js

PR #41, opened 2018-09-26
Cache the BEM element names in a hashtable

HendrikThePendric:
Tried a few scenarios and the one below is most performant cross-browser. See https://jsperf.com/bemclassnamesv3


adeelshahid:
Do we need _bemClassNames_ if we are only using it to just prefix, add block name and element type after it. We can probably re-write this function to remove a lot of logic by just using default params.

HendrikThePendric:
I have been leaning on it quite heavily, see some examples:
https://github.com/dhis2/ui/blob/master/src/core/Dialog/Dialog.js#L89-L104
https://github.com/dhis2/ui/blob/master/src/core/Input/Switch.js#L30-L49
https://github.com/dhis2/ui/blob/master/src/core/Tabs/Tabs.js#L243-L272

I'm fine with ditching it, as long as there is a good alternative solution

adeelshahid:
Let's keep it for now, we can refactor it later if need be.

src/utils/bemClassNames.js
```diff
@@ -6,15 +6,21 @@ export default function bemClassNames(blockName) {
...
     return {
         b: (...modifierArgs) =>
             modifierArgs.length === 0
                 ? blockName
                 : classNames(blockName, modifierArgs),
-        e: (elementName, ...modifierArgs) =>
-            modifierArgs.length === 0
-                ? `${blockName}__${elementName}`
-                : classNames(`${blockName}__${elementName}`, modifierArgs),
+        e: (elementName, ...modifierArgs) => {
```
adeelshahid: I don't know how the function works, but will the modifierArgs create a lookup miss

PR #43, opened 2018-09-27
Inputs, some refactor work

varl:
PR for some of the structural work.

- rename **FieldWrap** to **Field** (since it's more generic than the other one)
- rename **Field** to **IconField**
- move both **IconField** and **Field** to `Input/shared/...`
- move Input components to the `Folder/{index.js,styles.css}` convention


src/core/Input/shared/IconField/index.js
```diff
@@ -85,7 +87,7 @@ const Field = ({
 
 Field.propTypes = {
```
HendrikThePendric: Rename to match file type....

HendrikThePendric:
Or perhaps just call it `InputBase`/ `BaseInput`

src/core/Input/shared/IconField/index.js
varl: Oops, thanks.

varl:
Thanks for the FYI, I'll take another whack at that bugger.

varl:
@HendrikThePendric So, it's actually a `LabelField`?

varl:
I renamed the `IconField` to `LabelField` _and_ changed so that Input components either use `LabelField` or `Field` -- not both.

The `LabelField` is wrapped by a `Field` and passes the props it got from the parent to the `Field`. The interfaces are more-or-less compatible. I just added the `helpText` to the property spread for `LabelField` as it was already in the `propTypes` definition.

Now `SelectField` and `TextField` read a bit easier even though the prop complexity is just pushed down one level.

Importing the BEM another component is a sore sight to me - I see why it's done, but if I am not aware of that and I change `${labelFieldBem.e('input')}` to `${labelFieldBem.e('my-text-field')}` it falls apart. I'll think about this some more, this is too brittle for my tastes.

HendrikThePendric:
Excellent changes :-)

And a good comment about exporting the bem from one file to another. It sort of felt a bit wrong to begin with.... But better than retyping the component string.... Your idea about being able to pass a className and overriding that way would solve it. But that would be for another Branch.....

PR #45, opened 2018-09-27
BEM that (Header)Bar

HendrikThePendric:
A small demonstration of how this CSS would look leveraging the BEM system. Much flatter.

I just whipped this up very quickly. If this was a proper refactor, I'd probably change the following:

Instead of letting everything be elements of the same **B**lock I would probably give each JS file its own CSS file and **B**lock name. The advantage of this would be a) more and smaller CSS files, and b) no need for exporting the `bem` const from the `HeaderBar.js`. The downside is probably that this would require the `type` prop to be passed down to the child components.

Some of the names are now not so descriptive, i.e. `&__icon`, what icon? But if we would switch to more granular namespacing, this would be solved.

### DIFF
--- a/src/widgets/HeaderBar/Apps.js
+++ b/src/widgets/HeaderBar/Apps.js
@@ -3,9 +3,11 @@
 import React from 'react'
 import Icon from '../../core/Icon'
 
+import { bem } from './HeaderBar'
+
 export default function Apps() {
     return (
-        <div className="apps-container">
+        <div className={bem.e('apps-container')}>
             <Icon name="apps" />
         </div>
     )
--- a/src/widgets/HeaderBar/HeaderBar.js
+++ b/src/widgets/HeaderBar/HeaderBar.js
@@ -11,23 +11,25 @@ import Profile from './Profile'
 import { bemClassNames } from '../../utils'
 import './styles.css'
 
-const bem = bemClassNames('headerbar')
+export const bem = bemClassNames('headerbar')
 
 function HeaderBar({ type, title, selection, name, email, src }) {
     return (
         <div className={bem.b(type)}>
-            <div className="left">
-                <div className="logo">
+            <div className={bem.e('left')}>
+                <div className={bem.e('logo')}>
                     <Logo
                         color={type === 'blue' ? 'white' : 'blue'}
                         type="icon"
                         width="27.5px"
                     />
                 </div>
-                <div className="title">{title}</div>
+                <div className={bem.e('title')}>{title}</div>
             </div>
-            {selection && <div className="current-selection">{selection}</div>}
-            <div className="right">
+            {selection && (
+                <div className={bem.e('current-selection')}>{selection}</div>
+            )}
+            <div className={bem.e('right')}>
                 <NotificationIcon icon="message" count={8} />
                 <NotificationIcon icon="email" count={4} />
                 <Apps />
--- a/src/widgets/HeaderBar/NotificationIcon.js
+++ b/src/widgets/HeaderBar/NotificationIcon.js
@@ -5,10 +5,14 @@ import PropTypes from 'prop-types'
 
 import Icon from '../../core/Icon'
 
+import { bem } from './HeaderBar'
+
 function NotificationIcon({ icon, count }) {
     return (
-        <div className={`notification ${icon}`}>
-            {count > 0 && <div className="count">{count}</div>}
+        <div className={bem.e('notification', icon)}>
+            {count > 0 && (
+                <div className={bem.e('notification-count')}>{count}</div>
+            )}
             <Icon name={icon} />
         </div>
     )
--- a/src/widgets/HeaderBar/Profile.js
+++ b/src/widgets/HeaderBar/Profile.js
@@ -4,6 +4,8 @@ import React from 'react'
 import Icon from '../../core/Icon'
 import Paper from '../../core/Paper'
 
+import { bem } from './HeaderBar'
+
 function TextIcon({ name }) {
     let title = name[0]
     if (name.indexOf(' ') > 0) {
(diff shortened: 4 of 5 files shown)


adeelshahid:
I'll just add the review inline here which is common across the code.

1. All class names get wrapped in `bem.e('element-name')`.
It creates dependency to `import bem` across all components in HeaderBar.

2. All CSS class names get prefixed with `__`.
Although the structure is flatter because of naming control.

> Instead of letting everything be elements of the same Block I would probably give each JS file its own CSS file and Block name. The advantage of this would be a) more and smaller CSS files, and b) no need for exporting the bem const from the HeaderBar.js. The downside is probably that this would require the type prop to be passed down to the child components.

- The downside of this would be that the code would have more files to manage/read to change stuff.
- The props might need to get passed down at multiple levels.


> Some of the names are now not so descriptive, i.e. &__icon, what icon? But if we would switch to more granular namespacing, this would be solved.

This is the TextIcon/InitialsIcon which can be named better. Semantic improvements like this will happen continuously to make our code better. 

Refactoring improvements like the last one pointed out `&__icon` is a good example of continuous making code better as we go along. I am taking that point and renaming the **className** in my HeaderBar.

Overall from my perspective **bemClassNames** adds un-necessary overhead beyond the top level component.

HendrikThePendric:
I would be interested to find out whether @varl sees things the same way as @adeelshahid. Because if this is the case, I would like to suggest dropping the whole `bemClassNames` helper in its entirety.

I personally see the merits of BEM and this helper function that helps implement it. However, if this helper is used only on the top-level component as suggested, then there is actually no need for this helper at all. It's main benefits are easily creating **E**lement and **M**odifier class names. If these are not used, we might as well just type the component names.

In fact, if we were to use only a scoped class-name on the top level element/component, then we are not implementing BEM at all. I feel it is important to note this, because in actual fact this whole discussion seems to be about whether or not we should use BEM in the first place. I though this was decided already, and we had committed to it. However, I guess we are still at a fairly early stage in the project, which might justify us coming back on our decision if there are compelling reasons for this.

I personally don't think that the overhead of learning BEM classifies as a compelling reason: it's an industry standard and very easy to learn. In fact [just reading this](https://en.bem.info/methodology/quick-start/) is probably enough to get you going and only takes about 10 minutes.

So, the above are just my two cents. I would like this project to be implemented in a consistent way, so would like the CSS methodology to be the same throughout. I'm happy to remove the `bemClassNames` helper and refactor the CSS to remove the BEM system from it, if there are good reasons to switch and/or the rest of the team feels strongly against it. I get the impression that Adeel feels that way, so now just waiting to find out what Viktor thinks.

varl:
I spent some time tinkering with both solutions and this is what I found..

# Biases

Let me start off with saying that _I do not like BEM_. It is a convention which tries to fix the fact that all CSS is added to the global CSS scope. By using naming conventions it essentially namespaces components and creates scopes.

I would much rather solve the problem BEM tries to solve with [CSS Modules](https://github.com/css-modules/css-modules) & [postcss-modules](https://github.com/css-modules/postcss-modules). Unfortunately I have had difficulty implementing this build time for the UI library. If we can get *postcss-modules** running properly this discussion is a no-brainer to me.

# Problem

In **App.css**, the developer has put in:

```
.first {
    background: blue;
}

.second {
    background: pink;
}
```

As an **App developer I rely that my override class is always going to override the internal classes**. `first` and `second ` are very generic names and could be used inside of a component as well. **I still want my override classes to win.**

# BEM

And then the developer tries to nest the BEM components, so the situation in **App.js** with a BEM component becomes something like this:

```
<BEMTest override='first'>
    blue child
    <BEMTest override='second'>
        sibling with pink
        <BEMTest>green child</BEMTest>
    </BEMTest>
</BEMTest>
```

The BEM component is defined as:

```
export default function BEMTest (props) {
    return (
        <div className={bem.b('block')}>
            <div className={bem.e('element')}>
                red child
                <p className={`${bem.e('element', 'first')} ${props.override}`}>
                    {props.children}
                </p>
            </div>
        </div>
    )
}
```

And since it follows the BEM conventions it looks something like:

```
.d2ui-bem-test {
    &--block {
        background: red;
    }

    &__element {
        &--first {
            background: green;
        }
        &--second {
            background: orange;
        }
    }
}
```

Using BEM style namespaces I get this result:

image

The only way for a **Component developer** to screw up for the **App developer** is to have this in the BEM component CSS file:

```
& .first {
    background: yellow;
}
```

Which leads to:

image

The benefit of BEM is that this would be spotted immediately at a code review.

# No BEM

In **App.js**:

```
<NoBEMTest override='first'>
    blue child
    <NoBEMTest override='second'>
        sibling with pink
        <NoBEMTest>green child</NoBEMTest>
    </NoBEMTest>
</NoBEMTest>
```

The NoBEM component is structured the same way, but using nested CSS classes instead:

```
export default function NoBEMTest (props) {
    return (
        <div className="d2ui-no-bem-test defaults">
            <div className="element">
                red child
                <p className={`first ${props.override}`}>
                    {props.children}
                </p>
            </div>
        </div>
    )
}
```

Given the CSS:

```
.d2ui-no-bem-test {
    &.defaults {
        background: red;
    }

    & > .element {
        & > .first {
            background: green;
        }
    }
}
```

The following renders:

image

The nested rules take on a much higher specificity than the override class, and the *App developer*'s experience is not as simple. Even if the **App developer** changes his override classes the nested classes will **always** win.

image

# Summary

Given that we need the ability for an **App developer** to override the styles of a component, then we have two options:

- BEM **without** CSS Modules
- Nested CSS **with** CSS Modules

Since I don't have the time to write the plugins required to make CSS Modules work for us (and it's no guarantee anyway) I say we **proceed with BEM for now**.

varl:
>Given that we need the ability for an App developer to override the styles of a component, then we have two options:
>
>-   BEM without CSS Modules
>- Nested CSS with CSS Modules
>
>Since I don't have the time to write the plugins required to make CSS Modules work for us (and it's no guarantee anyway) I say we proceed with BEM for now.

It's not strictly a given that we need to allow developers to override styles.

varl:
I did some more experimenting and turns out we _can_ get CSS modules working: https://github.com/dhis2/ui/pull/48/files

It requires some work but ultimately that allows us to have complete isolation between components CSS which I think is a very good thing.

HendrikThePendric:
Closed this PR because we have committed to CSS modules now, which makes this code change irrelevant.

PR #48, opened 2018-10-01
Introduce postcss-modules

varl:
https://github.com/css-modules/css-modules
https://github.com/css-modules/postcss-modules

All CSS has local scope (`:local`) by default but global scopes are possible (`:global`).

```
/* composing classes */ 
.otherClassName {
  composes: className from "./style.css";
}

.otherClassNameGlobal {
  composes: globalClassName from global;
}
```

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,4 +1,9 @@
 /** @format */
+
+const path = require('path')
+const fs = require('fs')
+
+/** @format */
 module.exports = {
     plugins: {
         'postcss-import': {},
@@ -16,7 +21,17 @@ module.exports = {
             },
         },
         'postcss-flexbugs-fixes': {},
-        // 'postcss-rtl': {},
-        // cssnano: {},
+        'postcss-rtl': {},
+        'postcss-modules': {
+            getJSON: function(cssFileName, json, outputFileName) {
+                const cssName = path.basename(cssFileName, '.css')
+                const cssDir = path.dirname(cssFileName).split('src/')[1]
+                const jsonFileName = path.resolve(
+                    './build/' + cssDir + '/' + cssName + '.json'
+                )
+                fs.writeFileSync(jsonFileName, JSON.stringify(json))
+            },
+        },
+        cssnano: {},
     },
 }
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,7 +1,7 @@
 /** @format */
 
 const path = require('path')
-const fs = require('fs')
+const fs = require('fs-extra')
 
 /** @format */
 module.exports = {
@@ -26,9 +26,11 @@ module.exports = {
             getJSON: function(cssFileName, json, outputFileName) {
                 const cssName = path.basename(cssFileName, '.css')
                 const cssDir = path.dirname(cssFileName).split('src/')[1]
+                const jsonFileDir = path.join('./build', cssDir)
                 const jsonFileName = path.resolve(
-                    './build/' + cssDir + '/' + cssName + '.json'
+                    jsonFileDir + '/' + cssName + '.json'
                 )
+                fs.ensureDirSync(jsonFileDir)
                 fs.writeFileSync(jsonFileName, JSON.stringify(json))
             },
         },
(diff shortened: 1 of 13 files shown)


postcss.config.js
```diff
@@ -16,7 +21,17 @@ module.exports = {
...
-        // cssnano: {},
+        'postcss-rtl': {},
+        'postcss-modules': {
+            getJSON: function(cssFileName, json, outputFileName) {
+                const cssName = path.basename(cssFileName, '.css')
+                const cssDir = path.dirname(cssFileName).split('src/')[1]
+                const jsonFileName = path.resolve(
+                    './build/' + cssDir + '/' + cssName + '.json'
+                )
+                fs.writeFileSync(jsonFileName, JSON.stringify(json))
```
varl: Here we can do static parsing of the JSX files and replace the classes which are not computed runtime.

src/core/Logo/index.js
```diff
@@ -2,15 +2,17 @@
...
+import { componentStyles } from '../../utils'
 
 import './styles.css'
 import iconBlue from './svg/icon-blue.svg'
 import iconWhite from './svg/icon-white.svg'
 import logoBlue from './svg/logo-blue.svg'
 import logoWhite from './svg/logo-white.svg'
 
-const bem = bemClassNames('logo')
+import './styles.css'
```
varl: This CSS file contains the CSS modules, e.g. `._logo_1234xyz`.

src/core/Logo/index.js
```diff
@@ -2,15 +2,17 @@
...
+import styles from './styles.json'
```
varl: This JSON file contains the mapping from `.logo` to `._logo_1234xyz`. It's possible to enable `camelCase` so it's possible to just use `styles.logoIcon`.

```
// styles.json 
{
  "logo-icon": "._logo-icon_116zl_1",
  "logoIcon": "._logo-icon_116zl_1"
}
```

src/core/Logo/index.js
```diff
@@ -2,15 +2,17 @@
...
+const myst = componentStyles(styles)
```
varl: This function would do what `bemClassNames` does today but with the CSS modules.

src/utils/css.js
```diff
@@ -48,6 +48,12 @@ function classNames(prefix, modifierArgs) {
     return classes.join(' ')
 }
 
+export function componentStyles(styles) {
```
varl: We can export all styles into a single JSON and use it here, but that would mean that even if a user only wants a **Button** s/he would have to import the entire JSON file which maps classes to CSS modules.

src/utils/css.js
```diff
@@ -48,6 +48,12 @@ function classNames(prefix, modifierArgs) {
...
+    return function classNames(classy) {
+        return styles[classy]
```
varl: This is obviously the simplest possible function to return resolved values from the CSS modules, here we would support variable arguments, building up a complete class list from booleans, etc. which is needed runtime.

postcss.config.js
```diff
@@ -16,7 +21,17 @@ module.exports = {
             },
         },
         'postcss-flexbugs-fixes': {},
-        // 'postcss-rtl': {},
-        // cssnano: {},
+        'postcss-rtl': {},
+        'postcss-modules': {
```
varl: We can add `generateScopedName` to get names like `d2ui-component__className__hash` if we want to customize the generated names.

src/core/Logo/styles.css
```diff
@@ -1,9 +1,9 @@
...
+.logo {
     width: 11.25rem;
+}
 
-    &.icon {
-        width: 2.5rem;
-    }
+.icon {
+    width: 2.5rem;
 }
```
varl: This is obviously a simple file, the point is that it's completely flat with no extra namespacing.

HendrikThePendric:
I'll have to familiarize myself with PostCSS Modules a bit more, but the changes look good to me as far as I can tell.

PR #50, opened 2018-10-02
move animation files

adeelshahid:


src/utils/index.js
```diff
@@ -1,8 +1,8 @@
 /** @format */
+
 export * from './html'
 export * from './css'
 export * from './rtl'
 export * from './text'
-export * from './animations'
 export * from './timings'
-export * from './withAnimatedCloseHOC'
+export * from './animations/index'
```
HendrikThePendric: you don't need `/index` here I think

adeelshahid: For some reason `./animations` didn't get work on in my dev env. Try at your end if it works.

PR #57, opened 2018-10-03
Tabs using postcss-modules

adeelshahid:
**Changes**
- Used _postcss-modules_
- Updated package _code-style_ so style changes in master reflects **tabs-postcss**

#### Issues

- [ ] TabsDemo is missing from the App
- [ ] animatedScrollTo - not found in Tabs.js
- [ ] computeHorizontalScrollbarHeight.js has a `export default function`
- [ ] `.test-horizontal-scrollbar-height` not sure if this has anything todo with Tabs code.


HendrikThePendric:
I've addressed the issues you mention, apart from the last one, which was a question I will attempt to answer below:
`.test-horizontal-scrollbar-height` is being used by `computeHorizontalScrollbarHeight`. It creates a div and applies this class-name to it before appending it to the DOM. This way it is positioned off-canvas and guaranteed to have a horizontal scroll bar. Basically [the same as what is done here](https://github.com/material-components/material-components-web/blob/master/packages/mdc-tab-scroller/util.js#L39-L55).

PR #62, opened 2018-10-03
CircularProgress, Icon

adeelshahid:
- CircularProgress: Separate out SVG Icon
  semanatic rename progressBar to view as it's a CircularProgress

- Icon
  use undefined for onClick
  use correct syntax to refer to className

### DIFF
--- a/src/core/CircularProgress/index.js
+++ b/src/core/CircularProgress/index.js
@@ -4,7 +4,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import s from './styles'
 
-function Icon({ dashed }) {
+function Circle({ dashed }) {
     return (
         <svg viewBox="22 22 44 44" className={s('viewbox')}>
             <circle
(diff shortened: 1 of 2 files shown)


src/core/CircularProgress/index.js
```diff
@@ -4,28 +4,34 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import s from './styles'
 
+function Icon({ dashed }) {
```
HendrikThePendric: I think this name is too ambiguous. We already have an Icon. I know this one is only local to the file, but still. Perhaps just `Circle` or `SvgCircle` would be better.

src/core/CircularProgress/index.js
```diff
@@ -4,28 +4,34 @@ import React from 'react'
...
+                strokeWidth="3.6"
+                strokeDashoffset="0"
+            />
+        </svg>
+    )
+}
+
 function CircularProgress({ size, center, dashed }) {
-    const progressBar = (
+    const view = (
```
HendrikThePendric: I agree that `progressBar` was bad, because it's a circle. Could we perhaps call it `loader` or `progressIndicator` please

src/core/CircularProgress/index.js
adeelshahid: updated with _Circle_

src/core/CircularProgress/index.js
adeelshahid: updated with _loader_

PR #92, opened 2018-10-15
fixes(headerbar)

adeelshahid:
- Rename property selection to status in HeaderBar
- Use built-in core components to generate User Profile popup
- Attach default props to menu so it can be used independently from Menu
- Set correct color for menu Icon


varl:
I noticed that the App component is still using the old TextField.

adeelshahid:
Input still have to be adjusted. Probably one of the last thing with the HeaderBar

PR #94, opened 2018-10-16
Radio

varl:
- Uses same SVG system as Checkbox
- Removes old Input/RadioButton{,Group}
- Removes old Input/Checkbox

These Radio's doesn't use a Group component, instead the grouping is done by the `group` attribute. What the wrapping Form needs to do is to keep track of which Radio is selected, and pass the `checked` attribute to it. 

See the demo for an example.


varl:
>- `htmlFor={this.props.name}`, `id={this.props.name}` can be removed as `<label />` wrapper will auto-focus click event on the _input:radio_

done for checkbox and radio

>- we should allow `value` prop as `evt.target.checked` will always be true, `this.props.onChange(this.props.name, this.props.value)`, otherwise each input would require a different name to track state.

done for radio

>- we can use `name` attr. instead of group to auto-group input:radio by default and remove `group` attr.

name is used for groups, value is used to keep track of selection

PR #96, opened 2018-10-17
Styles

adeelshahid:
**Radio**
- Inline checked condition for icon name
- Sort propTypes by priority

**Button, SelectField**
Move _removeEventListener_ under _componentWillUnmount_. Remove _componentWillMount_

**Checkbox**
Remove !!, if 'undefined' checked, indeterminate by default would be false


src/core/Radio/index.js
```diff
@@ -11,19 +10,15 @@ class Radio extends React.Component {
...
-        }
-
+        let color = this.props.disabled ? 'grey-light' : 'grey'
         const icon = (
             <Icon
-                name={icoName}
-                className={`${icoColor} ${this.props.status}`}
+                name={
+                    this.props.checked
+                        ? 'radio_button_checked'
```
varl: I see we are missing a state here, a checked radio button should be `secondary-light`.

PR #98, opened 2018-10-18
Chip with working demo

HendrikThePendric:


src/core/Chip/styles.css
```diff
@@ -0,0 +1,112 @@
...
+    transition: all 150ms cubic-bezier(0.4, 0, 0.2, 1);
+    transition-property: background-color, color;
+    user-select: none;
+}
+
+.container:hover {
+    background-color: #e0f2f1;
+}
+
+.static {
```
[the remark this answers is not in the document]
adeelshahid: as an example 10,000 people were polled, and here is a brief overview of how group by type can help us order our CSS
https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/

src/core/Chip/index.js
```diff
@@ -0,0 +1,69 @@
...
+        )
+    }
+
+    render() {
+        const {
+            label,
+            onClick,
+            selected,
+            disabled,
+            dragging,
```
adeelshahid: we don't have a use case for this in UI, remove dragging

src/core/Chip/index.js
```diff
@@ -0,0 +1,69 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Icon from '../Icon'
+import s from './styles'
+
+// Only allow lower case letters, optionally seperated by underscore
+const mdIconNamePattern = /^[a-z]+_*[a-z]+$/
```
adeelshahid: remove the regex, we'll simplify this by adding `type` to allow `Image` or `Icon` as in other place in code.

src/core/Chip/index.js
```diff
@@ -0,0 +1,69 @@
...
+
+class Chip extends React.PureComponent {
+    onRemove = event => {
+        event.stopPropagation()
+        this.props.onRemove()
+    }
+
+    renderIcon(icon) {
```
adeelshahid: use `type` property Icon or Image, default to `icon` rather than regex logic to simplify code

src/core/Chip/index.js
```diff
@@ -0,0 +1,69 @@
...
+
+        return (
+            <div
+                className={s('container', {
+                    static: !onClick,
+                    selected,
+                    disabled,
+                    dragging,
+                })}
+                onClick={(!disabled && onClick) || undefined}
```
adeelshahid: - simplify this to `onClick={onClick}`
- handle disabling button using css as you already are adding the disabled class to `.container`
- add `Chip.defaultProps` for `onClick: undefined`

src/core/Chip/index.js
```diff
@@ -0,0 +1,69 @@
...
+            >
+                {icon && this.renderIcon(icon)}
+                {label && <span className={s('label')}>{label}</span>}
+                {onRemove && (
+                    <Icon
+                        name="cancel"
+                        className={s('remove-icon')}
+                        onClick={this.onRemove}
```
adeelshahid: use `this.props.onRemove` rather than canceling even bubbling ourself

src/core/Chip/index.js
```diff
@@ -0,0 +1,69 @@
...
+                    static: !onClick,
+                    selected,
+                    disabled,
+                    dragging,
+                })}
+                onClick={(!disabled && onClick) || undefined}
+            >
+                {icon && this.renderIcon(icon)}
+                {label && <span className={s('label')}>{label}</span>}
+                {onRemove && (
```
adeelshahid: onRemove is used in one place so replace with `this.props.onRemove`

src/core/Chip/index.js
```diff
@@ -0,0 +1,69 @@
...
+                        className={s('remove-icon')}
+                        onClick={this.onRemove}
+                    />
+                )}
+            </div>
+        )
+    }
+}
+
+Chip.propTypes = {
```
adeelshahid: - add type property to tackle `icon`, `image`
- add optional `src` in case Icon on the side is a image

src/core/Chip/index.js
```diff
@@ -0,0 +1,69 @@
...
+    label: PropTypes.string.isRequired,
+    onClick: PropTypes.func,
```
varl: I see that we should sort the prop types better, in order of importance.

src/core/Chip/styles.css
```diff
@@ -0,0 +1,85 @@
...
+
+.dragging {
+    box-shadow: 0 3px 1px -2px rgba(0, 0, 0, 0.2),
+        0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 1px 5px 0 rgba(0, 0, 0, 0.12);
+}
+
+.icon {
+    color: #494949;
+    font-size: 20px;
+    margin-left: 5px;
```
adeelshahid: Icon
- `margin-left` is `8px`
- Icon area is 24px, but star is 16px in size with 4px padding

src/core/Chip/styles.css
```diff
@@ -0,0 +1,85 @@
...
+
+.remove-icon:hover {
+    opacity: 0.82;
+}
+
+.selected {
+    background-color: #00796b;
+}
+
+.selected:hover {
```
adeelshahid: move `.selected` and `selected:hover` class relative to the `.container` at the top

src/core/Chip/index.js
varl: If we add the `src` prop couldn't we infer that it's an `image`?

then we could just use: `icon` or ``src` as props and drop `type`.

```
<Chip src="<url>" />
<Chip icon="<icon name>" />
```

src/core/Chip/index.js
HendrikThePendric: `.disabled` class does not have `pointer-events: none;` but `cursor: not-allowed;`. Both can't work together. Hence my current solution. A lot of components are using `cursor: not-allowed;` for disabled states (Button, Menu, LabelFields, etc), so if we want to change it here, maybe we should also visit other components to make sure we are consistent.

src/core/Chip/index.js
HendrikThePendric: This would trigger both `onRemove` and `onClick` because the click event bubbles up from the close icon to the chip.

src/core/Chip/styles.css
HendrikThePendric: Checked this with Joe, my styles are OK, the specs are outdated.

src/core/Chip/index.js
varl: i think `type` and `icon` read nicely and leave it flexible enough for other types without changing the api, no need for `src`:

```
<Chip icon="material_icon_1" />
<Chip type="image" icon="/img_icon.png" />
<Chip type="glyph" icon="\xE2\x9C\x8F" />
```

src/core/Chip/styles.css
```diff
@@ -0,0 +1,86 @@
+.container {
```
adeelshahid: group and sort CSS based on https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/

src/core/Chip/index.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Icon from '../Icon'
+import s from './styles'
+
+class Chip extends React.PureComponent {
+    onRemove = event => {
```
adeelshahid: use succinct names `evt` instead of `event`

src/core/Chip/styles.css
```diff
@@ -0,0 +1,86 @@
...
+.image-icon {
+    border-radius: 50%;
+    height: 24px;
+    margin-left: 4px;
+    width: 24px;
+}
+
+.label {
+    margin: 0 12px;
+    max-width: 150px;
```
adeelshahid: - we should add a `prop` *overflow* with `default = true`
The Chips on the Dashboard are already > 150px. Then add the css `.container.overflow`

screenshot 2018-10-18 at 16 35 05

src/core/Chip/index.js
```diff
@@ -0,0 +1,72 @@
...
+        event.stopPropagation()
+        this.props.onRemove()
+    }
+
+    renderIcon() {
```
adeelshahid: use a shorter name e.g. `viewIcon`

src/core/Chip/index.js
```diff
@@ -0,0 +1,72 @@
...
+            </div>
+        )
+    }
+}
+
+Chip.propTypes = {
+    label: PropTypes.string.isRequired,
+    onClick: PropTypes.func,
+    icon: PropTypes.string,
+    type: PropTypes.oneOf(['image', 'icon']),
```
adeelshahid: place default prop first e.g. `icon, image`

src/core/Chip/index.js
```diff
@@ -0,0 +1,72 @@
...
+                    />
+                )}
+            </div>
+        )
+    }
+}
+
+Chip.propTypes = {
+    label: PropTypes.string.isRequired,
+    onClick: PropTypes.func,
```
adeelshahid: move event functions `onClick` to the end after all the props

src/core/Chip/index.js
```diff
@@ -0,0 +1,72 @@
...
+    icon: PropTypes.string,
+    type: PropTypes.oneOf(['image', 'icon']),
+    onRemove: PropTypes.func,
```
adeelshahid: move event functions `onClick` to the end after all the type props

src/core/Chip/index.js
```diff
@@ -0,0 +1,72 @@
...
+    selected: PropTypes.bool,
+    disabled: PropTypes.bool,
+    dragging: PropTypes.bool,
+}
+
+Chip.defaultProps = {
+    onClick: undefined,
```
adeelshahid: - move `defaultProps` before `propTypes`
- place `onClick` after `type`

src/core/Chip/index.js
```diff
@@ -0,0 +1,72 @@
...
+    onClick: PropTypes.func,
+    icon: PropTypes.string,
+    type: PropTypes.oneOf(['image', 'icon']),
+    onRemove: PropTypes.func,
+    selected: PropTypes.bool,
+    disabled: PropTypes.bool,
+    dragging: PropTypes.bool,
+}
+
+Chip.defaultProps = {
```
adeelshahid: add default props for `selected, disabled, dragging`

- selected: false
- disabled: false
- dragging: false
- onRemove: undefined

src/core/Chip/index.js
```diff
@@ -0,0 +1,72 @@
...
+                className={s('container', {
+                    static: !onClick,
+                    selected,
+                    disabled,
+                    dragging,
+                })}
+                onClick={(!disabled && onClick) || undefined}
+            >
+                {icon && this.renderIcon(icon)}
+                {label && <span className={s('label')}>{label}</span>}
```
adeelshahid: remove `label && ` because `label` prop. is always required.

src/core/Chip/styles.css
varl: 2018-10-18-170226_326x267_scrot

The ones coded are the right ones atm.

src/core/Chip/styles.css
varl: 4358c7d, opted for that `overflow: true` means _show long names_ and `overflow: false` to mean _truncate the long name_.

src/core/Chip/index.js
varl: 4358c7d

src/core/Chip/index.js
varl: 4358c7d

src/core/Chip/index.js
varl: 4358c7d

src/core/Chip/index.js
varl: 4358c7d

src/core/Chip/index.js
varl: 4358c7d

src/core/Chip/index.js
varl: 4358c7d, they were ordered by functional importance (based on my comment).

src/core/Chip/index.js
varl: 4358c7d

src/core/Chip/index.js
varl: 4358c7d, Using functions to show icons if necessary to clean up the render function.

```
{this.showIcon()}
<span className={s('label', { overflow })}>{label}</span>
{this.showRemove()}
```

src/core/Chip/index.js
varl: What he means is that if we use this code:

```
.container.disabled {
    pointer-events: none;
    cursor: not-allowed;
}
```

Then when you hover the `.container`, you will not get the `cursor: not-allowed`, because no pointer events will be fired by the browser from the `.container` element.

varl: We can move the special logic to a function instead of having it inline.

b4b5314

src/core/Chip/styles.css
varl: ok

PR #101, opened 2018-10-19
style(InputField)

adeelshahid:
`kind: outline`
- fix CSS for border-bottom (left/right) radius
- fix CSS border-width

### Spec

screenshot 2018-10-19 at 10 17 41

### Implementation

screenshot 2018-10-19 at 10 18 15

### DIFF
--- a/src/core/InputField/index.js
+++ b/src/core/InputField/index.js
@@ -9,7 +9,11 @@ class InputField extends React.Component {
         focused: false,
     }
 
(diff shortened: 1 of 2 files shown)


varl:
- clicking on a InputField that is selected causes the `border-bottom` to flash

101-flash-on-click

- it is possible to use tab to get to `disabled` fields and type in them

101-disabled-input-fields

PR #108, opened 2018-10-22
style(UI): set default fonts for LTR, RTL

adeelshahid:
- remove index.css. App.css contains the related code
- set default fonts for LTR and RTL

### DIFF
--- a/src/core/UI/base/typography.css
+++ b/src/core/UI/base/typography.css
@@ -11,6 +11,16 @@
  * *****************************************************************************
  */
 
+
+
+[dir="ltr"] body, [dir="ltr"] div {
+    font-family: Roboto, sans-serif;
+}
+
+[dir="rtl"] body, [dir="rtl"] div {
+    font-family: "Noto Naskh Arabic", sans-serif;
+}
+
 body,
 .primary-text {
     font-size: 1rem;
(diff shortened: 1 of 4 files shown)


src/core/UI/base/typography.css
```diff
@@ -11,6 +11,16 @@
...
  */
 
+
+
+[dir="ltr"] body, [dir="ltr"] div {
+    font-family: Roboto, sans-serif;
+}
+
+[dir="rtl"] body, [dir="rtl"] div {
+    font-family: "Noto Naskh Arabic", sans-serif;
```
varl: have you added a package for this to `package.json` as well?

adeelshahid: I wanted your perspective on this as we talked briefly about i18n last week.

Currently we use `import 'typeface-roboto'`, but there is no `typeface-noto-naskh-arabic` or similar available.

If we can control these fonts as small packages published on our _namespaced_ `@dhis2/typeface-${font-face}` repo we can add consistency at this layer.

varl: if we add a typeface package to the `@dhis2` scope we could move the `typeface-roboto` dep to that package as well and bundle all of our typefaces in a single package.

adeelshahid: sounds like a good idea

PR #109, opened 2018-10-25
fix(api): url, endpoint is undefined in dev. environments

adeelshahid:
for dev. environment `url` and `endpoint` is undefined.

### DIFF
--- a/src/utils/api.js
+++ b/src/utils/api.js
@@ -19,6 +19,10 @@ if (isProd) {
     if (!endpoint) {
         endpoint = `${url}/api`
     }
+} else {
+    // for dev. environments
+    url = DHIS_CONFIG.baseUrl
+    endpoint = `${url}/api`
 }
 
 export const serverURL = url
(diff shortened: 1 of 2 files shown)


src/utils/api.js
```diff
@@ -19,6 +19,10 @@ if (isProd) {
     if (!endpoint) {
         endpoint = `${url}/api`
     }
+} else {
+    // for dev. environments
```
varl: We do the same above: https://github.com/dhis2/ui/blob/a72a485ef98733bb72aeeb07bd53852926bcc052/src/utils/api.js#L13-L15

Maybe we should remove the L13-L15 block as `DHIS_CONFIG` only exists in Dev?

adeelshahid: re-written for better readability

```js
const IS_PROD = process.env.NODE_ENV === 'production'

let url
let endpoint

if (IS_PROD) {
    url = '..'
    endpoint = `${url}/api`
    if (typeof manifest !== 'undefined') {
        let version = manifest.dhis2.apiVersion
        url = manifest.activities.dhis.href
        endpoint = `${url}/api/${version}`
    }
} else {
    // for dev. environments
    url = DHIS_CONFIG.baseUrl
    endpoint = `${url}/api`
}
```

PR #111, opened 2018-10-30
chore: remove global var DHIS_CONFIG dependency

adeelshahid:


varl:
@amcgee please do. I don't like `DHIS_CONFIG` global var either which is why this PR exists in the first place. The main reason I haven't merged it is because I didn't like the hardcoded URL either.

Go ahead with: `process.env.REACT_APP_DHIS2_BASE_URL` for now.

Once we get our own little system going maybe we can reduce that to `process.env.DHIS2_BASE_URL` or come up with something else.

As far as `d2` goes ... This lib will _never_ have an awareness of the full-on `d2` lib. At most we will have a dependency to `d2-api` to do requests and replace the `utils/api.js` functionality.

amcgee:
@varl I updated the PR to use the DHIS_CONFIG global variable if it exists (for backwards compatibility), the REACT_APP_DHIS2_BASE_URL environment variable if it exists, and to fail with a runtime exception if neither is set (when NODE_ENV === 'development' only)

There are any number of ways we could do this - we could fall back to a hard-coded value instead of eroring out, but I think this is an OK compromise for now.

TODO (line 9): Support non-local DHIS2 server URLs in production
TODO (line 18): Fully deprecate DHIS_CONFIG global and support cleaner DHIS2_BASE_URL environment variable

varl:
>@varl question - should the default (when no env var is set) be a hard-coded URL like play.dhis2.org/dev or localhost:8080 or should we throw an error in that case? I'm leaning towards the latter.

Error, please!

varl:
>However I think this is just a stop-gap measure until we have better wholistic application configuration.

Agreed. 👍

PR #119, opened 2018-11-01
fix(style-overrides)

adeelshahid:
we should better identify the target use-case in destination, as this over-ride will add un-necessary background to all `<a/>` tags on focus/click


adeelshahid:
indeed, we need to review our work :slightly_smiling_face:

PR #125, opened 2018-11-20
Babel, Fonts, No regenerators

varl:
- only use core-js polyfills for what we use
- avoid pulling in regenerator runtime and use promises instead
- **remove need for** `import 'ui/core/UI'`
- app needs to pull in Roboto font; it needs it anyway for everything else so doesn't make sense for the lib to require it, if not there, fallback to sans-serif.
- use RTL plugin to generate font-name for RTL mode; app needs to bundle "Noto Naskh Arabic".
- add font mixin for components to use


varl:
@adeelshahid what was the review comment? github dismissed it when i pushed a commit and i cannot get back to it? :|

adeelshahid:
@varl no comment 🙂 , probably some empty comment in PR

PR #128, opened 2018-12-19
Isolated components

varl:
I ran in to a big problem where the Application could ruin all styles for the UI components by doing for example:

```
div {
  padding: 20px;
}
button {
  background-color: purple;
  border: 5px solid orange; 
}
```

In order to make components isolated I've added a `.reset` class to all the components which uses `all: initial` to reset the styles to the browser defaults. This in turn had the consequence of changing the assumption which we had unwittingly built all the components on, and broke all the styles again.

I have gone through all the components and gotten them into a state where they look right again. I had to refactor SelectField and InputField again but now they work better, albeit at the cost of sharing less code.

As master is currently not in a great state, and this PR is already huge, I think we should get this in so we can start refactoring:

* simplify InputField and SelectField
* see if we can pull label out into it's own component again after the rewrite
* go over the component APIs and make them consistent and simpler
* see if we can automatically add `.reset` and `:root` to all stylesheets with colors and reset styles.


varl:
>see if we can automatically add `.reset` and `:root` to all stylesheets with colors and reset styles.

Done, kind of. `:root` is added automatically and resolves variables correctly with a fallback to the color hex.

For `.reset` i opted for a common css file with the reset class which each `styles.js` file loads and merges with the component styles. Then two variables are exposed: `cx` and `rx`.

`cx` is for normal classNames, without the reset class **prepended** (important!) to the class list.

`rx` is for classNames which should be reset using the reset class. The reset class must be **prepended** to the class list for the correct specificity!

PR #137, opened 2019-01-31
docs: link to dhis2/ui docs instead of create-react-app

amcgee:
Maybe I'm missing something, but having a link to create-react-app docs here seems wrong.  Updating to point at the gh-pages publication, but maybe this section is not needed since that page is liked from the top of the readme?

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,4 +46,4 @@ Make a PR from the component branch. Add a team member who can review the code a
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


varl:
Just a habit someone had to link to the CRA docs instead of having their massive generated README in the repo. I didn't think of removing it when I revamped the repo to a library.

PR #142, opened 2019-02-19
fix(reset): b and strong should be bold and i and em italic

HendrikThePendric:
@joakimia pointed out that the some of the resets might be a bit too aggressive. Specifically, the rule `font: inherit;` has some big consequences, it touches the following font-related styles:
```css
font-style: inherit; /* this is a problem for <i> and <em> */
font-variant-ligatures: inherit;
font-variant-caps: inherit;
font-variant-numeric: inherit;
font-variant-east-asian: inherit;
font-weight: inherit; /* this is a problem for <b> and <strong> */
font-stretch: inherit;
font-size: 100%;
line-height: inherit;
font-family: inherit;
```
I've added two additional rules to address these issues.

### DIFF
--- a/src/defaults/reset.css
+++ b/src/defaults/reset.css
@@ -91,6 +91,14 @@ video {
     font-size: 100%;
     vertical-align: baseline;
 }
+b,
+strong {
+    font-weight: bold;
+}
+i,
+em {
+    font-style: italic;
+}
 /* HTML5 display-role reset for older browsers */
 article,
 aside,


varl:
Some of these `inherit` values should perhaps be `initial`?

HendrikThePendric:
Yeah you could argue that too.... Using `inherit` makes it a very consistent UI.... Using `initial` will make the elements have the browsers default styles, which isn't the same across browsers, but makes it a very "native" UI... It just depends on what we want. Personally, I prefer consistency.

varl:
>Using `initial` will make the elements have the browsers default styles, which isn't the same across browsers, but makes it a very "native" UI... It just depends on what we want. Personally, I prefer consistency.

Ok, that's fair enough and I agree.

Along that line of reason we should replace `bold` (which just refers to the browser standard) with an integer defining the weight. I think `bold` is `700` in most browsers, so using `bold` is consistent enough.

PR #143, opened 2019-02-19
refactor: port to styled-jsx

varl:
I just want to share this. I recommend either checking out the branch locally and exploring or use the functionality on github to explore the code base at a specific commit.

### `styled-jsx`

```
  Package size: 26.44 KB
  Size limit:   50 KB
  With all dependencies, minified and gzipped
```
Lines of Code
```
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
JavaScript                      55            783             95           3186
CSS                              4             74             17            793
JSON                             3              0              0             30
-------------------------------------------------------------------------------
SUM:                            62            857            112           4009
-------------------------------------------------------------------------------
```

### PostCSS

Size
```
  Package size: 26.26 KB
  Size limit:   50 KB
  With all dependencies, minified and gzipped
```

Lines of Code
```
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
CSS                             20            481             82           3485
JavaScript                      44            745             94           3032
JSON                            19              0              0            222
-------------------------------------------------------------------------------
SUM:                            83           1226            176           6739
-------------------------------------------------------------------------------
```

### `styled-jsx`

```
File sizes after gzip:

  63.05 KB build/static/js/main.dfd56472.js
  4.2 KB build/static/css/main.1238f341.css
```

### PostCSS

```
File sizes after gzip:

  56.96 KB build/static/js/main.7f2b9ccc.js
  10.88 KB build/static/css/main.f172d73f.css
```

## Compiled component

Compare the two compiled `Button` components to verify that what is being exported is just basic React elements without any framework wrapper components added.

### `styled-jsx`

```js
function Button(_ref) {
  var type = _ref.type,
      children = _ref.children,
      kind = _ref.kind,
      size = _ref.size,
      icon = _ref.icon,
      label = _ref.label,
      disabled = _ref.disabled,
      onClick = _ref.onClick;
  return _react.default.createElement("button", {
    disabled: disabled,
    onClick: onClick,
    className: "jsx-".concat(_styles.default.__hash) + " " + ((0, _classnames.default)('base', "kind-".concat(kind), "size-".concat(size), {
      'icon-only': icon && !label && !children,
      icon: icon
    }) || "")
  }, icon && _react.default.createElement(_Icon.default, {
    className: (0, _classnames.default)('button-icon'),
    name: icon
  }), label || children, _react.default.createElement(_style.default, {
    id: _styles.default.__hash
  }, _styles.default));
}
```

### PostCSS

(technically has little to do with postcss but using the term for consistency)

```js
function Button(_ref) {
  var type = _ref.type,
      children = _ref.children,
      kind = _ref.kind,
      size = _ref.size,
      icon = _ref.icon,
      label = _ref.label,
      disabled = _ref.disabled,
      onClick = _ref.onClick;
  return _react.default.createElement("button", {
    disabled: disabled,
    onClick: onClick,
    className: (0, _styles.rx)('base', "kind-".concat(kind), "size-".concat(size), {
      'icon-only': icon && !label && !children,
      icon: icon
    })
  }, icon && _react.default.createElement(_Icon.default, {
    className: (0, _styles.default)('button-icon'),
    name: icon
  }), label || children);
}
```

## Notes

- This just straight up moves all the CSS from `.css` files to the `styled-jsx/css` template strings
- There is a lot of duplicated CSS with PostCSS which can be removed using `styled-jsx`
- There are a lot of nested override rules which are not needed with `styled-jsx`
- `style.json` (generated) and `styles.css` can be removed completely from **each** component
- Components only need `index.js` and `styles.js`
- Since no css/json/js bindings are needed, we can produce cjs/es bundles properly without duplicating css files
- We can simplify the watch/build process
- We need to figure out how to do RTL again (it is possible to use `postcss` with `styled-jsx`)
- Then again, we get a chance to fix the `dir` attribute problem properly for our CSS scopes
- `styled-jsx` does not mess with the exported React components, so the library does not leak to the host application
- This gives us flexibility to replace `styled-jsx` in the future if we need to (escape hatch)
- This removes the depedency on `css-loader` and `style-loader` and replaces it with a runtime dependency


amcgee:
This looks like a great change to me, especially after acting on this point: 

> * There is a lot of duplicated CSS with PostCSS which can be removed using `styled-jsx`

Personally I like `styled-jsx` a lot, and think scoped CSS-in-JS is a Good Thing (tm).  Plus the escape hatch is nice to have.

For completeness, what would an application need to do to include the `styled-jsx` runtime and what would happen if it weren't included?  (I think I know the answer to that one ;-))

varl:
>For completeness, what would an application need to do to include the `styled-jsx` runtime and what would happen if it weren't included? (I think I know the answer to that one)

The way it's set up now, the `styled-jsx` runtime is a direct dependency so applications do not need to include it.

If the application does include it, it is going to be fine for now, but when the versions start drifting apart, then we will have to start using [`yarn install --flat`](https://yarnpkg.com/en/docs/cli/install#toc-yarn-install-flat) to force Yarn to reconcile a single version, or specify `styled-jsx` as a peer dependency in `ui`.

If it isn't included at all, then entire thing doesn't work:

>./node_modules/@dhis2/ui/core/Button/index.js
>Module not found: Can't resolve 'styled-jsx/style' in '/home/varl/temp/styled-test/node_modules/@dhis2/ui/core/Button'

varl:
As HeaderBar is moving out, I flattened the folder structure:

```
build/
+-- Button
+-- Card
+-- Checkbox
+-- Chip
+-- CircularProgress
+-- css
+-- Divider
+-- DropdownButton
+-- HeaderBar
+-- Help
+-- Icon
+-- InputField
+-- LinearProgress
+-- Logo
+-- Menu
+-- Radio
+-- SelectField
+-- SplitButton
+-- Switch
+-- utils
    +-- animations

21 directories
```

This accomplishes two things:

1. Draws a clear line in the sand that the `ui` library is the **core** library.

2. Nice imports:

```js
import Button from '@dhis2/ui/Button'
import { Card } from '@dhis2/ui'
```
And for the `reset.css` it'll be something like: `import '@dhis2/ui/css/reset.css'`

varl:
Some more notes:

- So far it seems to be a better idea to duplicate component CSS to keep each component isolated.

- Overriding e.g. `Icon` styles from the wrapping `Button` is tricky to do, better to use another `div` or something to apply e.g. padding/margin.

- The Material Design Icon font forces the host app to have a `css-loader`, so we should probably bundle the SVGs we need for the `ui` components instead.

- That the components (widgets) does API-fetching is a Bad Idea™. We will quickly end up with having to maintain multiple branches of the HeaderBar (v31, v32, etc.). Better to have the Apps fetch the data for the components and pass in relevant bits and bobs as props.

varl:
- Generic class names like `.title` and `.logo` might get added from the host application, so it is important to think about when writing the components.

- If we want to keep all UI components in the `ui` library (e.g. HeaderBar) we need to have `d2-i18n` in the `ui` as well.

- Complex components which have other dependencies would add them to the `ui` list of dependencies.

varl:
I have ported all components to `styled-jsx` and this is the final tally of the build:

```
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
JavaScript                      52            741            100           3160
JSON                             2              0              0              2
CSS                              2              0              0              2
-------------------------------------------------------------------------------
SUM:                            56            741            100           3164
-------------------------------------------------------------------------------
```

Overall I'm quite pleased with how `styled-jsx` works. There are some quirks, and just porting the components is probably not the best way to do this, so the components do need polish. Overall though, it all worked over expectations.

There are two things which are outstanding:

- RTL mode
- IE11 bug fixes for fields

I do think that it's a decent starting point.

varl:
Sticking with the core/widgets separation for now.

From `size-limit`:

```
  Package size: 27.45 KB
  Size limit:   50 KB
  With all dependencies, minified and gzipped
```

amcgee:
👍

Issue #141, opened 2019-02-16
Wishlist:  Keyboard friendly input components


teug91:
It is not possible to tab to a select field, radio, checkbox, or switch. Navigating with enter and arrow keys within a select field would also be a nice feature.

varl:
Absolutely! This is on the feature list. 👍

varl:
@teug91 is this still a problem in v2?

teug91:
@varl 

The main issue is that you cannot use enter to toggle a switch/radio/checkbox or tab to them. The select field can be controlled using the keyboard, but there is an issue with the label (see #216).

Repo I made for testing:
https://github.com/teug91/test-app

varl:
Thanks @teug91, will take a look.

Mohammer5:
I just created a test story locally to check this, here's what I noticed:

### 1. SelectField (filled & outlined)

* The "focused" state isn't indicated
* It's still focused, so entering a value with the keyboard will select a value
* Hitting "enter" while the SelectField has focus will open the dropdown

### 2. InputField (filled & outlined)

* Works as expected 
  * Displays the focused state

### 3. Radio & Checkbox

* Doesn't handle focus at all (no keyboard events possible)

### 4. Checkbox

* The value of the event passed to the `onChange` callback is `on`/`off`, not the value passed to the `Checkbox` component

## Conclusion

1. The `SelectField` should be relatively easy to fix as it uses the same wrapper as the `InputField`
2. The `Checkbox` and `Radio` component need to be fixed to handle the focus event

Branch that I tested on is: `form_stories`

PR #146, opened 2019-02-21
feat(headerbar): enter-shortcut to follow applink

Birkbjo:
<del>Adds support to press enter to go to the first app when searching for an app.
<del>
The approach might seem a bit weird, let me explain:
I did not want to use refs to track focus or set it programmatically, I also did not want to make Item a stateful component, as a thought was to have the listener added there - but that also means new event listeners for every item! Without this we cannot know what AppItem is in the users focus. Therefore I just have a flag "hasTabbed" to know if the user has tabbed, and only allow the 'enter to follow applink' when the user has not tabbed yet. 
<del>
One downside I've found is that I wanted 'tabbing" from the search to go to the second item, as the first item in the list seems to be already selected, and setting 'tabIndex=-1' would break 'backtabbing' to the first item. There's probably ways around this, but I did not want to complicate the code more. 
___

This PR kind of shifted focus, as I just wanted to fix a simple use case, but ended up leaking a lot of features related to navigating the apps. However I believe I've ended up with a intuitive result. And even though #143 is still being considered, this PR should not have much consequences for that (1 line of CSS).

I ended up with not handling "tabbing" programmatically, and just control the focus when using arrow-keys. The tabbing is entirely handled by the browser, and for me both ways works intiutively for me now. 
Would be nice if someone else pulls the PR and test for themselves. 

Some notes:

- "Simulated" focus and enter-shortcut to follow first app while searching
- Apps can be navigated through with up- and down-arrows. Searchfield will not loose focus. And can thus be navicated while searching.
- Arrows work only when searchbox is focused
- Tabbing out of the searchbox currently focuses first app, this might change when we update rest of UI to be keyboard-friendly (see #141)
- Backtabbing from list to searchbox will clear the simulated first app, but will once again be focused when changing the input. This felt natural for me, so I did not add more logic to prevent this. Might want to set the selectedIndex to 0 when searchfield is focused though. 
- Esc closes the menu

Minor code notes:
- I started using "handleEvent", as I like this pattern for eventHandlers, therefore I renamed the other methods as well. I see that this is not consistent in the repo though. 
- Event listeners are added when opening the menu, as I believe this shouldn't leak beyond the scope of the app-menu.


amcgee:
@Birkbjo It looks like this doesn't support tabbing through the list of apps and then hitting `enter` to navigate to the highlighted one, is that correct?  Maybe I'm missing something.

I think it's more intuitive to support `enter` no matter what app is highlighted, and also to support `shift + tab` to increment selection back towards the beginning of the result set.

amcgee:
Oops, hasty comment before reading your description, sorry.

I think it's wrong to only support "enter" to select the first item, it will be very confusing to the user.  Wouldn't it be possible to make `<List>` a controlled component which accepts "focusedIndex" rather than "firstFocused" (possibly with a callback if the List does the tab handling) and keep the statefulness here where it already exists?  Then it should be possible to support `enter` whether or not the user has tabbed through the list, and there shouldn't be a requirement to add event listeners to each item

Birkbjo:
It does work as it uses the browsers native focus, so when pressing enter on a link, it will be activated.

amcgee:
Ah ok got it!  Thanks for the clarification.

In that case I think this works, though it makes the implementation awkward and prone to loss-of-context if revisited by a future dev...

One consideration - I think intuitively a user will want to use the arrow keys to select through the list of applications, as you might with a type-ahead text input, rather than with the tab key.  For accessibility we might need to support tab as well, I'm not sure.

In the 2.30 headerbar the arrow keys allowed navigation through the apps panel, although they also moved the cursor in the search textbox which isn't ideal.

Birkbjo:
I agree with the arrowkeys, and something I've thought about. In that case we would probably need to control the focus programatically with an index as you mentioned, rather than relying on native 'tabbing' (and probably disable native tabbing with tabindex='-1' on the elements to prevent conflicts). 

The conflicting context you are mentioning is a problem though, as the user may want to use the arrowkeys to move the cursor in the search-field, but instead moving through the apps. Anyways its something a user will quickly figure out, so we just need to decide on arrows and tabs or just tabs. 

Preventing moving the cursor could probably easily be fixed by checking for the arrow key-codes and calling preventDefault().

varl:
Having the first item pre-selected is a nice addition. 

Regarding the arrow keys I do not like overriding the default behaviour, so as long the input field has focus the arrow keys should act normally. If the input loses focus (e.g. `tab` is hit), then the arrow keys could navigate in the list which is now focused.

amcgee:
I buy that.  A nice touch might be to support "down arrow" and then arrow keys to navigate among the panel, but that's a bonus.

joeDHIS:
Thanks for the @, I'm late giving input here, but agreed that we should not override default behaviour for the arrow keys when focus is in the text/input field. Once the user has tabbed to the list then we can, and should, allow for keyboard navigation of the list.

'Down arrow' to move from input to list is a nice idea, but default down arrow behaviour when focused in a field is 'go to end of inputted text' and we should not override that.

Birkbjo:
I've now moved to a more controlled approach, instead of relying on native tabbing, as you mentioned @amcgee. 
I noticed some weird behaviour on Firefox on mac, where it would not tab to links. There's apparently an OS setting where you can choose whether to 'Press Tab to move keyboard focus between: Text boxes and lists only, and All controls'. The default is Text boxes and lists only, and firefox respects this setting.

Moving to controlled tabbing also makes it possible to add arrow navigation pretty easily. Currently the input field will still be focused when tabbing. To prevent this we would need to allow tabbing just from the input-field to list, but not further between the items as we want to control the tabbing ourselves, and not have conflicts between what is 'native focused' and what is 'in our control focus'. We would also need some logic to allow back-tabbing to the list when first item is selected.

Before I look further into this, I guess I need your input for what the UX should be. As the first item is 'always' focused, (along with the input) do we want the first tab to focus the second item, or just move the focus away from the input field (and still focus the first item)?

amcgee:
Nice @Birkbjo!

@joeDHIS I understand not wanting to override that native browser textbox behavior, but I do think it's a common pattern for getting from search box to search results (see google.com)

Some prior art:
- **ICONS, NO ARROWS, TABS** [Zapier](https://zapier.com/apps) supports search-as-you-type for their app list, which displays as a grid of icons.  
- **TEXT, ARROWS, NO TABS** [Google](https://www.google.com)
- **LIST, ARROWS, NO TABS** [Dropbox file search](https://www.dropbox.com/h)

I'm sure there are plenty more

amcgee:
> Before I look further into this, I guess I need your input for what the UX should be. As the first item is 'always' focused, (along with the input) do we want the first tab to focus the second item, or just move the focus away from the input field (and still focus the first item)?

My naïve opinion would be that if we support tabs (see above) the tab should move the aria-focus (aka the blue glow in chrome) from the textbox to either the whole app selector popup (with the first item still highlighted) or to the first item (also highlighted).  I feel like supporting tabs gets weird fast, but @joeDHIS might disagree with me

joeDHIS:
Supporting tabs does indeed get weird fast, we need to strike the balance between expected behaviour/effective patterns. I've been doing some thinking, and I'm currently thinking about this set of behaviours, which is a blend of the ideas discussed above:

- Tabbing to apps menu icon, then pressing enter opens the apps menu
- Focus is automatically placed into search field when apps menu opens
- Typing in search field filters apps menu
- Arrow left/right moves text cursor when focus is on search field
- Arrow down/up navigates app items when focus is on search field. Eg. pressing down arrow 3 times while focus is in search field would highlight the third app in the menu. Enter key opens the highlighted app. Down moves left to right, moving to the next row when necessary (thus, all apps are reachable using down/up only - this theoretically makes more sense if the markup is a unordered list.) ((Apologies if this is explained badly))
- Tab also moves from each app-icon item, removing focus from the search input. If tab is pressed on the last app-icon the focus returns to search input. (is this possible? Temporarily defining an isolated tabindex?)
- Esc closes the apps menu, but focus remains on the apps menu icon.

These are untested principles, so ideally I'd like to test them with an example if possible. Very open to all opinions here. There is no set rules here, we can override defaults if we feel it brings enough value. We should of course be consistent with whatever behaviour we define.

amcgee:
> * Tab also moves from each app-icon item, removing focus from the search input. If tab is pressed on the last app-icon the focus returns to search input. (is this possible? Temporarily defining an isolated tabindex?)

Quick note - this loop would truncate the tabindex list and make the rest of the page inaccessible with the keyboard, so I don't think we want to implement it as a loop.

joeDHIS:
> > * Tab also moves from each app-icon item, removing focus from the search input. If tab is pressed on the last app-icon the focus returns to search input. (is this possible? Temporarily defining an isolated tabindex?)
> 
> Quick note - this loop would truncate the tabindex list and make the rest of the page inaccessible with the keyboard, so I don't think we want to implement it as a loop.

Ah yeah, I think I got myself confused thinking about the loop here. I think if possible the down/up arrow keys should loop through the app icons. I agree the tabs should not.

Birkbjo:
Thanks for the input everyone!

This PR kind of shifted focus (no pun intended!), as I just wanted to fix a simple use case, but ended up leaking a lot of features related to navigating the apps. However I believe I've ended up with a intuitive result. And even though #143 is still being considered, this PR should not have much consequences for that (1 line of CSS).

I ended up with not handling "tabbing" programmatically, and just control the focus when using arrow-keys. The tabbing is entirely handled by the browser, and for me both ways works intiutively for me now. 
Would be nice if someone else pulls the PR and test for themselves. 

Some notes:

- "Simulated" focus and enter-shortcut to follow first app while searching
- Apps can be navigated through with up- and down-arrows. Searchfield will not loose focus. And can thus be navicated while searching.
- Arrows work only when searchbox is focused
- Tabbing out of the searchbox currently focuses first app, this might change when we update rest of UI to be keyboard-friendly (see #141)
- Backtabbing from list to searchbox will clear the simulated first app, but will once again be focused when changing the input. This felt natural for me, so I did not add more logic to prevent this. Might want to set the selectedIndex to 0 when searchfield is focused though. 
- Esc closes the menu

Minor code notes:
- I started using "handleEvent", as I like this pattern for eventHandlers, therefore I renamed the other methods as well. I see that this is not consistent in the repo though. 
- Event listeners are added when opening the menu, as I believe this shouldn't leak beyond the scope of the app-menu.

varl:
Headerbar as moved to dhis2/ui-widgets.

PR #147, opened 2019-02-26
Chore/getting started docs

jenniferarnesen:
Add docs to help developers get started with contributing


jenniferarnesen:
Wondering why there is a webpack config, and why we bother with the size limit check?

PR #177, opened 2019-03-27
refactor: extract shared constants from input/select

Mohammer5:
Hi @varl 

I tried to fix the styling issues in the select component and saw that there are many shared constants between the select and input component. As I saw there's an `src/icons` directory already, I thought I can just continue from there.

But @HendrikThePendric told me that the `ui-core` lib is built in such a way that every component is independent from each other. So I was wondering if what I did is ok after all or if this explicitly unwanted.
I didn't want to adjust styling on top of work that could be reversed..

So here's a small summary of what I did:
1. I took the Status icons, moved related stuff into `src/icons/constants.js`, `src/icons/helpers.js` and `src/icons/styles.js`
1. I extracted some shared stuff between `InputField` and `SelectField` and moved it to `src/forms/constants.js`
1. I moved the styles located in `src/SelectField/index.js` to `src/SelectField/styles.js`


varl:
Thanks for raising this, @Mohammer5, and props for a good working memory, @HendrikThePendric. 😁

>But @HendrikThePendric told me that the ui-core lib is built in such a way that every component is independent from each other.

The reason we built it this way was because of how the `ui-core` library was constructed. When we used PostCSS and CSS Modules, it could not de-duplicate the stylesheets it would inject into the page.

The consequence of this was that sharing styles between components lead to duplicate stylesheets being injected into the page. First off, this was bloat. It could also cause strange conflicts in the stylesheets.

N.B. **When we switched to `styled-jsx` this is no longer the case.**

So to sum up to the best of my knowledge; it is now possible and preferable to share code between components.

src/InputField/index.js
```diff
@@ -226,8 +171,8 @@ InputField.propTypes = {
     help: PropTypes.string,
     icon: PropTypes.element,
     trailIcon: PropTypes.element,
-    status: PropTypes.oneOf(['default', 'valid', 'warning', 'error']),
-    size: PropTypes.oneOf(['default', 'dense']),
+    status: iconStatusPropType,
+    size: PropTypes.oneOf([inputSizes.DEFAULT, inputSizes.DENSE]),
```
varl: Since you import `inputSizes` already, maybe we could expose a list of the sizes and do something like: `PropTypes.oneOf(inputSizes.alternatives)`?

Any drawbacks?

src/InputField/index.js
```diff
@@ -226,8 +171,8 @@ InputField.propTypes = {
...
     kind: PropTypes.oneOf(['filled', 'outlined']),
```
varl: Eventually I would like to redo the `kind` props as well into sharing the scheme better between components.

src/InputField/index.js
Mohammer5: Hmm what would be the usage of said list except for using it for the propTypes?
The user of `ui-core` certainly would only care about the constants.

We could do the following:
```js
import { identity, map } from 'lodash/fp'
const extractValues = map(identity)

InputField.propTypes = {
    // ...
    status: PropTypes.oneOf(extractValues(iconStatuses)),
    size: PropTypes.oneOf(extractValues(inputSizes)),
    // ...
}
```

Then we wouldn't need to adjust all propTypes of components using those sizes.
On the other hand.. this introduces complexity, some components might supports only certain inputSizes and when adding new sizes, the input components would have to be touched anyways in order to support the new size if they should handle the new size.

Or did I overlook another possible use-case for `inputSizes.alternatives`?

src/InputField/index.js
Mohammer5: I thought about this as well, but then realized that the components that have a `kind` propType (InputField and SelectField) have different kinds, so I didn't see the necessity to pull this out of the file as that'd be over-engineered code splitting to me.

I think until we see there's shared usage, we don't need to extract them. But I will def. add constants for this, just like for the other values.

Mohammer5: Okay, nevermind what I said... InputField and SelectField share the same kinds....

Mohammer5:
I've extracted the `kinds` into a form constant.
I also exposed the constants that the InputField and SelectField use in their modules, so the users of this lib can do:

```
import Select, { iconStatuses, inputSizes } from '@dhis/ui-core/SelectField'
```

src/InputField/index.js
varl: You're right, verbose and simple is better here.

varl:
> I've extracted the `kinds` into a form constant.
> I also exposed the constants that the InputField and SelectField use in their modules, so the users of this lib can do:
> 
> ```
> import Select, { iconStatuses, inputSizes } from '@dhis/ui-core/SelectField'
> ```

No, we won't allow deep imports anymore. Our libraries will have a single entry-point (`main`/`module` in `package.json`). So it will need to be: `import { Select, iconStatuses, inputSizes } from '@dhis2/ui-core'`.

But, do the users of the `ui-core` library really need to use `iconStatuses` and `inputSizes` directly? This creates that pattern that we want to expose those options for all components.

src/InputField/index.js
```diff
@@ -226,14 +183,20 @@ InputField.propTypes = {
...
+        types.PASSWORD,
+        types.URL,
+    ]),
     focus: PropTypes.bool,
     disabled: PropTypes.bool,
     required: PropTypes.bool,
 }
 
-export { InputField }
+export { InputField, types, iconStatuses, inputSizes, inputKinds }
```
varl: `types, iconStatuses, inputSizes, inputKinds` can be removed.

src/SelectField/index.js
```diff
@@ -316,16 +262,16 @@ SelectField.propTypes = {
...
     disabled: PropTypes.bool,
     required: PropTypes.bool,
+    icon: PropTypes.element,
+    size: PropTypes.oneOf([inputSizes.DEFAULT, inputSizes.DENSE]),
+    kind: PropTypes.oneOf([inputKinds.FILLED, inputKinds.OUTLINED]),
+    status: iconStatusPropType,
 }
 
-export { SelectField }
+export { SelectField, iconStatuses, inputSizes, inputKinds }
```
varl: `iconStatuses, inputSizes, inputKinds` can be removed.

src/SelectField/styles.js
```diff
@@ -329,3 +330,24 @@ export default css`
...
+        pointer-events: none;
+    }
+`
+
+export const menuOverride = resolve`
+    max-height: 300px;
+    overflow-y: auto;
+`
+
+export const selectIconStyles = resolve`
```
varl: Do you need `resolve` here?

It is better if you can use `css`.

src/icons/Status.js
```diff
@@ -1,46 +1,143 @@
...
             viewBox="0 0 48 48"
-            className={className}
+            className={[iconStyle.className, styles.className, className].join(
+                ' '
+            )}
         >
-            <path d="M0 0h48v48H0z" fill="none" />
-            <path d="M24 4C12.95 4 4 12.95 4 24c0 11.04 8.95 20 20 20 11.04 0 20-8.96 20-20 0-11.05-8.96-20-20-20zm-4 30L10 24l2.83-2.83L20 28.34l15.17-15.17L38 16 20 34z" />
+            {children}
+            <style>{iconStyle.styles}</style>
```
varl: The tag should be `<style jsx>`. Then you should be able to switch to:

```
const iconStyle = css`
    svg {...}
`

<style jsx>{iconStyle}</style>
```

src/icons/styles.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import { resolve } from 'styled-jsx/css'
+import { colors } from '../theme.js'
+
+export const iconStyle = resolve`
```
varl: `resolve` is not the primary way to use `styled-jsx`, see comment above.

src/SelectField/styles.js
Mohammer5: The reason I used `resolve` is because I changed the status icons in `src/icons/Status.js` to reuse duplicate code, but styled-jsx prevents styled from bubbling down, so I couldn't just add a style for the svg in the `Valid` component as the svg will be rendered by the `Icon` component.

**I can revert that**, but then I think styled-jsx enforces an anti-pattern on us as I'm not able to create abstract components anymore unless I use inheritance.
The Icons all share this code:
```jsx
        <svg
            xmlns="http://www.w3.org/2000/svg"
            width="48"
            height="48"
            viewBox="0 0 48 48"
            className={[iconStyle.className, styles.className, className].join(
                ' '
            )}
        >
            {children}
            <style>{iconStyle.styles}</style>
            <style>{styles.styles}</style>
        </svg>
```

src/icons/Status.js
Mohammer5: see comment above, depends on what we agree on there

src/icons/styles.js
Mohammer5: see comment above, depends on what we agree on there

src/SelectField/styles.js
varl: I do not think that abstract Icon component is a helpful abstraction, it adds another layer of indirection. I prefer this style:

```js
const iconStyle = css`
 svg {
     width: 24px;
     height: 24px;
     margin-right: 8px;
}
`

export function Default({ className }) {
    return (
        <svg
            xmlns="http://www.w3.org/2000/svg"
            width="48"
            height="48"
            viewBox="0 0 48 48"
            className={className}
        >
            <path d="M0 0h48v48H0z" fill="none" />
            <style jsx>{`
                svg {
                    fill: ${colors.grey700};
                }
            `}</style>
            <style jsx>{iconStyle}</style>
        </svg>
    )
}

export function Valid({ className }) {
    return (
        <svg
            xmlns="http://www.w3.org/2000/svg"
            width="48"
            height="48"
            viewBox="0 0 48 48"
            className={className}
        >
            <path d="M24 4C12.95 4 4 12.95 4 24c0 11.04 8.95 20 20 20 11.04 0 20-8.96 20-20 0-11.05-8.96-20-20-20zm-4 30L10 24l2.83-2.83L20 28.34l15.17-15.17L38 16 20 34z" />
            <style jsx>{`
                svg {
                    fill: ${colors.blue600};
                }
            `}</style>
            <style jsx>{iconStyle}</style>
        </svg>
    )
}

export function Warning({ className }) {
    return (
        <svg
            xmlns="http://www.w3.org/2000/svg"
            width="48"
            height="48"
            viewBox="0 0 48 48"
            className={className}
        >
            <path d="M2 42h44L24 4 2 42zm24-6h-4v-4h4v4zm0-8h-4v-8h4v8z" />

            <style jsx>{`
                svg {
                    fill: ${colors.yellow500};
                }
            `}</style>
            <style jsx>{iconStyle}</style>
        </svg>
    )
}

export function Error({ className }) {
    return (
        <svg
            xmlns="http://www.w3.org/2000/svg"
            width="48"
            height="48"
            viewBox="0 0 48 48"
            className={className}
        >
            <path d="M24 4C12.96 4 4 12.95 4 24s8.96 20 20 20 20-8.95 20-20S35.04 4 24 4zm2 30h-4v-4h4v4zm0-8h-4V14h4v12z" />

            <style jsx>{`
                svg {
                    fill: ${colors.red500};
                }
            `}</style>
            <style jsx>{iconStyle}</style>
        </svg>
    )
}
```

The `colors.red500` could be a `theme.error` instead, and same for the other states. Simpler is better.

varl:
Added comment about the abstraction of `Icon`.

src/SelectField/styles.js
varl: >I can revert that, but then I think styled-jsx enforces an anti-pattern on us as I'm not able to create abstract components anymore unless I use inheritance.

I'm not so sure that I consider that an anti-pattern.

Mohammer5: > Could you help me understand why this is a useful abstraction?

For the following two reasons:
1. It's much easier to add more icons
1. It's very easy to change icon details

For me, it's harder to see the differences between the different icons if they're "cluttered" with the svg code, which my brain will always read first as I don't know whether there are differences until I read the whole code.

But I get the point that my way of thought/approaching code is not how other devs do it, so I totally understand why it's a cognitive overhead for the others. I'm curious how others see this!

varl: If the other icons do not share the same `viewBox` and/or dimensions? Then the those attributes would have to be parameters as well?

Now the icon abstraction would look kind of similar to the `svg`:

```
const Icon = ({ viewBox, height, width }) => <svg viewBox={viewBox} height={height} width={width}>{children}</svg>
```

Another concern I have is that the `path` element only makes sense inside of the XML namespace  of http://www.w3.org/2000/svg; it belongs as close to the `<svg>` tag as possible. 

🤔

Mohammer5: > If the other icons do not share the same viewBox and/or dimensions? 

This would be a good reason for refactoring of course, but that'd be changing requirements. From what I can tell so far, this is very unlikely to happen, so we'd be preparing for an unlikely eventuality.

But I'll revert the change. It's too unimportant to have a controversy about this

Mohammer5: See latest commit

PR #178, opened 2019-03-27
fix: adjust InputField and SelectField alignments

Mohammer5:
# What I changed

1. The `InputField` and `SelectField` components now match the styleguide.
1. The native dom elements are now split into single components together with their styled-jsx styles
1. I've added docs and stories

## What I had to do to get there.

1 It's impossible to style the native `fieldset` and `legend` element when it comes to details,
so I've written a custom Fieldset element.
1. In order to reuse the Fieldset element / shared logic between the `InputField` and `SelectField`, I've extracted the logic into shared components, which I have put into `src/forms`.

# What's left to discuss

1. The components I've put into `src/forms/` are called `LabelFilled` and `LabelOutlined`. But the names for these components are wrong, as they are not just the label, they also do a bit more, so I'm not sure how to name them properly right now
1. I've removed the `icon` and `trailIcon` props from the `InputField` and `SelectField` components as I was unable to get them working properly. Is there any known use-case for them? Or do you know any app that makes use of those? To me it seems as this has been implemented for a single use case in some app (so I'd rather make this a breaking change and disallow it as it 1. breaks the alignment of the texts and 2. is not reflected in the styleguide)


varl:
Cool! 👍 

One thing to note; we are _not_ following the MUI specs. Any similarities are accidental. :)

The input field specs from Joe which we are codifying are here: https://projects.invisionapp.com/d/main#/console/14549267/317906756/inspect

joeDHIS:
> Cool! 👍
> 
> One thing to note; we are _not_ following the MUI specs. Any similarities are accidental. :)
> 
> The input field specs from @joeDHIS which we are codifying are here: https://projects.invisionapp.com/d/main#/console/14549267/317906756/inspect
> 
> Could you verify that we are following those specs in terms of alignments, etc.?

Yes, that's right.

src/InputField/index.js
```diff
@@ -214,6 +220,7 @@ InputField.defaultProps = {
     focus: false,
     disabled: false,
     required: false,
+    styles: {},
```
varl: I am not certain we want to allow this.

Mohammer5: I think we shouldn't, I'll revert some changes.
It was harder to style the fieldset in combination with the legend element, 
but we def. shouldn't allow overriding styled (otherwise it'll be done),
thanks @varl

varl:
@Mohammer5 this is getting rather large, how far to you want to take it before it's ready for review and eventually merge it?

Mohammer5:
@varl is almost done, one commit is missing. But I thought about splitting it into several chunks, just not sure how to proceed, but will try to improve the history by squashing some of the commits.

varl:
@Mohammer5 i have updated the base to `next`, I think we want to synchronise releasing these changes with `origin/next`.

Mohammer5:
@varl I have adjusted the inconsistencies between the `InputField` and `SelectField` and fixed some minor visual bugs..

Although there are many commits, there's not too much being changed.. I'll update the description of this issue to lay out what I've done in this PR.

docs/SelectField.md
```diff
@@ -0,0 +1,44 @@
...
+
+This is a rich select field, allowing to show errors,
+hint texts and well as validity states like valid, warning and error.
+The `list` property supports a second level, which will be rendered as
+optgroup. Right now, the parent option will be rendered as a choosalbe option
+as well and it's label will be the label for the optgroup.
+
+## API
+
+This is a list of supported prop types:
```
varl: Great! But documenting APIs manually becomes stale quickly.

We can generate it to show it with Storybook as a tab: https://github.com/hipstersmoothie/storybook-addon-react-docgen (would be in a different PR)

varl:
Do you know of a solution for the gap in the border here?:
2019-04-09-082509_121x93_scrot

There is no space for the icon to the right?:
2019-04-09-082522_68x81_scrot

stories/SelectField.stories.js
```diff
@@ -9,80 +9,119 @@ const options = [
...
+
+    onChange = (name, value) => {
+        this.setState({ value })
+    }
+
+    render() {
+        const { value, ...props } = this.props
+
+        return (
+            <div style={{ width: '200px' }}>
```
varl: I think we can remove this wrapper `div` so that the `SelectField` and `InputField` get the same amount of space in the demo.

docs/SelectField.md
Mohammer5: Oh, very nice tool! More value with less effort :heart: good fine!

Mohammer5:
> Do you know of a solution for the gap in the border here?:

@varl Yes, easily, I just mimicked the fieldset's behaviour here. But it's not part of the styleguide.. not sure why I did that.. good catch

stories/InputField.stories.js
```diff
@@ -1,139 +1,106 @@
...
+)
+
+createStory('InputField (Outlined)', createWrapperOutlined)
+createStory('InputField (Outlined and dense)', createWrapperOutlinedDense)
+createStory('InputField (Filled)', createWrapperFilled)
+createStory('InputField (Filled and dense)', createWrapperFilledDense)
+
+function createStory(name, wrapperCreator) {
+    storiesOf(name, module)
+        .add('No placeholder, no value', wrapperCreator(), { notes })
```
varl: Nice! :100: 

Really like that you added the docs you wrote to `notes`. :+1:

Mohammer5:
> There is no space for the icon to the right?:

@varl I can't reproduce this, did you update your local branch?
i3_screenshot_2019-04-09-14-47-55

varl:
@Mohammer5 Based on a discussion in Slack with @cooper-joe we can remove custom leading and trailing icons and only have status icons for trailing.

> > There is no space for the icon to the right?:
> 
> @varl I can't reproduce this, did you update your local branch?

You're right, looks fine now after an update. I was running code from yesterday.

I noticed you went with this structure (`forms`, `InputField`):

```
InputField/
├── index.js
└── InputField
    └── Input.js
```

I think it's too early to deepen the hierarchy as we simply do not have a lot of files, I would extract `Input` into its own component and expose it. Remember that we are dealing with atomic elements which cannot be used by themselves, but a consumer of the library could create new **molecules** based on our **atoms**.

```
InputField/
└── index.js
Input/
└── index.js
```

Mohammer5:
> I would extract Input into its own component and expose it.

@varl I think it's very good that you bring this up early on already! I wanted to do what you are proposing before I decided against it.

The reason that I've put the components into `src/InputField/`, `src/SelectField/` and `src/forms/` is that they have styles that are very purpose specific and are likely subject to change.
When the style guide changes just a little (which would be a breaking change to all apps using those components), while it  wouldn't create a breaking change, if not exposed.

I do see sense in making them public but this would require good documentation (including stating there's a change these might change in the future and will cause the apps using them to adjust). Question is: Should this be part of this PR or a second PR

varl:
>I do see sense in making them public but this would require good documentation (including stating there's a change these might change in the future and will cause the apps using them to adjust). Question is: Should this be part of this PR or a second PR

Ah, good point. We can move the directory structure now and in the `src/index.js`:

```
/* These will be exposed at some point, but are kept internal for now until the API stabilizes */
// export { Input } from './Input'
```

varl:
Very nice refactor! Good work. :+1: 

I'll change my review and merge after the final fixes. :)

varl:
:dancer:

PR #183, opened 2019-04-09
feat: version 2 of ui-core

varl:
BREAKING CHANGE: This changes the `list` prop passed to DropdownButton and SplitButton to `component`, which enables any component to be used as a menu.


src/FieldLabel/Filled/index.js
```diff
@@ -0,0 +1,168 @@
...
+    </div>
+)
+
+Filled.propTypes = {
+    label: propTypes.string.isRequired,
+    children: children.isRequired,
+    hasValue: propTypes.bool.isRequired,
+    htmlFor: propTypes.string.isRequired,
+
+    tailIcon: propTypes.element,
```
HendrikThePendric: Do we really need a `tailIcon`? A `leadingIcon`/`icon` prop would make more sense to me. A valid input with tailicon would now produce two icons in a row

src/SelectField/index.js
```diff
@@ -1,145 +1,96 @@
...
         document.addEventListener('click', this.onDocClick)
-
-        this.setState({ labelWidth: this.labelRef.current.offsetWidth })
     }
 
     componentWillUnmount() {
         document.removeEventListener('click', this.onDocClick)
     }
 
     onDocClick = evt => {
```
HendrikThePendric: In the new implementation, why do we need to keep track of the document clicks? Just tapping into the focus and blur event on the select element should be enough....

src/Logo/index.js
```diff
@@ -30,8 +30,8 @@ LogoIcon.defaultProps = {
 }
 
 LogoIcon.propTypes = {
-    className: PropTypes.string,
-    fill: PropTypes.string,
+    className: propTypes.string,
+    fill: propTypes.string,
```
HendrikThePendric: I don't think people should be allowed to specify any fill here. According to the DHIS2 identity, we have 4 versions of the logo. I think this component should reflect that.
- Icon only VS icon with text
- Blue on white VS white on blue

src/Logo/index.js
```diff
@@ -94,11 +94,9 @@ Logo.defaultProps = {
 }
 
 Logo.propTypes = {
-    className: PropTypes.string,
-    fill1: PropTypes.string,
-    fill2: PropTypes.string,
+    className: propTypes.string,
+    fill1: propTypes.string,
+    fill2: propTypes.string,
```
HendrikThePendric: See previous comment

src/DropdownButton/index.js
```diff
@@ -92,20 +60,12 @@ class DropdownButton extends Component {
...
 
-                {open && (
-                    <div className="menu" ref={c => (this.elMenu = c)}>
-                        <Menu
-                            list={this.props.list}
-                            onClick={this.props.onClick}
-                        />
-                    </div>
-                )}
+                {open && <div className="menu">{this.props.component}</div>}
```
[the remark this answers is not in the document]
varl: Refactored so the `className` is redundant.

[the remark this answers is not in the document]
varl: 40b88e5

src/FieldLabel/Filled/index.js
varl: Good catch, I have verified with @cooper-joe that we do not need `trailIcon` except for status icons.

varl: d8ab0ae

src/Logo/index.js
varl: 👍

src/Logo/index.js
varl: Agreed, I've improved the prop names and locked the external API.

4371430

src/SelectField/index.js
varl: Probably should have been to keep track of if the label on select fields should shrink or not, but that feature is broken so I'll clean up the code and remove it.

varl: 78439ce

src/DropdownButton/index.js
```diff
@@ -92,20 +60,12 @@ class DropdownButton extends Component {
                         <span className="button-icon">{this.props.icon}</span>
                     )}
 
-                    <span className="menu-label">{this.props.label}</span>
+                    {this.props.label || this.props.children}
```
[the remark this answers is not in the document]
varl: I think that having both options here is redundant, so let's standardise on the `<button>` way. I've also added `name` and `value` to `Button` and the left part of the `SplitButton` so it is possible to use them with normal form controls. They are optional and will default to empty strings.

The onClick handler passed in as a prop will also receive `name` and `value` as the first and second parameters. 

51f2efa

varl:
>     1. Storybook problems/suggestions
>        
>        1. DropDownButton/ SplitButton With Menu doesn’t work (nothing happens when you click)
>        2. Menu and SelectField throw an error

Fixed!

>        3. InputField demos are using some abstraction over the InputField, the InputFieldWrapper. It would be better to just use the InputField directly, so it is visible in the JSX tab how to use it.
> 
>     2. SplitButton and DropDownButton use a very simple div with position absolute to render stuff into a “Popover-menu”. This approach is fragile: there are going to be situations where the menu isn’t rendered fully if the container has overflow-y hidden, and also potential z-index problems if parents also have certain z-index properties (a child's z-index is set to the same stacking index as its parent). It would be better to render stuff in a portal (React.createPortal) as the last child of the document body. Stuff does become more complex this way because the position will have to be computed, but it will be more robust.

Good point. Will take a look.

>     3. States: I think it would be good to add a loading status to the InputFields and SelectField. Use cases could be: validating InputField input remotely, fetching options for SelectField, fetching options for some AutoComplete component (based on InputField). As such this loading state should when disabled (validating or fetching select options) and when enabled (fetching option for TypeAhead). I think this would classify as a valid state, because it is also mutually exclusive to all the other states.

Also good point!

> 
>     4. Checkboxes/Switches: I don’t think these components themselves need to have a status property, but we could keep them there for a consistent API. What I am missing though, is the possibility to add a help text. SelectField and InputField do have this, so I would expect it here too. It would be more consistent and would be of practical use.
> 
>     5. Icons: I see quite a few icons here. There are also quite a few in the HeaderBar repo. Are these all going to be consolidated into one icons repo?

The idea so far has been to embed the icons which are _actually_ used by components and to keep those icons internal to the library. Eventually we might look at a `ui-icons` library as well.

varl:
c09e238 adds loading states for input fields

varl:
Added Portal for DropMenu

varl:
The Help component is pulled out from InputField, SelectField, and used as a sibling instead, e.g.

```jsx
         <>
             <Checkbox name="Ex" label="Checkbox" onChange={noop} />
             <Help>A lil' help text</Help>
         </>
```

varl:
Also moving to this style of API:

```jsx
<Button destructive small>
    Label me!
</Button>
```

Mohammer5:
@varl I just saw that you've extracted the help text into a `Help` component.
I like that, unified help text styles across all components.

But not using them inside our components directly has meddled with the alignments.
Take the `InputField` for example. The help text is not aligned with the label or the input text anymore, so it'd be up to the app to fix those styles. I think that's the wrong approach.
The InputField component needs different spacing values for the Help text than the Checkbox for example.

So, I think we should keep the `Help` component on the 1st level and expose it, but our components that can be used with a help text should use them internally as well.

varl:
>The InputField component needs different spacing values for the Help text than the Checkbox for example.

Generally I think this looks OK:

image

image

The main thing I was trying to avoid was to expand the API by adding another prop.

Mohammer5:
@varl The Help text and the label text are not aligned. Of course it looks "ok" in terms of usability.
But if we keep this strict in the `ui-core` library, we'll follow the alignment principle.

The reason why I want this is that, although not required to be functional, this will reduce the noise the user will unconsciously feel when using our apps. These little disturbances will make the app feel uncomfortable. And as we're at the core of all our apps (in theory...), all apps will inherently benefit from us doing it right.

> The main thing I was trying to avoid was to expand the API by adding another prop.

I totally get this point and agree, we should try to keep the api of the componenets as simple and small as possible. But the help texts are an essential part of form components, so I'm wondering what will happen if we leave this out.

I think that each app will then either create it's own component including the help text component (which will introduce inconsistency across apps), or do it individually for each component, which results in even more inconsistency.

So maybe there's a third way I'm not seeing right now? Others might whey in and share their opinions in regards to this. From my point of view, I think that the `Help` text is essential for form input elements and will be used in many cases anyways.

varl:
> @varl The Help text and the label text are not aligned. Of course it looks "ok" in terms of usability.
> But if we keep this strict in the `ui-core` library, we'll follow the alignment principle.

Is the alignment principle that it should be aligned to the label?

Mohammer5:
@varl The alignment principle is that more or less everything should be align, regardless of whether they're inside the same component or not. During my first job, the designers called it "optical line", I didn't find anything about it, so it might be the wrong term, but it describes exactly what I mean.

It makes a lot of sense to the eye if everything starts on the same horizontal line, giving the app a feeling of unity rather than being scattered into multiple parts.

@cooper-joe should whey in on this, as he's the one who will finally have to decide if this is correct or important after all!

cooper-joe:
My recommendation for this is to apply 16px padding-left to all help text. This is a compromise that I feel achieves the best of the different approaches:
- 16px padding left aligns help text with text-input labels, enough to create a scannable vertical rhythm on text-input heavy forms
- with radio/switch/check 16px does not align, but it would take very different values to make a label/help alignment here. Adapting padding to each element would negate any sense of vertical rhythm.
- why not 0px padding left on all help text? Indenting the text helps the user to understand that the help text applies to the element that comes directly over it. Indentation reinforces parent>child relationship.
- a consistent padding value allows for all elements to use the help text without needing to adapt/ask a designer. For example, if someone wanted to add help text under an 'org.unit selector', it would be easy to do that and consistent with all other help text.

Mohammer5:
Sounds very good to me! Especially as all input elements, except checkboxes and radio buttons will have 16px inner spacing on the left anyways!

varl:
Proposed commit message which will trigger version 2.0.0.

```
feat: version 2.0.0 of ui-core

This release does not contain any new components over v1, but it does manage to change almost everything.

Features include:

- simpler components
- better composition
- better isolation from CSS leakage
- consistent component APIs
- better documentation
- better development tools when working on components

BREAKING CHANGE: The component API has been revamped to be simpler and more consistent. See the documentation for a full list.
```

Mohammer5:
> - better documentation

@varl This part is completely missing right now. Is there a way we can easily auto-generate markdown files for each component containing the properties, whether they're required or not, their type and default value? I big plus would be to be able to edit that document to add custom text while being able to update the props table when changing the component without losing the modifications!

varl:
> 
> 
> > * better documentation
> 
> @varl This part is completely missing right now. Is there a way we can easily auto-generate markdown files for each component containing the properties, whether they're required or not, their type and default value? I big plus would be to be able to edit that document to add custom text while being able to update the props table when changing the component without losing the modifications!

What I mean by "better documentation" is:

- We are using react-docgen in a Storybook plugin to get the prop tables in there:
image

- We have also integrated dhis2/design-system docs where relevant:
image

We can do even more, like adding react-docgen strings to each prop to give them a description.

Mohammer5:
@varl Oh nice! I didn't notice that, I just clicked on `Notes`...
Disregard my PR then! I like this way more

PR #190, opened 2019-04-29
docs(missing components): create list of all missing components of d2-ui

Mohammer5:
I added a doc with all the components of `d2-ui` which I think are missing.

There are some components that shouldn't be part of `ui-core`, I've added a comment to those.
Furthermore there are also some components of which I think that they're not needed, they either look like components that are only used in one place (`Heading`?) or add unnecessary complexity (like the layout components).

If you agree with my thoughts about which components not to add to `ui-core`, I'll remove them from the list.


varl:
Added an issue instead #225.

PR #192, opened 2019-05-01
feat(drop-menu): compute position relative to anchor element

HendrikThePendric:
### Context
This PR ties in to the comment I made here https://github.com/dhis2/ui-core/pull/183#pullrequestreview-224460229:
> SplitButton and DropDownButton use a very simple div with position absolute to render stuff into a “Popover-menu”. This approach is fragile: there are going to be situations where the menu isn’t rendered fully if the container has overflow-y hidden, and also potential z-index problems if parents also have certain z-index properties (a child's z-index is set to the same stacking index as its parent). It would be better to render stuff in a portal (React.createPortal) as the last child of the document body. Stuff does become more complex this way because the position will have to be computed, but it will be more robust.

Based on this comment, the DropMenu was created, which addressed the concerns. However, the position of this component would quite likely be incorrect. In fact the only scenario in which the position would be correct is if the button is the last child of the body without any margin.

### Solution
`DropMenu` now has an additional prop `anchorEl` which represents a DOM node against which the `DropMenu` can position itself, and a function `getPosition` that computes the correct top and left position fro the `DropMenu`.

### Some notes on using refs
In essence, I think that the current implementation (i.e. as last child of the body and absolutely positioned against an anchor) is the mosts robust way of showing a "popover". However, it does mean that every component that uses the `DropMenu` must be able to provide a ref to a DOM node. This means using refs on the anchor element/component. This should always be possible, but there are several scenarios with different solutions:
- If the anchor a regular DOM node: no problem, just add a ref
- If the anchor is a class based React component: you can add a ref to that, but it will just return the component-instance, not a DOM node. To get an actual DOM node, you have two choices:
   - Add a ref for the DOM node to the component instance, so it can be read this way (recommended)
   - Leave the component as is and use `ReactDOM.findDOMNode` (deprecated in StrictMode)
- If the anchor is a stateless/functional component, you can wrap this component in a call to `React.forwardRef` ([see docs](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-to-dom-components))
- In scenarios where you can't or don't want to change the anchor-component's code, you can always add a wrapper element...

There are several way's of working with refs. For example:
- When using callback refs the value of the ref is a DOM node
- When using `React.createRef()` the value of the ref is an object with a `.current` property, which is the DOM node.

Because of these differences, I decided it would be better that the `DropMenu` got an additional `anchorEl` prop, instead of a `anchorRef` prop. This way the component that renders `DropMenu` can decide how it want to work with refs, as long as it just passes the actual referenced DOM node to the `DropMenu`

### DIFF
--- a/src/Button/index.js
+++ b/src/Button/index.js
@@ -5,48 +5,42 @@ import propTypes from 'prop-types'
 
 import styles from './styles.js'
 
-const Button = React.forwardRef(
-    (
-        {
-            type,
-            children,
-            icon,
-            name,
-            value,
-            disabled,
-            onClick,
-            className,
+const Button = ({
+    type,
+    children,
+    icon,
+    name,
+    value,
+    disabled,
+    onClick,
+    className,
+    primary,
+    secondary,
+    destructive,
+    small,
+    large,
+}) => (
+    <button
+        disabled={disabled}
+        onClick={evt => onClick && onClick(name, value)}
+        className={cx(className, {
             primary,
             secondary,
             destructive,
             small,
             large,
-        },
-        ref
-    ) => (
-        <button
-            disabled={disabled}
-            onClick={evt => onClick && onClick(name, value)}
-            className={cx(className, {
-                primary,
-                secondary,
-                destructive,
-                small,
-                large,
-                'icon-only': icon && !children,
-                icon,
-            })}
-            type={type}
-            name={name}
-            value={value}
-            ref={ref}
-        >
-            {icon && <span className="button-icon">{icon}</span>}
-            {children}
+            'icon-only': icon && !children,
+            icon,
+        })}
+        type={type}
+        name={name}
+        value={value}
+    >
+        {icon && <span className="button-icon">{icon}</span>}
+        {children}
 
-            <style jsx>{styles}</style>
-        </button>
-    )
+        <style jsx>{styles}</style>
+    </button>
 )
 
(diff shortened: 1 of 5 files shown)


Mohammer5:
Looks good, but there's no adjustment when the viewport is resized.
Could be useful for devices that can change orientation.

HendrikThePendric:
https://github.com/dhis2/ui-core/pull/192/commits/da072e573d27ea47b80e63bacd742ef999aba5bb adds a resize listener. Perhaps this should be debounced.... I wanted to keep the code simple, and not that many computations are done to get the position, so I think it's OK to keep it like this. Let me know if you think I should add a debounce.

Mohammer5:
> Let me know if you think I should add a debounce.

@HendrikThePendric I think it's fine this way, it's not a heavy computation, should be fine.
Normally no one just resizes the browser like crazy. Orientation changes will be 1 resize event only.

varl:
> [da072e5](https://github.com/dhis2/ui-core/commit/da072e573d27ea47b80e63bacd742ef999aba5bb) adds a resize listener. Perhaps this should be debounced.... I wanted to keep the code simple, and not that many computations are done to get the position, so I think it's OK to keep it like this. Let me know if you think I should add a debounce.

It should only run when the DropMenu is mounted, so I think it's fine to leave it as it is now until it becomes a problem.

PR #195, opened 2019-05-03
refactor(checkbox): extract nested elements into their own components

Mohammer5:
I have extracted the components rendered by the Checkbox component into individual components,
each with their own style which is the styled-jsx way.
This improved the overall readability of the `Checkbox` component's code.

Note:
I've set `indeterminate` to `false` in the `onChange` callback function.
The value is only used to check whether it's falsy or not, so there's no difference between `false` and null.
Maybe I didn't get it right, but the condition in the onChange tells me, that it'll be always falsy after the first onChange?


varl:
The indeterminate logic is a bit annoying.

It's something like:

**Indeterminate A:**

 [-] A
- - [x] B
- - [ ] C

**A is checked:**

- [x] A
- - [x] B
- - [x] C

**B is unchecked, A becomes indeterminate.**

 [-] A
- - [ ] B
- - [x] C

So a checkbox is indeterminate until the first action is taken on it, but it can become indeterminate again if a checkbox in its group becomes unchecked. But that has to be handled from the outside as each checkbox doesn't know about their group children.

varl:
So we have a `src/Input` folder which contains the `Input` component for the `InputField`. Should we move `src/Input/index.js` to `src/InputField/Input.js` or should we reorganize `src/Input` to accommodate more inputs?

- `src/Input/Text.js`
- `src/Input/Checkbox.js`
- `src/Input/Radio.js`

Mohammer5:
> [...], but it can become indeterminate again if a checkbox in its group becomes unchecked. But that has to be handled from the outside as each checkbox doesn't know about their group children.

I just checked the version of the [checkbox of the next branch'es HEAD](https://github.com/dhis2/ui-core/blob/next/src/Checkbox/index.js). There's no way the Checkbox can change back to indeterminate, there's no `componentDidUpdate` the will reset the state, depending on the property, it only happens once during construction.

Should I add that behaviour? It will mean that the implementing Component will have to take care of that state though.

> So we have a src/Input folder which contains the Input component for the InputField. Should we move src/Input/index.js to src/InputField/Input.js or should we reorganize src/Input to accommodate more inputs?

Very good question! Right now, I'm not even sure if we should expose them at all as they all have styling which works only within our `ui-core` components. Maybe we move them back into the folder of the components using them until they're reused by multiple components. What do you think?

varl:
>Maybe we move them back into the folder of the components using them until they're reused by multiple components.

Yeah, I think that's probably best for now.

varl:
>Should I add that behaviour? It will mean that the implementing Component will have to take care of that state though.

Could we let `indeterminate` be a prop controlled by the implementing Component and avoid handling the state for it internally?

Mohammer5:
@varl Agree, that sounds like the simplest, yet best solution.
Will add it to this PR

I think the other thing, with the inputs, should be part of another PR.

varl:
Agreed, approved with the `indeterminate` prop. Nice work. 🚀

Mohammer5:
@varl I just realized that the other state value is `defaultChecked`.
In my opinion, this could be removed as well, as the initial form data should handle this, just like the initial value for a text input field would be provided via the `value` prop as well. Then this component could be 100% functional. **That would be a breaking change though!**

varl:
>That would be a breaking change though!

That's why we should do it for v2.

PR #197, opened 2019-05-03
refactor: make fields controlled

varl:
The Switch, Checkbox, and Radio could easily be converted to functional components just based on props, which is nice.

### DIFF
--- a/src/InputField/index.js
+++ b/src/InputField/index.js
@@ -11,23 +11,19 @@ class InputField extends React.Component {
         focus: this.props.focus,
     }
 
-    onFocus = evt => {
+    onFocus = e => {
         this.setState({ focus: true })
-        this.props.onFocus(this.props.name, evt.target.value)
+        this.props.onFocus(e)
     }
 
-    onBlur = evt => {
+    onBlur = e => {
         this.setState({ focus: false })
-        this.props.onBlur(this.props.name, evt.target.value)
-    }
-
-    onChange = evt => {
-        this.setState({ value: evt.target.value })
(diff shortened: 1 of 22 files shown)


src/InputField/index.js
```diff
@@ -9,7 +9,6 @@ import { theme } from '../theme'
 class InputField extends React.Component {
     state = {
         focus: this.props.focus,
```
Mohammer5: This could arguable be removed as well. But I think for now it's fine as it's used for styling purposes our apps shouldn't care about

src/SelectField/Select.js
```diff
@@ -99,8 +100,9 @@ export class Select extends Component {
 }
 
 Select.propTypes = {
-    value: propTypes.string,
+    name: propTypes.string.isRequired,
     onChange: propTypes.func.isRequired,
+    value: propTypes.string,
```
Mohammer5: This could be made required. Empty string would suffice that condition.
But I don't care too much about it

src/SelectField/index.js
```diff
@@ -90,11 +85,12 @@ class SelectField extends React.Component {
                 dense={dense}
             >
                 <Select
+                    name={name}
                     value={value}
                     disabled={disabled}
                     filled={filled}
                     dense={dense}
-                    onChange={this.onChange}
+                    onChange={() => onChange(name, value)}
```
Mohammer5: I'm not sure if the `value` of this one will be the new value? I think it'll be the old value.
But see the main comment regarding this `onChange` style.

PR #199, opened 2019-05-05
 feat(gen component docs): add script to generate component docs

Mohammer5:
I've created a small script that generated docs for our publicly exposed components.
It could easily be added to a hook (pre commit, pre publish?) to generate these docs on the fly.

(I'd go for precommit as the generation of the docs might throw errors, could be done silently though unless there's an error thrown. Then we could do the docs creation in the prepublishOnly npm hook)


Mohammer5:
Already existing through storybook plugin

PR #204, opened 2019-05-06
fix: styled-jsx needs to be a peer dep for all libs

varl:
Because of https://github.com/zeit/styled-jsx/issues/64 styled-jsx needs
to be a peer dependency for all libraries, and pulled in once by the
App, so the instance can be shared.


varl:
#204 will resolve this.

PR #206, opened 2019-05-07
fix: setup dependencies based on app-runtime

varl:


package.json
```diff
@@ -44,18 +44,19 @@
     "husky": "^2.0.0",
     "react-dev-utils": "^9.0.0",
     "size-limit": "^1.0.0",
+    "styled-jsx": "^3.2.1",
```
varl: Yarn doesn't install peer dependencies, so we need to state it as a dev dependency to get local development to work, both for Storybook but also for running the build process.

PR #207, opened 2019-05-08
refactor(chip): extract sub components into styled-jsx style components

Mohammer5:
I'm creating this PR as a draft PR because I'm not sure this is the right direction for these components.
It's just a gut feeling, but it feels little weird, splitting the components like this just to apply certain styles, I lack a better alternative that matches our criteria right now though..
Anyways, this PR decreases the complexity of the `Chip` component itself.

In the second commit (2e72e83) I removed some style rules that haven't been used, there's no `.icon` class anywhere. I guess this is a leftover from when @varl switched to styled-jsx, as the icon components (passed to `Chip` via the `icon` prop) could have the `.icon` class name. But with the styled-jsx approach this style wouldn't be applied anyways. How should we handle that style? In my opinion the style should be applied to the icon component by the app itself rather than by this library, so there wouldn't be a need to change the current implementation of `Chip`


src/Chip/Remove.js
```diff
@@ -29,7 +30,7 @@ export class Remove extends PureComponent {
     render() {
         const onRemove = this.props.onRemove
 
-        if (!onRemove) return
+        if (!onRemove) return null
```
varl: Why `null` instead of `undefined`?

Mohammer5: I think it doesn't really make a difference, but interestingly I wanted to know if there's any official docs about this (I usually returned undefined so far) and found this:
https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

So I just copied that

varl: Oh, that's interesting, it _does_ affect the CodePen linked: https://codepen.io/gaearon/pen/Xjoqwm?editors=0010

Try switching the `return null` to `return`. Then hit the `Hide` button. Now it's gone completely whereas if you return `null` the button is still there. React treats them differently, one stops rendering and one doesn't.

>Returning null from a component’s render method does not affect the firing of the component’s lifecycle methods. For instance `componentDidUpdate` will still be called.

Mohammer5: > Now it's gone completely whereas if you return null the button is still there.

Huh... That's VERY weird behavior, will never use undefined again!

PR #210, opened 2019-05-10
feat(checkbox): accept children as an alternative to the label prop

Mohammer5:
This PR prepares the checkbox component to be used for the org unit tree widget.
The `Checkbox` component now accepts contents either via label prop or via children, by exposing the `Label` component as `Checkbox.Label`.

This way, more items can be added that are aligned with the label text automatically, which do not trigger the checkbox to be (de-)selected when interacted with.

In order to achieve this, I changed the container element from `label` to `div`, changed the `Label`'s element from `span` to `label` and added the `htmlFor` property which uses the name that's provided to the Checkbox component.

A story to demonstrate this way of constructing a checkbox has been added as well.

### DIFF
--- a/src/Checkbox/Icon.js
+++ b/src/Checkbox/Icon.js
@@ -11,6 +11,7 @@ const icons = resolve`
         height: 24px;
         width: 24px;
         fill: ${theme.default};
+        cursor: pointer;
     }
 
     .checked {
@@ -19,6 +20,7 @@ const icons = resolve`
 
     .disabled {
         fill: ${theme.disabled};
+        cursor: not-allowed;
     }
 
     .error {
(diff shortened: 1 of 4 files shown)


Mohammer5:
@HendrikThePendric @varl That's a very good argument!
I think it might be better to switch to children completely in this case then. It's what we're doing in other components as well PLUS gives us more freedom of how we want to use the checkbox.

That way there'd still be the possibility of having nested checkboxes.. but I think that's something no one should/would do, just as no one will have `a`s inside an `a` (or same with `label`s). At least none of the devs involved with dhis2.

varl:
I don't understand why a Checkbox would ever take `children`? :thinking:

varl:
Is the Tree component based off of this branch? Maybe I can look there?

Mohammer5:
@varl @HendrikThePendric I had a night to sleep over this and I think you're right. This can be handled entirely by the org unit tree widget, we don't need to make to Checkbox component overly complicated.

Note: 
:+1: Very nice team work, this way we'll keep unnecessary stuff out of the code!

PR #212, opened 2019-05-12
fix: required star

teug91:
Added required star for Outlined SelectField and InputField

### DIFF
--- a/src/FieldLabel/Outlined/Label.js
+++ b/src/FieldLabel/Outlined/Label.js
@@ -64,6 +64,10 @@ const styles = css`
         position: relative;
     }
 
+    .required span::after {
+        content: '*';
+    }
+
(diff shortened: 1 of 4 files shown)


src/FieldLabel/Outlined/Label.js
```diff
@@ -64,6 +64,10 @@ const styles = css`
         position: relative;
     }
 
+    .required span::after {
+        content: '*';
```
Mohammer5: The other components that add the asterisk also have a space between the label and the asterisk, please add that.

teug91:
Added spacing for the pre-existing Filled SelectField and InputField as well.

varl:
I think the status check is failing for a reason unrelated to the PR. Merging.

PR #215, opened 2019-05-13
feat: node component

Mohammer5:
This will add a very fundamental tree component.
It's not final yet as the specs are work in progress, but it shouldn't be too much to change, especially as this contains only the very basic functionality.

@varl I tried to add nice/shiny behavior by using transitions to open/close the content of a tree or rotate the arrow, but came to the conclusion that it adds more distraction and complexity, plus we aren't trying to sell anything, so we don't need to offer super shiny extra nice UI. Or do you see this differently?
(maybe @cooper-joe wants to have a say in this also)


varl:
We have a couple of Tree components floating around, e.g. this one: https://github.com/dhis2/import-export-app/tree/master/src/components/Tree

Have you taken a look at it?

varl:
This looks pretty slick. One thing I don't like is that the State component is the only one which is exposed in the "JSX props" tab in Storybook so it's hard to get an idea of the usage through the docs. Anything we can do to fix that?

Mohammer5:
> Have you taken a look at it?

@varl No, haven't seen this one before, but it doesn't seem to be so different, except it stores the "open" state itself and has a "checked" state for each node as well. Looks a bit too complicated for a simple tree component.

Right now I think that a controlled Tree component makes most sense so that the widgets will have full control over what happens (like the yet-to-be-implemented org unit tree). Also seems like the whole `ui-core` lib is heading this way.

> One thing I don't like is that the State component

This one I did only for development purposes. I want to remove that one once the Tree component is done and add static stories for the different states it can have.

> so I am in favor of leaving them out for the sake of speed and simplicity.

@cooper-joe Thanks!

Mohammer5:
@varl I think this Tree component is kind of ready? (Once approved I'll remove the dev stories)

varl:
Does the `Tree` use the `Checkbox` components? Why is the checkboxes in the `Tree` missing the focus state you added to other form components?

varl:
2019-05-30-162823_627x460_scrot

What is the "dashed" vertical line before each child?

Mohammer5:
@varl The focus stuff hasn't been merged yet ([PR 230](https://github.com/dhis2/ui-core/pull/230)).

Checkboxes are not part of the tree, they're just used to demo the tree component in the story.

Have to fix the lines bug

src/Tree/index.js
```diff
@@ -0,0 +1,139 @@
...
+export const Tree = ({
+    children,
+    hasLeafes,
+    open,
+    onToggleOpen,
+    arrowTopOffset,
+}) => {
+    const className = cx('tree', {
+        open,
+        'has-leafes': hasLeafes,
```
varl: Does "leafes" refer to leaf nodes? Plural of "leaf" (🍃) is "leaves".

src/Tree/index.js
```diff
@@ -0,0 +1,139 @@
...
+    })
+
+    return (
+        <div className={className}>
+            <Arrow
+                open={open}
+                hasLeafes={hasLeafes}
+                onToggleOpen={onToggleOpen}
```
varl: Does this fire on toggle or on open? I would prefer either `onToggle` or `onOpen`, respectively.

src/Tree/index.js
```diff
@@ -0,0 +1,139 @@
...
+                arrowTopOffset={arrowTopOffset}
```
varl: Is this for internal (to the component) or external (to the developer) use?

varl:
Good work! Demo looks good, will take another look in the morning. 👍

src/Tree/index.js
varl: Looking at it now, I think it should be `onClick`.

Mohammer5: well it's basically toggling the open state.. one could assume that `onToggle` refers to the expanded/collapsed state, so I guess it might work, but `onToggleOpen` is def. more precise

src/Tree/index.js
Mohammer5: I thought about a way to push down the arrow if the tree content is larger. But `arrowTopOffset` was a stupid idea.. I replaced it with giving the internal proper class names, now the app can just hook into that (I provided a story as an example how to do that)

src/Tree/index.js
varl: I see what you mean about it toggling a state.

The name `onToggleOpen` implies to me that it only fires when it is opened, but when I check the implementation it's bound to the `onClick` handler so it will fire every time; not just when it goes from closed to open. That's what I find confusing.

varl: Either way, might change now after the discussion on Slack anyway so we'll see how it flies. :)

Mohammer5:
@varl and I have talked a bit about the implementation details. We came to the conclusion to simplify the api by providing the label via a render prop, allowing the Tree component to check whether is has leaves or not by checking if there are any children.

We then decided to rename the `Tree` component to `Node` as the `Node` doesn't know anything about it's children (if they're more `Node`s or something entirely different).

Only "downside" of the render prop is that passing `PureComponent`s to the render prop will mean they will be re-rendered all the time.

src/Tree/index.js
Mohammer5: I think we could split it into `onExpand` and `onCollapse`, the names are much better and allow for finer control, I'd say

varl: Splitting them sounds good to me. We use `onOpen` and `onClose` else-where  in `ui-core`, would those terms be appropriate here too? 

If possible I want to make the API be as consistent as possible and not use new terminology in each component.

Mohammer5:
@varl The `onToggleOpen` callback has been split into `onOpen` and `onClose`

varl:
@Mohammer5 nice!

Mohammer5:
I just discussed with @varl that our current standard is to accept strings only when using the `label` prop, so I changed it to `component`, to be consistent with the other components, like the `DropdownButton` or `SplitButton` components.

varl:
Good work! Want to do the honors of merging, @Mohammer5?

Mohammer5:
@varl thanks, le done

PR #219, opened 2019-05-13
docs: document gotcha for global rules bleeding into components

### LINKED ISSUE #213: Generic class names are causing global styles to be applied in some cases
Specific case: disabled Menu item
The disabled menu item has class `disabled`. This results in some global styles being applied. The global styles are coming from some old analytics plugins (used for generating pivot tables event reports, etc). We are hesitant to change these old plugins and wonder if ui-core can take a different approach to avoid having generic classnames. Here is a screenshot of some applied styles:

image

varl:
Resolves #213.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,3 +67,70 @@ automatically starts.
 It is very important that the commit that lands on **master** follows
 the conventional commit format, since that is what is used to
 automatically determine the next version.
+
+## FAQ
+
+#### How to avoid a global style rule from affecting a ui-core component?
+
+The best practice is that each component has styles which are scoped to
+that component to avoid style rules that leak out from one component and
+bleeds into another component.
+
+Generally all the CSS-in-JS solutions grant this out of the box, and
+this is only a problem when using global stylesheets, e.g. by importing
+`index.css` in an application and adding global rules to it.
+
+`index.css`:
+```
+div {
+    padding: 20px;
+}
+
+.disabled {
+    opacity: 0.5;    
+}
+```
+
+These styles will bleed into all components that use `div` or
+`.disabled` and doesn't itself set those rules and win by specificity.
+
+It becomes a real problem if the rules above does something
+`!important`.
+
+`index.css`:
+```
+div {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,3 +68,77 @@ automatically starts.
 It is very important that the commit that lands on **master** follows
 the conventional commit format, since that is what is used to
 automatically determine the next version.
+
+**Never push straight to master, always go through a PR!**
```
Mohammer5: We should be able to lock master so that no one can push directly anyways.
Is there a reason why we're not using this feature?

varl: The @dhis2-bot needs to push to master when it cuts the release, that's why it's not locked. I do think it's locked for non-admins though. This is just there for clarity.

varl:
> You've offered a way to fix the styles that are wrongly applied to other elements. The solution includes a class name that has to be passed down, which I think is not ideal.

That's true.
 
> Let's say we have a component somewhere in the code like this:
> 
> ```js-jsx
> <div>
>     {...}
>     <Component disabled />
> </div>
> ```
> 
> And there are styles that change the style for `.disabled` and `div` which affect the component.
> I think the code should be changed to this:
> 
> ```js-jsx
> <div>
>     {...}
>     <div className="fix-container">
>         <Component disabled />
>     </div>
> </div>
> ```
> 
> ```css
> .fix-container .disabled {
>     /* will undo the overriding of global styles for this component */
>     opacity: 1 !important;
> }
> ```
> 
> Then all styles can be overriden, also for nested elements inside `Component`, which wouldn't be accessible through the `className` prop. Also the code for using the component itself doesn't need to be changed. Removing a container once the bleeding-in styles have been removed should be a but easier, could be done through git as well (even if the implementation of `Component` itself has changed) by stripping out the fixing commit (might be easier to do if there are a lot of changes to prevent styles from bleeding in)

Nice! Excellent workaround for the cascading problem.

varl:
@Mohammer5 updated with your workaround.

PR #222, opened 2019-05-15
fix(input field and select field): crop long labels

Mohammer5:
Title explains everything


varl:
Nice, this behaves as I would expect. :+1:

PR #223, opened 2019-05-15
feat: use a portal to overlay the loader

### LINKED ISSUE #217: The CircularProgress component is behind other elements on the page
I'm using this markup:
`<CircularProgress overlay>`

it is showing up behind the other elements. In the screenshot, you can just barely see the blue circle behind the dashboard items

image

varl:
Adds a Portal to the Overlay so its always rendered on top of the other components.

Removes the "overlay a single component" for now. It's not used anywhere and the app can easily build this using the component as is.

Fixes #217.

### DIFF
--- a/src/ComponentCover/index.js
+++ b/src/ComponentCover/index.js
@@ -4,7 +4,7 @@ import css from 'styled-jsx/css'
 import cx from 'classnames'
 
 const ComponentCover = ({ children, className }) => (
-    <div className={cx('component-cover', className)}>
+    <div className={className}>
         {children}
         <style jsx>{`
             div {
--- a/src/ComponentCover/index.js
+++ b/src/ComponentCover/index.js
@@ -3,8 +3,8 @@ import propTypes from 'prop-types'
 import css from 'styled-jsx/css'
 import cx from 'classnames'
 
-const ComponentCover = ({ children }) => (
-    <div className="component-cover">
+const ComponentCover = ({ children, className }) => (
+    <div className={cx('component-cover', className)}>
         {children}
         <style jsx>{`
             div {
@@ -12,12 +12,10 @@ const ComponentCover = ({ children }) => (
                 align-items: center;
                 justify-content: center;
 
-                position: relative;
-                left: 0;
-                top: 0;
+                position: absolute;
 
-                height: 100%;
-                width: 100%;
+                height: inherit;
+                width: inherit;
 
                 z-index: 900;
                 background: rgba(200, 200, 200, 0.6);
@@ -28,6 +26,7 @@ const ComponentCover = ({ children }) => (
 
 ComponentCover.propTypes = {
     children: propTypes.node,
+    className: propTypes.string,
 }
 
 export { ComponentCover }
--- a/src/ComponentCover/index.js
+++ b/src/ComponentCover/index.js
@@ -1,0 +1,32 @@
+import React, { Fragment } from 'react'
+import propTypes from 'prop-types'
+import css from 'styled-jsx/css'
+import cx from 'classnames'
+
+const ComponentCover = ({ onClick, children }) => (
+    <div className="component-cover" onClick={onClick}>
+        {children}
+        <style jsx>{`
+            div {
+                display: flex;
+                align-items: center;
+                justify-content: center;
+
+                position: relative;
+                left: 0;
+                top: 0;
+
+                height: 100%;
+                width: 100%;
+
+                background: rgba(200, 200, 200, 0.6);
+            }
+        `}</style>
+    </div>
+)
+
+ComponentCover.propTypes = {
+    onClick: propTypes.func,
+}
+
+export { ComponentCover }
--- a/src/ComponentCover/index.js
+++ b/src/ComponentCover/index.js
@@ -19,6 +19,7 @@ const ComponentCover = ({ children }) => (
                 height: 100%;
                 width: 100%;
 
+                z-index: 900;
                 background: rgba(200, 200, 200, 0.6);
             }
         `}</style>
--- a/src/ComponentCover/index.js
+++ b/src/ComponentCover/index.js
@@ -3,8 +3,8 @@ import propTypes from 'prop-types'
 import css from 'styled-jsx/css'
 import cx from 'classnames'
 
-const ComponentCover = ({ onClick, children }) => (
-    <div className="component-cover" onClick={onClick}>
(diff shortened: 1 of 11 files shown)


Mohammer5:
@varl I just asked @jenniferarnesen the following:

> Do you know if the Overlay is only used to cover the whole page? or is it used inside some other components as well to display loading state?

If they're used to display a loading state inside a container, this fix will def. break those components.

varl:
> @varl I just asked @jenniferarnesen the following:
> 
> > Do you know if the Overlay is only used to cover the whole page? or is it used inside some other components as well to display loading state?
> 
> If they're used to display a loading state inside a container, this fix will def. break those components.

Yes, I'm aware. Nothing is using the component except for Dashboards: https://github.com/search?l=JavaScript&q=org%3Adhis2+org%3Adhis2+circularprogress&type=Code

It's not a feature available in the old circular progress either.

I think I would prefer if we could do something like this, and not have the `overlay` prop at all for the CircularProgress component:

```js-jsx
// to cover the entire page
<ScreenCover>
    <CircularProgress />
</ScreenCover>

// to cover a component
<ComponentCover>
    <CircularProgress />
</ComponentCover>
```

Mohammer5:
@varl I like that idea! You want to change it yourself?

varl:
This is technically a breaking change, but since it is not used anywhere yet we can sneak this in as a feature.

src/ScreenCover/index.js
```diff
@@ -1,23 +1,61 @@
...
+    componentDidMount() {
+        document.addEventListener('click', this.onDocClick)
+    }
+
+    componentWillUnmount() {
+        document.removeEventListener('click', this.onDocClick)
+    }
+
+    onDocClick = evt => {
+        console.log(this.elContainer, evt)
```
HendrikThePendric: Oops

src/ScreenCover/index.js
```diff
@@ -1,23 +1,61 @@
...
+            <div className="screen-cover" ref={this.elContainer}>
+                <div className="children">{this.props.children}</div>
+
+                <style jsx>{`
+                    .screen-cover {
+                        display: flex;
+                        align-items: center;
+                        justify-content: center;
+
+                        position: fixed;
```
HendrikThePendric: The container needs a z-index methinks. Perhaps the children don't... But the `.screen-cover` does.

src/ScreenCover/index.js
```diff
@@ -1,23 +1,61 @@
...
+            this.elContainer.current &&
+            this.elContainer.current === evt.target
+        ) {
+            this.props.onClick()
+        }
+    }
+
+    render() {
+        return (
+            <div className="screen-cover" ref={this.elContainer}>
```
HendrikThePendric: Suppose we'd change this line to:
```html
<div className="screen-cover" onClick={this.props.onClick}>
```
Wouldn't that have the same effect as a click on the document? (considering the `div.screen-cover` will in fact cover the whole screen) And then this could be a stateless component 🎉 

But what if this `ScreenCover` has a loader in it and we want `props.onClick` to get triggered when clicking on that too? Then the loader should get `pointer-events: none;`. I think this would make sense.

src/ComponentCover/index.js
```diff
@@ -0,0 +1,32 @@
...
+                align-items: center;
+                justify-content: center;
+
+                position: relative;
+                left: 0;
+                top: 0;
+
+                height: 100%;
+                width: 100%;
+
```
HendrikThePendric: I think we should add a z-index here too...

src/ScreenCover/index.js
varl: The `ScreenCover` is used for the `Modal` too, and using `pointer-events: none;` killed the interactions with the `Modal` itself (including the action buttons), so that felt a bit heavy handed.

src/ScreenCover/index.js
varl: b8697ec

src/ScreenCover/index.js
varl: 6f6b48b

src/ComponentCover/index.js
varl: c341aec

varl:
We are quickly moving into the territory of needing a documented standard for what goes in what range of `z-index`.

Currenly I've followed:

- ScreenCover has `10000` because it should cover everything
- DropMenu has `1000`
- ComponentCover has `900` to place it below dropdown menus but above normal stacking index.

jenniferarnesen:
It's not working quite right yet. But maybe I am not using it correctly?

Here is the dom:
image


And the result:

image

src/ScreenCover/index.js
HendrikThePendric: > killed the interactions with the Modal itself (including the action buttons)

Aha.... interesting. That could have to do with event propagation. A slightly different markup could fix things, something like this:
```jsx
<div className="screen-cover">
    <div className="backdrop"></div>
    { children && <div className="children">{this.props.children}</div>}
</div>
```
Because the backdrop is a separate div, the events in the children won't clash with the events in the backdrop.

varl: b0d8923

src/ScreenCover/index.js
```diff
@@ -1,18 +1,40 @@
...
 import css from 'styled-jsx/css'
 import cx from 'classnames'
 
-export const ScreenCover = ({ onClick }) => (
-    <div className="screen-cover" onClick={onClick}>
+const ScreenCover = ({ children, onClick }) => (
+    <div className="screen-cover">
+        <div className="backdrop" onClick={onClick} />
+
+        <div className="children">{children}</div>
```
Mohammer5: You could extract the backdrop and children into their own components to avoid the classname

```jsx
const Backdrop = ({ onClick }) => (
  <div onClick={onClick}>
    <style jsx>{`
      div {
        height: 100%;
        width: 100%;
        background: rgba(200, 200, 200, 0.6);
      }
    `}</style>
  </div>
)

const Children = ({ children }) => (
  <div>
    <style jsx>{`
      div {
        position: absolute;
        top: 50%;
        left: 50%;
        width: auto;
        height: auto;
        transform: translate(-50% -50%);
      }
    `}</style>
  </div>
)
```

And then:
```jsx
<Backdrop onClick={onClick} />
<Children children={children} />
```

varl: 6357383

PR #224, opened 2019-05-16
feat: add Pop component with stories and docs

Mohammer5:
This is jus a PoC of the Popover.
The storybook stories are just for development, once this component is done, they will be replaced with static stories that display the various states the Popover can have.

Questions in my mind:
1. Is this a `ui-core` component or ` ui-widget` component?

So far it:
1. allows limitless amount of levels
1. never renders off screen

Todo:

- [ ] Serious refactoring
- [ ] RTL?
- [ ] Proper Storybook stories


varl:
This looks like a `DropMenu` + `Menu` component to me. Could we integrate this into those? Or those into this?

Mohammer5:
@varl The Menu is just used for demo in the stories. In theorie, anything could be displayed inside the Popover, not just a menu, which allows for more customization.

The DropMenu is an internal right now, made by @HendrikThePendric 
We talked about it yesterday and came to the conclusion that the DropMenu can be replaced, but we'll do it iteratively.  First, we'll release the Popover, then we can replace the internal DropMenu in the components that are using it.

Mohammer5:
@varl @HendrikThePendric 
I just encountered a problem I'm not 100% sure yet how to solve it.

I guess it makes sense to disable scrolling when showing the popover, but this causes the problem that the body's scrollbar disappears, which in turn means that content is moving horizontally. I haven't found a reliable way yet to keep showing the scrollbar while preventing scrolling.

The second alternative is that I recalculate the position on scroll
And the last alternative is that the user can scroll but the menu stays at the same position on the screen (just like the chrome context menu; try opening the context menu on a website and scroll, it's behaves like `position: fixed`)

Personally, I'd go for the last option, just as chrome does, as it's 
* the easiest option
* allowing the user to scroll to some other part of the page with info while keeping the menu open (although that'll probably never happen?)

Update: I forgot to send this comment... I just tried to implement the third option and it works quite well.
There's only one thing that left me puzzled.. maybe you guys know what I'm doing wrong here.
I need to do a `forceUpdate` otherwise the ref of the popover is not found (try it yourself, commenting that line and you'll see that the positioning is wrong). But it seems rather like hack than an actual solution...

Mohammer5:
@varl I just had a chat with @HendrikThePendric, trying to solve that nasty `forceUpdate` thing, which turns out to be a bigger problem, which I'll explain here:

A `ref` shouldn't be used inside the render method, because a `ref` can only be created **after** the first call to `render` (how should react know what DOM element to create otherwise, and a ref can only be created once the DOM element has been rendered).
So the ref is available in `componentDidMount`, but until there's a prop/state change, the `render` method isn't called again, which means that the rendered DOM represents the computation of the first `render`-call return value.

We were able to find a workaround by storing the position of the `Content` in the state and calculating the correct position in `componentDidUpdate`, triggering a re-render with the proper position. We thought this should work for both `componentDidUpdate` and `componentDidMount`.

But it doesn't work for `componentDidMount` because the anchor element's ref comes from the parent component, which suffers from the same problem, it uses the ref in the `render` method (passing it to the `Popover` component) and doesn't update afterwards. Even when the `Popover` updates, the parent component won't update (why should it).

For now this means to me that the `open` prop **must be** false during the first render call and the `alwaysOpen` prop should be removed (that one could be done by passing `open={true}` anyways.

HendrikThePendric:
The majority of what @Mohammer5  wrote above is correct, but the last paragraph is outdated:
> For now this means to me that the open prop must be false during the first render call and the alwaysOpen prop should be removed (that one could be done by passing open={true} anyways.

@Mohammer5 and I kept doing some research and found that there is a way to get this to work when the `open` prop is `true` initially. It means a small change in the API, but I don't think it's a problem, because so far these type of props have only been used by the DropMenu, which was not exposed....

So all that needs to change, is that we don't pass elements directly, but we pass an object with a ref as property. I.e. the new `React.createRef()` API does this by default, and I think it is to circumvent this problem. So that means that instead of having the `anchorEl` prop, we need the `anchorRef` prop.

In a nutshell the reason is this:
- When passing the element directly, this will be undefined when the component is rendered by it's parent. And passing undefined is passing "by value", so it will remain undefined for the entire child component lifecycle
- When passing an object with property `current`, we are passing "by reference". So even if current is undefined when the parent renders it's child, we can access current again later and see what its current value is then.
- React guarantees that all refs are set once the components mount, so all is well.

Mohammer5:
@varl @HendrikThePendric

I've added more meaningful stories for this component.
It's not perfect, because the thing with refs in stories isn't handled well so far.. But it's as good as I could get.

I saw that there's a markdown doc for the [popover](https://github.com/dhis2/ui-core/blob/Popover/stories/info/molecules/popover.md). I think the name `Popover` should be reserved for the component described in the docs. 
So right now I think it's best to rename the `DropMenu` (just for now; the name shouldn't matter to much as it's an internal component anyways), and re-use that name for my implementation of the `Popover`.
What do you guys think about that?

varl:
Good work.

I see what you mean about `Context` and `ContextMenu`; though I find the name `Context` vague ... What does a context component do? What kind of context? There are multiple options the a UI context alone. So is this a `PositionContext` component or something else?

HendrikThePendric:
IMAO **this** is the Popover. What is described in [the markdown doc](https://github.com/dhis2/ui-core/blob/Popover/stories/info/molecules/popover.md) seem to be a **Tooltip**.

Mohammer5:
@HendrikThePendric Interesting proposition! I think it's a good idea, I just have no idea yet what the prop interface should look like, should we allow something like
`position="top-center"` (for the popover) and `position="left-top"` (for the context menu).
We could also make position accept either a string or array of strings to provide fallback positions if the favorite position doesn't fit on the screen.

@varl Yeah `Context` sucks as a name. If we make this the underlying component for all tooltip/popup/popover components (like @HendrikThePendric suggested), this should have a generic yet descriptive, proper name.. Popover might do? `Tooltip` and `Popup` could use it under the hood, and creating a `ContextMenu` should be totally possible as well

HendrikThePendric:
@Mohammer5 
> Popover might do?

You don't like `Popper` as I suggested??

> Should we allow something like position="top-center" (for the popover) and position="left-top" (for the context menu)

I don't think so... Basically, the `Popper` itself would need:

- A ref to it's own DOM node to determine how big it is
- A ref to an anchor node to position itself against
- Instructions on which point of that anchor to position itself against.
- Instructions on how to position itself against that point

That would mean we would end up with props a bit like this:
```javascript
const attachPointPropType = PropTypes.shape({
    vertical: PropTypes.oneOf(['top', 'middle', 'bottom']).isRequired,
    horizontal: PropTypes.oneOf(['left', 'center', 'right']).isRequired,
})

Popper.propTypes = {
    // other prop types
    anchorPosition: attachPointPropType,
    popoverPosition: attachPointPropType,
}
```

The above approach is very flexible, and will allow us to leverage the `Popper` in many use-cases. But I suppose the position logic could become a little bit more complex.

Alternatively, we could go for a single prop, called position, which would have to cover all of the relative positions that we encounter. So far I can think of these already:
- `bottom-center`, for drop-menu, tooltip/popover showing below
- `top-right`, for sub-menu (LTR support?)
- `middle-right` from tooltip/popover showing on the right
- `middle-left` from tooltip/popover showing on the left
- `top-center` for tooltip/popover showing above
I think the other approach is better, it's more flexible and more declarative.

varl:
> @HendrikThePendric Interesting proposition! I think it's a good idea, I just have no idea yet what the prop interface should look like, should we allow something like
> `position="top-center"` (for the popover) and `position="left-top"` (for the context menu).
> We could also make position accept either a string or array of strings to provide fallback positions if the favorite position doesn't fit on the screen.

Something like `<Pop top left />` as booleans could work too. Of course it becomes a problem if something says `<Pop top left bottom right />` but that can be validated. We have other components which have mutually exclusive props too.

HendrikThePendric:
@varl I like `<Pop />` 👍 

And reg. the props to set the position: I'd be happy going with a simpler set of props to define where the Pop should appear compared to the anchor. But I do want to remark a few things:
- In the code that computes the positions, the anchor positions are needed anyway. So we will have to have some sort of mapping between the available positions, and their corresponding anchor points.
- What you gain in simplicity, you lose in flexibility. The image below is there to illustrate the amount of permutations that are possible by keeping the richer props API. I have only included the options for positioning a Pop to the right of an anchor. Obviously, there are many more options there.

=> Suggestion: since this is a component that will be used by other components, and is less likely to be used directly... Could we not apply the complex API to this component and expose the simpler positioning  props API to the components that leverage the Pop? For example:
- Tooltip exposes `top` `right` `bottom` `left`
- Internally, The tooltip will render stuff in a `Pop` and if the position is "right" then the anchorPoint should be vertical: "middle" and horizontal: "right", and the popPoint should be vertical: "middle" and horizontal: "left"

positions_on_right_matrix

Note: whilst typing this, I realised that perhaps my idea of having these props as an object doesn't help. And naming isn't great. We could just split it up in 4: `anchorHorizontal` `anchorVertical`, `popHorizontal` `popVertical`. Still not great...

varl:
> **Suggestion:** since this is a component that will be used by other components, and is less likely to be used directly... Could we not apply the complex API to this component and expose the simpler positioning props API to the components that leverage the Pop?

Sure, sounds good if it's necessary.

Mohammer5:
I like the name `Pop`!

I'll adjust the positioning logic according to @HendrikThePendric 's proposal.
The only thing that's not possible yet would be [this popover](https://github.com/dhis2/ui-core/blob/Popover/stories/info/molecules/popover.md).
So I'll add another property `spacing` (defaults to 0), which applies a spacing between the `Pop` and the anchor.

varl:
What the.. `Pop` wasn't a serious suggestion. 😆

I guess it stuck and it's short. 👍

varl:
Thanks for the illustration @HendrikThePendric. That's good stuff.

@Mohammer5 So the `Pop` component will be a low-level internal component we use to wrap other components like tooltips and dropdown menus?

Mohammer5:
> @Mohammer5 So the Pop component will be a low-level internal component we use to wrap other components like tooltips and dropdown menus?

@varl exactly! Doesn't need to be internal though, I don't see a reason why we shouldn't expose it. I'm sure it can be used in other ways as well

Mohammer5:
I will finish the `Pop` first, so we can release it. Once that is done, we can refactor other components to use that one

Mohammer5:
@varl @HendrikThePendric 

So far, I've added a PoC of how it could work. Before I start refactoring it, I'd like to discuss it though.
I've added a `ROTATION` constant so that when trying which position does work, all options will be tried once a working one is found.
But there's no "intelligent" logic, like

* Given the `Pop` is set to attach it's top corner and right side to the top corner and left side of the anchor
* When the `Pop` doesn't fit on the screen
* Then it should try to use the opposite site

This will get quite complicated once we'll try to add RTL to the mix....

varl:
The topLeft and topRight looks alright in Storybook, but bottomRight and bottomLeft don't, is that intentional for the PoC?

src/Pop/index.js
```diff
@@ -0,0 +1,116 @@
...
+            />
+        )
+
+        if (!level) {
+            return createPortal(content, document.body)
+        }
+
+        return createPortal(
+            <div>
+                <ScreenCover withoutBackgroundColor onClick={onClose} />
```
varl: Why do you need the `ScreenCover` here?

Mohammer5:
> The topLeft and topRight looks alright in Storybook, but bottomRight and bottomLeft don't, is that intentional for the PoC?

@varl yes, that's why I wanted to bring up the discussion:
> But there's no "intelligent" logic, like
> 
> * Given the Pop is set to attach it's top corner and right side to the top corner and left side of the anchor
> * When the Pop doesn't fit on the screen
> * Then it should try to use the opposite site

This seems rather tricky to me with all the possibilities that we have. Either we specify some fallback positions for every position or we magically transform the rotation.
Both ways will get very ugly once we introduce RTL.
The previous version only knew 2 ways and 2 alternatives (it it doesn't fit to the left, display it to the right. If it doesn't fit to the top, position it to the bottom). Now we have **a lot** more..

src/ScreenCover/index.js
```diff
@@ -3,14 +3,16 @@ import propTypes from 'prop-types'
 import css from 'styled-jsx/css'
 import cx from 'classnames'
 
-const Backdrop = ({ onClick }) => (
+const Backdrop = ({ onClick, withoutBackgroundColor }) => (
```
varl: I don't understand why we need a backdrop when creating the menus? It is not legal according to the stacking context in the design system: https://github.com/dhis2/design-system/blob/master/principles/spacing-alignment.md#stacking

The Popup components belong on the _Appliation top layer_ with a `z-index` of `2000`. The `ScreenCover` is a _Blocking layer_ which belongs on `z-index` of `3000`.

_If necessary_: rename `withoutBackgroundColor` to `transparent` and set the background to `transparent` instead of `none`.

varl:
> Both ways will get very ugly once we introduce RTL.

To test the waters, we should start with a simplified RTL version where we only adjust the text, but leave the interface intact for now.

I know it's not optimal, but having something consistent and functional is better than running the risk of destroying the code base early by opting in to the complexity of full-on side-by-side RTL conversion.

Mohammer5:
@varl @HendrikThePendric 

I'll replace the `ScreenCover` with an internal background cover component and will change the ridiculously high z-index to 2000. Now that the sub menus are not opened on hover anymore, I can give it a try to render them inside the DOM tree instead of through a portal (as just discussed with @varl)

Still we have the problem of the "intelligent" alternative position logic. Any ideas how we could work that out efficiently?

Mohammer5:
@varl @HendrikThePendric 

I've changed things a little.. I added fallback positions for each possible position, I also added the possibility to provide custom fallback positions if different ones are wanted.
Stories look ok again, but code needs some refactoring, looks a bit messy right now.

src/Pop/positions.js
```diff
@@ -0,0 +1,379 @@
...
+
+export const CENTER_CENTER_BOTTOM_RIGHT = [
+    { vertical: 'center', horizontal: 'center' },
+    { vertical: 'bottom', horizontal: 'right' },
+]
+
+export const CENTER_CENTER_BOTTOM_CENTER = [
+    { vertical: 'center', horizontal: 'center' },
+    { vertical: 'bottom', horizontal: 'center' },
+]
```
varl: 😱

Mohammer5: I'm still thinking, it should be possible to create them dynamically so we don't have three gazillion constants there... that will def. be part of the refactoring

varl: Nice refactor.

Mohammer5:
@varl I have changed things a little, but I"m not so sure I'm happy with this..
Maybe we need a completely different solution than the one I came up with. The current solution either involves a lot of constants or computation during script execution (we could change this to be lazy, but still a lot of computation has to be done to cover all cases)
Would like to know your (and @HendrikThePendric's; if he was available) thoughts on this because this seems quite a bit of computation for just a core component

varl:
@Mohammer5 Yeah, agreed. It seems complex, though to be fair it is complicated to ensure that a popup is never off screen or invisible or breaks the layout when it pops up.

I'd be quite happy to make some trade-offs here to reduce the computations. Let's see what @HendrikThePendric has to say when he gets back from vacation next week and give this a break for now. Maybe we'll come up with something entirely different if we stop looking at it for a bit. :man_shrugging:

HendrikThePendric:
> Maybe we need a completely different solution than the one I came up with

I'm not sure if inverting positions when stuff doesn't fit is the way to go. It's very complex and could also lead to unexpected behavior: suppose a drop-**down**-menu does not fit by a margin of 2px. And because of this we flip it upwards. Now it shows **above** the anchor, which is very unexpected, and it did _almost fit._ 

Instead I would opt for just shifting the `Pop` a bit so it fits, just to prevent scroll bars to appear i.e.:
- First compute the Pop's position assuming an infinite plane
- Then asses whether this fits into the `body`'s current dimensions. If not, shift the pop so it does
This is simpler and more predictable

Also, I don't really understand the whole concept behind fallback positions....

Mohammer5:
@HendrikThePendric I would def. prefer your approach, but let's say

* There's a `Pop` with some content right next to the border or the viewport
* When clicking on an item inside it opens another `Pop` which is significantly larger than the first one (i. e. because the text children have longer text and are set to `white-space: nowrap`)

The new `Pop` would completely overlap the smaller one, leaving the user no option to close everything and find the path again (this will get even worse when you've clicked already through several levels of `Pop`s)

If we wanted to prevent this edge case, it'll get even worse than just checking a fallback.

The idea why I came up with fallbacks is:
If the `Pop` should be on the left (like a normal context menu), if it doesn't fit, the most logical next position would be on the right). The fallback of TOP_RIGHT_TOP_LEFT is TOP_LEFT_TOP_RIGHT.
And all fallback position kind have the same feeling. They either always centered or always displayed over the anchor or always outside of the anchor

varl:
What's the simplest way that covers 80% of the cases?

HendrikThePendric:
**Reg. the inverting of positions in case things don't fit**
@Mohammer5 showed me a few compelling examples of why inverting is good. I can indeed be annoying to have to go through a nested menu all again because you ended up opening the wrong submenu, and it now covers the pervious menu. 

It does however introduce a lot of complexity in the code. It would be nice if we could somehow delegate this to the components that would suffer from this issue... So far, I think we identified only sub-menus and tooltips as components that would suffer:
- When a tooltip completely covers the UI element it is providing info about that is not good. But personally, I can live with it.
- For the nested menus, the problem is more severe, I think. But we could argue that the only thing the Pop is responsible for is making sure it appends an element to the end of the body that fits into the body. The Menu should then be responsible for laying out its sub-menus. Please note that because the initial menu is now in a portal above the rest of the DOM, we can rest assured that anything we add to this portal will also be above the rest of the DOM. So we could use much simpler positioning strategies from this point onwards.

=> I am not sure what to do with this part. Perhaps the usefulness of the current solution outweighs its complexity. I guess we would have to spend some time trying to work out how we could _delegate the complexities_.

**Reg. markup and positioning**
The current implementation uses fixed positioning and when deciding if something _fits_ it looks at the window edges. Upon reflection it might make a lot more sense to position the `Pop` relative to the `body` instead. That way (given body height > window height), when a user scrolls, we don't have to correct the position. So that would mean using slightly different markup (see below), position `absolute` vs `fixed`, and a slightly different way to determine whether the `Pop` fits
```jsx
<Fragment>
    <BackgroundCover />
    {/* this div would have position absolute and the computed top/left */}
    <div>
        {content}
    </div>
</Fragment>
```
=> I think we should definitely change this part. Makes much more sense to me.

HendrikThePendric:
Another thought..... I know we are not keen on adding adding too many dependencies, but it sort of feels we are re-inventing the wheel a little bit here. [Popper.js](https://github.com/FezVrasta/popper.js) seems to be a tried-and-tested positioning engine that [also covers inversion](https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/popper-documentation.md#modifiers..flip). It's 7.25 kb gzipped.

Mohammer5:
> What's the simplest way that covers 80% of the cases?

@varl @HendrikThePendric I think that 80% of the cases are 4 or 5 initial positions
* top/left + top/right (standard context menu)
* top/right + top/left (context menu alternative)
* top/center + bottom/center (standard tooltip)
* center/left + center/right (tooltip alternative)

Every other position is quite unlikely and in most cases a `Modal` should be preferred.
I think the simplest solution is to only allow these combinations with appropriate fallbacks, which can be customized with props.

Do you have any scenario in mind where it makes sense not to use one of the 4 cases mentioned above?

If you do, I think it might make sense to use `popper.js` (the one that @HendrikThePendric found/mentioned). It has ~100 contributors, it seems to be actively maintained (the last commit was yesterday) and the licence is MIT, so we don't have to worry about the positioning logic.

Mohammer5:
What we can do as well is: We have a limited set of possible positions and have a custom off-screen logic for each. Whenever we need a new one, we can implement that one specifically (there shouldn't be too many new cases) which would be a truly iterative process that takes the least time (when not using popper.js I guess)

HendrikThePendric:
I've been thinking about what @varl said:
> What's the simplest way that covers 80% of the cases?

And about my suggestion:
> Instead I would opt for just shifting the Pop a bit so it fits, just to prevent scroll bars to appear

Wouldn't the above be a solution that covers (at least) 80% of cases? Because to me it seems that the problem we are trying to solve by inverting the positions is only a problem under quite a specific set of circumstances:
- We need multiple layers of `Pop`s. The only use-case I can think of so far is the current implementation of the sub-menu. (I am not seeing the Tooltip overlapping its anchor as a major issue here)
- And the user is on a very small screen, or the initial anchor is far to the right of the screen

Perhaps the current solution using inversion is good. I am just a little bit concerned that we now have almost 400 lines of code just to work out where to place something. The current solution also introduces some iterations into the logic, because it keeps has to "try" different positions. It shouldn't have too much impact on performance I guess. If we go with this approach, I do think it is important that we refactor the code so that everything is relative to the body and positioned absolutely, because with the current window based approach, we'd need a scroll handler and recompute the position all the time.

I think I would also prefer to either go with the current solution (after a little refactoring), or a solution that has a less complex positioning logic. I don't think that going for a partial implementation of the current approach will help much.

Just my 2 cents

varl:
>I think I would also prefer to either go with the current solution (after a little refactoring), or a solution that has a less complex positioning logic. 

The current implementation is too complex for the scope of `ui-core`. Components in `ui-core` should be _dumb_ and this one is rather clever.

Does one component have to solve all use cases?

Would it be simpler to divide the problem into distinct parts?

If we are going for a full-blown positioning engine (i.e. lots of business logic) then the correct place is `ui-widgets`. If built in `ui-widgets` we _could_ look at using PopperJS too.

HendrikThePendric:
@varl how do you feel about ui-core components relying on a ui-widget component internally? You think that's OK?

HendrikThePendric:
> Does one component have to solve all use cases?

No. If we go with a simpler solution, the submenu can solve its own overlapping issues. Or we could even go for a slightly different sub-menu UI to prevent the issue all together. For example, if we used collapsable list items in a single `Pop` we wouldn't even be having this discussion.

Mohammer5:
> how do you feel about ui-core components relying on a ui-widget component internally? You think that's OK?

I think this is a no-go!

@varl @HendrikThePendric I think the simplest solution would be not to find out an alternative positioning automatically. We could just have a fallback prop that's either another position (e. g. the position is `anchor: top/right` & `pop: top/left` and the fallback is `anchor: top/left` & `pop: top/right`)

This fallback capability is already there, I'd just need to remove the automatic fallback calculation.
We could also say that the `fallbackPosition` could look like this:
```js
fallbackPosition: propTypes.arrayOf(propTypes.shape({
  anchor: {
    horizontal: propTypes.string,
    vertical: propTypes.string,
  },
  pop: {
    horizontal: propTypes.string,
    vertical: propTypes.string,
  },
}))
```

This way all widgets (and I think a `Menu` with no-offscreen behavior is already a widget) have full control over how much complexity they want.

varl:
> @varl how do you feel about ui-core components relying on a ui-widget component internally? You think that's OK?

Absolutely not.

That solution would require the `ui-widget` to assemble higher-level components using e.g. a static tooltip component from `ui-core` + the Pop component from `ui-widget`.

E.g.
```js-jsx
// TooltipPopup.js in @dhis2/ui-widgets

import { Tooltip } from '@dhis2/ui-core'
import { Pop } from '@dhis2/ui-widgets'

export const TooltipPopup = () => (
<Pop {...props}>
    <Tooltip />
</Pop>
)
```

HendrikThePendric:
Glad you both agree that core should rely on widgets. I feel the same.

So whichever choice we make here will have a knock-on effect on other components that need to be positioned relative to an anchor: if we conclude that the `Pop` requires a certain level of complexity, and thus needs to become a widget, then components like a dropdown-button, split-button, tooltip, and menu will also be imported from widgets (because the ones in core are not directly usable) Which doesn't feel right....

Choices choices choices

Mohammer5:
@varl @HendrikThePendric Neither of you has said anything about my suggestion to remove automatic positioning from the component while keeping the `fallbackPositions` prop. You don't agree on that one, because then it'd be a "dumb" component, at least the code for generating all the fallbacks can be removed completely

varl:
> So whichever choice we make here will have a knock-on effect on other components that need to be positioned relative to an anchor: if we conclude that the `Pop` requires a certain level of complexity, and thus needs to become a widget, then components like a dropdown-button, split-button, tooltip, and menu will also be imported from widgets (because the ones in core are not directly usable) Which doesn't feel right....

I don't think that it's a problem that the user has to scroll if a DropdownButton or SplitButton opens at the bottom of a screen, so DropMenu is fine in that case. E.g. from GitHub:

github

The important thing is that the user CAN scroll _down_ to access it.

To me it's much worse in the situation where a menu just disappears off screen:

side-scroll

Another important distinction is if a menu opens on hover or on click. In the case of a SplitButton or a DropdownButton they open on click, so even if the user starts to fickle with scrolling, the element is going to be there. The hover menu we have is very bad since if you miss a pixel you lose it. It's fine to open a submenu on hover, but it shouldn't close when it is no longer hovered.

For submenus I this something like [ContextMenu](https://github.com/s-yadav/contextMenu.js/blob/master/contextMenu.js#L422-L536) does is perfectly adequate:

side-scroll

We don't need _one_ component that can do everything either.

I'd rather have a `DropMenu` component which DropdownButton and SplitButton can use. The `Tooltip` can have it's own little Portal for displaying a tooltip correctly, and the `Menu` (actually it's in the `MenuItem`) can have it's own implementation that specializes in opening submenus so they don't go off-screen.

We only extracted the `DropMenu` component since otherwise it would be a straight duplication between the other components.

varl:
> @varl @HendrikThePendric Neither of you has said anything about my suggestion to remove automatic positioning from the component while keeping the `fallbackPositions` prop. You don't agree on that one, because then it'd be a "dumb" component, at least the code for generating all the fallbacks can be removed completely

I'm not sure how much code that removes. It reduces the complexity but even if we remove the entire `fallbacks.js` file, there is still positioning logic in 3 more files. It is still complex.

We will rarely have more than [three levels because the design system limits us to that](https://github.com/dhis2/design-system/blob/master/atoms/menu.md#menu-levels) so we do not need perfect positioning for arbitrary nested menus. At most we will have:

```Menu => SubMenu => SubSubMenu```

To reduce the overall complexity I do not think we need to provide a positioning engine component in `ui-core`. If we could have a single `Pop` component internally, that would be nice, but if we cannot I think that implementing the necessary logic where it's needed makes sense:

- The `Menu` needs the ability to position a SubMenu to the _left_ of the root `Menu` instead of to the _right_ if it doesn't fit in the viewport. If the SubSubMenu again fits to the right, put it there. If not, put it to the left. Rinse and repeat.

- It looks like the [Tooltip](https://github.com/dhis2/design-system/blob/master/atoms/tooltip.md) and [Popover](https://github.com/dhis2/design-system/blob/master/molecules/popover.md) could share the same internal component, maybe even the `DropMenu` can be repurposed (and renamed) and used across the board for Tooltip/Popover/DropdownMenu/SplitButton.

**Comment by HendrikThePendric** [comment]
> The Menu needs the ability to position a SubMenu to the left of the root Menu instead of to the right if it doesn't fit in the viewport. If the SubSubMenu again fits to the right, put it there. If not, put it to the left. Rinse and repeat.
Yes. Completely agree. This is what I was trying to suggest as well. Delegating this to the menu will allow us to keep things simple in the Pop.

**Comment by varl** [comment]
@Mohammer5 Does that make sense to you for the `Pop` component to only deal with four positions: top, right, bottom, left?

There's a single edge-case, and that's if a `Pop`ped component hits the edges of a viewport.

In the horizontal case it would need to be positioned to avoid causing the horizontal scrollbars to appear.

In a vertical case it will only be a problem on if they popup on hover, which I don't think we should do. Primarily since hover is not a concept that works on touch screens and touch screens are becoming commonplace on laptops as well as dominating on mobile devices.

**Comment by HendrikThePendric** [comment]
Sorry to butt in. I can't resist.
> the Pop component to only deal with four positions: top, right, bottom, left?

At first I thought that was not possible, due because a tooltip needs to vertically centred when positioned on the side of an anchor, whilst the sub-menu needs to have the top aligned with the top of the menu-item. If we chose to let the menu take care of laying out its own sub-menus, then this objection is no longer valid. I feel pretty confident that we can have a fairly straightforward solution given these constraints:
1. No flipping/inversion, only shifting the Pop if it would "stick out" otherwise
2. Only support 4 positions
3. Position against the body, not the window, so we can avoid repositioning on scroll.

> There's a single edge-case, and that's if a Popped component hits the edges of a viewport.

FYI I am advocating checking if the component hits the edges of the body, not the window as I stated a few times above...

**Comment by varl** [comment]
> > 
> 
> FYI I am advocating checking if the component hits the edges of the body, not the window as I stated a few times above...

That makes sense. I don't want the element to follow along on scroll either. 👍

**Comment by Mohammer5** [comment]
@varl @HendrikThePendric so you're basically suggesting we'll go back to the [version we had on 14 days ago](https://github.com/dhis2/ui-core/tree/6006fb47dfd6ba6690ea335585d512fd23137b19) with a few adjustments?.

Changes I have to make:
* Use `body` instead of viewport (the Pop can be scrolled off-screen, and can be opened while off-screen)

> No flipping/inversion, only shifting the Pop if it would "stick out" otherwise

I think for the menu it should flip side, so it behaves like every other menu I've seen so far out there:
image

> Only support 4 positions

The menu actually only needs 2 positions (left/right), I don't see no reason to make it more complex. Did I forget about a usecase?

**Comment by HendrikThePendric** [comment]
> so you're basically suggesting we'll go back to the version we had on 14 days ago with a few adjustments?

Yes I think that was fairly close to what we need. Things shouldn't become much more complex than that.

> > Only support 4 positions

> The menu actually only needs 2 positions (left/right), I don't see no reason to make it more complex. Did I forget about a usecase?

I'm not quite sure I fully understand what you are saying, but will still try to respond. I think, with the `Tooltip` in mind, the `Pop` should support 4 directions. And for the menu/sub-menu and flipping/inverting things to the opposite side, I agree left || right is enough. But I think that is not in the scope of the `Pop`, the menu should do the flipping...

**Comment by varl** [comment]
> I think for the menu it should flip side, so it behaves like every other menu I've seen so far out there: {img}

> The menu actually only needs 2 positions (left/right), I don't see no reason to make it more complex.

Agreed, but we can build that into the menu and let the `Pop` component be as simple as possible.

**Comment by HendrikThePendric** [comment]
This is probably not the appropriate place to post this. But if it turns out that getting the inversion/flipping is hard to get right for the Menu component, then we could also discuss with Joe if we could perhaps implement a different mechanism for a layered menu. They could be in a single container too:
[Multi-Level Accordion Menu](https://codyhouse.co/demo/multi-level-accordion-menu/index.html) (I like this one a lot)
[Responsive Multi-Level Menu](https://tympanus.net/Development/ResponsiveMultiLevelMenu/)

**Comment by Mohammer5** [comment]
Ok, to summarize what needs to be done:
1. The `Pop` will be used for 1-dimensional information, like tooltips, etc.
1. The `Menu` component will have to be modified to flip sides (should be a different PR).
1. The logic for checking whether an element fits inside a container can be extracted so it can be used by both the `Pop` and `Menu`, just the positioning logic for finding the next position will be different

When the `Pop` is shifting to find the next possible position to be fully visible, I'd suggest to go in 5px steps in order to not to have too many positions to check if it has to be moved 400px for example.

And the edge case that the `Pop` won't fit into the body, I think we can just ignore that one, it should be handled by the component implementing the `Pop`.

All on the same page?

**Comment by Mohammer5** [comment]
> But if it turns out that getting the inversion/flipping is hard to get right for the Menu component

@HendrikThePendric The flipping shouldn't be a problem, I think, we just need to see if the sub menu fits, if yes, do nothing, if no, check the other side. If that one fits, put it there, if it doesn't, we could fallback to a modal (for mobile devices which I'd really like to support because tablets/phones are the future, especially in poorer countries)

**Comment by varl** [comment]
> 1. The `Pop` will be used for 1-dimensional information, like tooltips, etc.
> 
> 2. The `Menu` component will have to be modified to flip sides (should be a different PR).
> 
> 3. The logic for checking whether an element fits inside a container can be extracted so it can be used by both the `Pop` and `Menu`, just the positioning logic for finding the next position will be different

👍 

> When the `Pop` is shifting to find the next possible position to be fully visible, I'd suggest to go in 5px steps in order to not to have too many positions to check if it has to be moved 400px for example.

Smart.
 
> And the edge case that the `Pop` won't fit into the body, I think we can just ignore that one, it should be handled by the component implementing the `Pop`.

That might be fair. I'm not entirely clear on how that would work but I like the principle of it.

**Comment by Mohammer5** [comment]
> That might be fair. I'm not entirely clear on how that would work but I like the principle of it.

@varl The implementing component should say something like:

```css
.child-inside-pop {
  max-height: 100vh;
  overflow-y: auto;
}
```

varl:
@Mohammer5 Ah, OK. Different edge case than what I was thinking about where the e.g. Tooltip would not fit into the body in this fashion:

outside

Mohammer5:
I just wanted to start doing this, but there's another detail we'll have to clarify.
We agreed to only support 4 sides with this component, does that mean that the pop will always be position against the center of said sides? Like this:

image

Or do we want the side to be aligned to the top?

image

HendrikThePendric:
I would say the first option, vertically centred

Mohammer5:
> I would say the first option, vertically centred

Agree, that's what I though about as well

varl:
Vertically centered for the `Pop` component.

The `Menu` needs the popup to be aligned to the top of the anchor, but that's out of scope for the `Pop` component.

Mohammer5:
> When the Pop is shifting to find the next possible position to be fully visible, I'd suggest to go in 5px steps in order to not to have too many positions to check if it has to be moved 400px for example.

This was a really stupid idea I came up with.. During development I noticed I could just calculate the correct value on the fly without having to try rendering a position first.

varl:
> This was a really stupid idea I came up with.. During development I noticed I could just calculate the correct value on the fly without having to try rendering a position first.

The concept makes sense (to avoid unnecessary work), but looks like you came up with a way to avoid all the work instead. 💯

Mohammer5:
I've added a spacing property to allow some space between the anchor and the pop.
The name `Pop` could be changed to `Tooltip` as this is basically a tooltip component now.

If you agree to that, I'll add the possibility to display an arrow-like triangle at the center of the pop's side facing the anchor.

varl:
Couldn't the `Pop` component be used instead of `DropMenu` for the DropdownButton and SplitButton?

Mohammer5:
> Couldn't the Pop component be used instead of DropMenu for the DropdownButton and SplitButton?

I'll try and check what happens (if that's possible, it should be another PR thought)

varl:
Sure, it just impacts what the name is.

Mohammer5:
Ok, I've looked at both the `DropdownButton` and `SplitButton`, and the `Pop` component is not the right use case here:

Both components provide an api to render anything compoent inside the `DropMenu` (which both components use internally). This means that the dropdown could be a simple `Pop` but it could also be a `Menu` component (which will handle to off-screen stuff itself).

Having said that, I think we could/should adjust both the `DropdownBytton` and the `SplitButton` a bit (@HendrikThePendric, I think you made those components, right?). Now that we'll have mutilple, different components which will handle the off-screen/out-of-body functionality themselves, we can replace the internal `DropMenu` and let the implementing component/app handle that:

Before suggestion:
```jsx
{open && (
    <DropMenu
        component={this.props.component}
        onClose={() => this.setState({ open: false })}
        anchorEl={this.anchorRef.current}
    />
)}
```

With suggestion:
```jsx
{open && React.cloneElement(
  this.props.component,
  {
    onClose: this.onToggle,
    anchor: this.anchorRef,
  }
)}
```

This would introduce a standard to give the anchor prop the name `anchor` and that the menu/pop component will have to accept an `onClose` callback.

Mohammer5:
I guess I can safely rename it to tooltip and add the triangular arrow icon (which can be enabled by a prop)

Mohammer5:
@varl @HendrikThePendric I've renamed this one to `Tooltip` and added a `withArrow` prop.
You can check that stuff in the stories.

I guess that's it for this component?

src/Tooltip/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+    render() {
+        if (!this.props.open) return null
+
+        const { side, children, onClose, withArrow } = this.props
+        const { position, adjustment } = this.state
+
+        return createPortal(
+            <div>
+                <BackgroundCover onClick={onClose} />
```
varl: I'm not sure I think that a Tooltip should have a background cover. That feels extremely invasive UX-wise.

Ping @cooper-joe.

cooper-joe: Correct, a tooltip should not have a background cover. 
(However, is this PR not for the popover? Either way, neither tooltip nor popover should use a Cover.)

Mohammer5: It's just an invisible cover to catch the click outside of the tooltip.

varl: It's invisible, but it still blocks the UI. It's possible to handle the "click outside" without a div covering 100% of the screen.

src/Tooltip/Content.js
```diff
@@ -0,0 +1,88 @@
...
+            className={cx('container', {
+                'with-arrow': withArrow,
+                [`with-arrow-${invertSide(side)}`]: withArrow,
+            })}
+        >
+            <div className="content">{children}</div>
+
+            <style jsx>{`
+                .container {
+                    background: white;
```
varl: We have a `colors.white` in `theme.js`.

src/Tooltip/Content.js
```diff
@@ -0,0 +1,88 @@
...
+                    box-shadow: 0 0 3px rgba(0, 0, 0, 0.6);
+                    position: absolute;
+                }
+
+                .content {
+                    position: relative;
+                    z-index: 2;
```
varl: Avoid `z-index` in general.

`z-index` of `2` is illegal since it is within the application stack.

src/Tooltip/Content.js
```diff
@@ -0,0 +1,88 @@
...
+
+                .with-arrow:after {
+                    content: '';
+                    position: absolute;
+                    top: 0;
+                    left: 0;
+                    width: 100%;
+                    height: 100%;
+                    background: white;
+                    z-index: 1;
```
varl: Avoid `z-index` in general.

`z-index` of `1` is illegal since it is within the application stack.

src/Tooltip/Content.js
```diff
@@ -0,0 +1,88 @@
...
+                }
+
+                .with-arrow:before {
+                    content: '';
+                    background: white;
+                    position: absolute;
+                    z-index: 0;
```
varl: Avoid `z-index` in general.

`z-index` of `0` is illegal since it is within the application stack.

src/Tooltip/Content.js
```diff
@@ -0,0 +1,88 @@
...
+                    left: 0;
+                    width: 100%;
+                    height: 100%;
+                    background: white;
+                    z-index: 1;
+                }
+
+                .with-arrow:before {
+                    content: '';
+                    background: white;
```
varl: We have a `colors.white` in `theme.js`.

src/Tooltip/index.js
```diff
@@ -0,0 +1,122 @@
...
+Tooltip.propTypes = {
+    /**
+     * Must be created with `React.createRef()`
+     */
+    anchorRef: reactRef.isRequired,
+
+    /**
+     * Tooltip will always be centered to the center of the anchor's side
+     */
+    side: propTypes.oneOf(['top', 'right', 'bottom', 'left']).isRequired,
```
varl: Default should be `side='top'`

src/Tooltip/index.js
```diff
@@ -0,0 +1,122 @@
...
+import { Content } from './Content'
+import {
+    arePositionsEqual,
+    getPosition,
+    getScrollAndClientOffset,
+    propPosition,
+} from './helpers'
+
+/**
+ * The Tooltip component is a content container that behaves like a context menu
```
varl: This is surprising to me.

src/Tooltip/index.js
```diff
@@ -0,0 +1,122 @@
...
+                />
+
+                <style jsx>{`
+                    div {
+                        left: 0;
+                        height: 100%;
+                        position: absolute;
+                        top: 0;
+                        width: 100%;
+                        z-index: 2000;
```
varl: This should use `layers.applicationTop` from `theme.js`.

stories/info/molecules/pop.md
```diff
@@ -0,0 +1,19 @@
+# Context
```
varl: This file has the wrong name and the text is wrong.

varl:
After some discussion we realize that this is actually the Popover component:

Some changes are needed:

- `withArrow` should be the default
- `side="top"` should be default
- The `Pop.md` document needs an overall (or removal)
- There are comments still referring to the component as a Context menu, which needs to be removed too.
- Other cleanups noted inline.

src/Tooltip/Content.js
Mohammer5: As discussed, z-index inside elements with z-index are relative inside but to not effect the outside z-index and other elements that have the same z-index as the outer one

src/Tooltip/Content.js
Mohammer5: As discussed, z-index inside elements with z-index are relative inside but to not effect the outside z-index and other elements that have the same z-index as the outer one

src/Tooltip/Content.js
Mohammer5: As discussed, z-index inside elements with z-index are relative inside but to not effect the outside z-index and other elements that have the same z-index as the outer one

Mohammer5:
> Some changes are needed:
> 
> * withArrow should be the default
> * side="top" should be default
> * The Pop.md document needs an overall (or removal)
> * There are comments still referring to the component as a Context menu, which needs to be removed too.
> * Other cleanups noted inline.

@varl 

* `withArrow` now is `noArrow` and defaults to falsy (undefined)
* `side` now defaults to `top`
* `Pop.md` has been removed as it's not used anyways. `Popover.md` now links to the design specs
* Removed all "Context" comments
* Inline notes have been taken care of
* `Tooltip` now is called `Popover`

.storybook/frame-styles-decorator.js
```diff
@@ -8,9 +8,9 @@ const FrameStylesWrapper = fn => (
         <style jsx>{`
```
varl: I think we should drop this from Storybook entirely before we start to build components that rely on this existing in Applications.

Mohammer5: Agree

Mohammer5:
@varl @HendrikThePendric 

Ok, so this issue is basically done... I just have one problem that I simply can't figure out.
It's in the following stories:

* `Popover` > `Left`
* `Popover` > `Left with spacing`
* `Popover (out of body)` > `Top`

For some reason the calculation is off (by 15px if I remember correctly)... But when I use the function to calculate the value in the browser's console, everything is correct. It's also correct in all other case, but these 3 fail consistently and I can't explain why.

I tried to delay the calculation, but even when delaying it by 1000ms, it's still calculated wrongly

I've removed the `FrameStyleWrapper` with the latest commit previous to this comment, I was hoping that that would fix it, but it didn't...

HendrikThePendric:
I've had a quick poke around but couldn't really reproduce the issue you mention (calculation being 15px off).

I was testing in Chrome and later on also tried FireFox. What I did notice in FF was that the styles hadn't been applied properly and some examples actually looked "the wrong way around" in FF. With this I mean that the button was on the left when it should have really been aligned to the right.... I did manage to find out why that was: The `Btn` component used in the examples is using this style `width: fit-content`. This value for width is not supported by FF. It does support `width: -moz-fit-content` (but then Chrome doesn't)...

So.... I am not at all sure whether this is related to the issue you had, but I suppose it could be. In either case this needs to be fixed.

HendrikThePendric:
Before we merge this component into master, I think we need to test a lot more scenarios. I would specifically like to verify:
a) that this component behaves correctly when the body height exceeds the window height and the Popover anchor is below the fold.
b) that this element behaves correctly when there is other floating content around: will it float over the top or not?

And shoot... I just realised that because we moved from positioning-against-the-window (which would have required a scroll listener) to positioning against the body... The position of this element is not going to be correct when the anchor is (in) an element with fixed positioning.... I think we can live with that...

HendrikThePendric:
@Mohammer5 I believe you will be starting work on this component again soon. The design-system has evolved quite a bit since we last discussed things, and looks to be complete now.

I would argue that in the end, this is what we need:
- A `Popover` component, which should be able to do the following:
    - Render above all other content (in a portal)
    - Position relative to another element, supporting all forms of relative positioning that may be required (i.e. by the Tooltip, the Select, the Menu)
    - Prevent rendering off-screen / cause scrollbars to appear
    - Should stay in the correct place while scrolling
- A `Tooltip` component, that should do something like this
    - Be positionable (leveraging the Popover)
    - Render text
    - Comply with the design-system 
- Update various other components to also use this

HendrikThePendric:
@varl @Mohammer5 I think this one can be closed right?

Mohammer5:
@HendrikThePendric Yes please... :D

PR #226, opened 2019-05-20
fix: add font-family to everything

### LINKED ISSUE #208: Button not using theme font
The font applied is `system-ui` from the browser's default stylesheet.
The font should be taken from the theme instead.

varl:
Fixes #208.

### DIFF
--- a/.storybook/reset-decorator.js
+++ b/.storybook/reset-decorator.js
@@ -1,0 +1,12 @@
+import React from 'react'
+
+import { CssReset } from '../src'
+
+const CssResetWrapper = fn => (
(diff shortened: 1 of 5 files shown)


.storybook/reset-decorator.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+import { CssReset } from '../src'
+
+const CssResetWrapper = fn => (
+    <div style={{ margin: '10px' }}>
```
Mohammer5: It's completely irrelevant, but you could just write `margin: 10`, instead of `margin: '10px'`. For the lazy people ;) No need to change anything here!

PR #227, opened 2019-05-20
feat(buttonstrip): add button strip component

HendrikThePendric:
To give the buttons some spacing I first tried the suggested solution: applying the `justify-content: space-between;` style to the container. However, this gave the wrong results because this would let the container use the entire flexbox to layout its content. See [this page](https://css-tricks.com/almanac/properties/j/justify-content/) for an illustration.

Then I tried a slightly obscure CSS solution. I'll just talk you through it:
- Using `div >` so it only applies the styles to the containers direct descendants
- Had to use `:global` because the children are not in the jsx scope
- Using the `*` selector to select any type of element, since we don't know if this is always going to be a `button`
- Give all the elements apart from the first one `:not(:first-child)` a margin-left 

It works a treat, but feels slightly hacky. Lemme know if we should merge this in.


Mohammer5:
Thanks for the PR.
I think the css selector is too complicated and too expensive, a much simpler solution looks like this:

```css
div > :global(*) {
    margin-left: 16px;
}
```

The container is a flex container with the default value for `flex-wrap` (which is `nowrap`), so overflowing items will just overflow and stay visible, the margin left of the first item won't cause it to break into the next row.

But this selector still targets all elements in the DOM tree and then filters them by being inside a div with the styled-jsx class, I think we should only target the `button`s, as that will be the most likely scenario when the `Modal` is used.

```css
div > :global(button) {
    margin-left: 16px;
}
```

<hr />
If we want to change the container to `flex-wrap: wrap`, then I'd propose the following styles
(I'm against this though, if the app needs to have a more complex structure inside the action component, it should take care of the styles itself. When targeting only direct child button components, there should be no conflict with the above applied margin either):

```css
div {
    flex-wrap: wrap;
}

div > global(button) {
    margin-left: 16px;
}

div > global(button:first-child) {
    margin-left: 0;
}
```

HendrikThePendric:
I think your suggestion makes sense. Using `*` as the last selector, and the nested selector being interpreted by the browser from right to left is not a good combination.

Upon reflection, I'm not sure whether this whole exercise makes any sense... We are now:
- Making assumptions (quite likely valid ones) about the type of elements that are going to be rendered into the `Actions`
- And we are then also interfering with these elements. Which leads (/could lead) to unexpected behaviour.

On the other hand, not making these assumptions, leads to a pretty quirky situation for developers using our component. Having to add className + style, or a wrapper span all the time is not convenient at all.

I would say that the current approach is flawed, and not doing anything is also not an option, so we should look for another solution. I can think of two directions we could explore:
1. **Add a helper component** Suppose we add a `ButtonStrip` component. This would only accept Buttons (as in ui-core Button component). And this component would take care of the spacing...
2. **Changing the props API** Suppose `Actions` wouldn't accept children, but instead it would take an array of `actions` where every action would have a `label`, `onClick`, and `type`.

HendrikThePendric:
Come to think of it, a ButtonStrip component could also be useful in a different way. We could have a compact and a spaced variant. Compact would place all buttons adjacent to each other, and make sure only the start and end buttons have border-radius on the correct corners.
Spaced could be to solve the current issue

varl:
Hm. Take a look at the `SplitButton` .. could the ButtonStrip solve that case too?

Right now, I'm leaning towards the App passing in a class and dealing with the spacing, but the ButtonStrip component might be interesting.

HendrikThePendric:
Yes, `SplitButton` would be a good example where we could also use it

`src/SplitButton/index.js` lines 43-57 would be replaced by the following
```jsx
<ButtonStrip compact>
    <Button {...this.props}>{this.props.children}</Button>

    <Button {...this.props} onClick={this.onToggle}>
        {icon}
    </Button>
</ButtonStrip>
```

varl:
Alright, try that and let's see some code. :+1:

Mohammer5:
Sounds good to me as well!

varl:
Can we do anything with the cut off focus?

2019-05-22-103217_392x70_scrot

HendrikThePendric:
OK, I've now introduced the `ButtonStrip` component and added some stories for it. I have also used it to:
- Build the `SplitButton`
- Align the `Button` components in `Modal.Actions`

Some explanation of how I ended up with this solution is in order, I think. I explored several options that all came with their own downsides:
1. Using `React.cloneElement` to add an additional `className` to the `Button` children in the `ButtonStrip`. This would also mean using the `resolve` function of styled-jsx. I decided against this option because it would introduce the overhead of both `resolve` and `cloneElement`.
2. The current solution, which uses a complex (slow) CSS selectors. I like this solution because it only overrides the styles that need to be overridden.
3. I also explored another option that used more efficient selectors, but the downside of that was that I then had to reset styles. For example, I would first set border-radius to 0 on all buttons, and then I would need to set the border radius for the first and last child. In order to do so, I would have to start exporting values from the button styles. IMO this was getting a bit messy.

HendrikThePendric:
> Can we do anything with the cut off focus?
Shoot, I hadn't considered that. Will have a look, might be the achilles heel :-(

HendrikThePendric:
OK, so now I have "fixed" the focus styles of the compact ButtonStrip. The downside is that we now have double borders between the buttons.

I do have a (much) better fix in mind, but this would involve refactoring the button CSS. I think this could also improve how some of the current configurations look (i.e. primary button focus style looks a bit off) But we'd have to discuss that first...

src/ButtonStrip/helper.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import { Button } from '../Button'
+
+export const buttonArrayPropType = (props, propName, componentName) => {
+    const prop = props[propName]
+
+    let error = null
+    let count = 0
```
Mohammer5: can be:

```js
if (React.Children.count(prop) < 2) {
    new Error(`${componentName} should have at least 2 children.`)
}
```

HendrikThePendric:
Reg. the button CSS refactoring: in a nutshell, I like to stop working with a border+border-radius to create outlines that "stick out", but instead style the button's `::after` pseudo-element. Quite likely this will simplify the stylesheet a too.

src/ButtonStrip/helper.js
HendrikThePendric: Thanks for that. I looked at that method in the docs this morning, but it hadn't clicked in my head that the `children` in that example was equivalent to the `pop` in our case. DUH.

HendrikThePendric:
OK, so now I've updated the docs and merged master into this branch to prevent conflicts. Ready to merge this in now, if you both approve.

A lot has happened, so I think a short summary is order:
- The initial reason for starting this branch was to layout the buttons in the `Modal.Actions` component
- Since there was no reliable/satisfactory way of doing this, I created a new component, the `ButtonStrip`. This ButtonStrip could then also be used for constructing the `SplitButton`, when used in "compact-mode".
- I used the new `ButtonStrip` for the `Modal.Actions` and the `SplitButton` and @varl found an issue with the focus state of the compact mode.
- To fix this issue, I implemented the focus outlines in the `Button` component in a different way

Now everything works as expected, but this PR touched quite a few parts of the code-base. So please check the following stories:
- Buttons
- ButtonStrip
- Modal
- SplitButton

stories/info/atoms/buttonStrip.md
```diff
@@ -0,0 +1,20 @@
+# ButtonStrip
+
+A ButtonStrip is a framing element that can be used to layout an inline set of `Buttons`. This component will only accept instances of the `Button` component as its children and requires at least 2 children.
+
+## Usage
+
+In its default configuration, the `ButtonStrip` will add 16px of whitespace between the buttons, and it will align them on the left (flex-start). It also accepts the following properties:
+
+| Prop       | Type      | Description                                                    |
```
varl: The prop table is generated automatically and shown in a tab in storybook. No need to have a manually updated version.

PR #229, opened 2019-05-22
fix: introduce the stacking context

varl:
Based on: https://github.com/dhis2/design-system/blob/master/principles/spacing-alignment.md#stacking

### DIFF
--- a/src/ComponentCover/index.js
+++ b/src/ComponentCover/index.js
@@ -3,6 +3,8 @@ import propTypes from 'prop-types'
 import css from 'styled-jsx/css'
 import cx from 'classnames'
 
+import { layers } from '../theme.js'
+
 const ComponentCover = ({ children, className }) => (
     <div className={className}>
         {children}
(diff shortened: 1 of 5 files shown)


src/ComponentCover/index.js
```diff
@@ -17,7 +19,7 @@ const ComponentCover = ({ children, className }) => (
                 height: inherit;
                 width: inherit;
 
-                z-index: 900;
+                z-index: ${layers.applicationTop - 1};
```
varl: The `ComponentCover` belongs between the application base layer and the application top layer.

We want it to cover all the components in the application base layer, but allow dropdown menus to render above it. This seems like an OK workaround.

src/FieldLabel/Filled/index.js
```diff
@@ -70,11 +70,6 @@ const styles = css`
         width: 100%;
     }
 
-    .focus .content,
-    .value .content {
-        z-index: 2;
-    }
```
varl: I tested this in the InputField and SelectField, but I don't notice a difference with `z-index: 2` or not, so figured I'd remove it. If there is good reason to keep it and I missed a subtle error, I'll put it back.

PR #230, opened 2019-05-22
fix: make form inputs focus-able

### LINKED ISSUE #141: Wishlist:  Keyboard friendly input components

Mohammer5:
This PR will make make the `Radio` and `Checkbox` components focus-able and will make the `SelectField` component show the focus state.

This will introduce a breaking change!
The `Checkbox` component now requires a `value` attribute.

Fixes #141.


Mohammer5:
Should I remove the form stories? I just used them for testing..

varl:
Fixes 141.

varl:
> 
> 
> Should I remove the form stories? I just used them for testing..

I used them to verify, you can remove them now, or, comment them out and save them for testing?

src/Checkbox/Icon.js
```diff
@@ -13,6 +13,11 @@ const icons = resolve`
         fill: ${theme.default};
     }
 
+    .focus :global(path) {
+        stroke: #646464;
+        stroke-width: 2px;
```
varl: Asking @cooper-joe about this.

src/Checkbox/index.js
```diff
@@ -9,67 +9,115 @@ import { Icon } from './Icon'
...
+                            user-select: none;
+                        }
+
+                        .disabled {
+                            cursor: not-allowed;
+                            color: ${theme.disabled};
+                        }
+
+                        .focus {
+                            outline: 1px;
```
varl: @cooper-joe This too.

src/Checkbox/Icon.js
Mohammer5: Ah, good catch! I completely forgot that I wanted to ask for a focus style!
@cooper-joe please help out

Mohammer5:
@varl Should we add a tabIndex property to our input components that defaults to 0?
Right now, the `Checkbox` and `Radio` will always have 0, which might be wrong if some apps make use of the tabindex property.

src/Checkbox/Input.js
```diff
@@ -1,27 +1,60 @@
...
+                `}</style>
+            </Fragment>
+        )
+    }
+}
 
 Input.propTypes = {
     onChange: propTypes.func.isRequired,
+    onFocus: propTypes.func.isRequired,
+    onBlur: propTypes.func.isRequired,
```
varl: Why are these required on the Input but not on the Checkbox?

varl:
> 
> 
> @varl Should we add a tabIndex property to our input components that defaults to 0?
> Right now, the `Checkbox` and `Radio` will always have 0, which might be wrong if some apps make use of the tabindex property.

Not a bad idea.

src/Checkbox/Input.js
Mohammer5: The `Input` will have to update the parent component (the `Checkbox` in this case) about the focus state, but the app that implements the checkbox might not care about the focus state at all.
If added the `isRequired` to the `SelectField`'s `Select` component to make it consistent across our input components.

varl: Ok.

src/Checkbox/index.js
```diff
@@ -9,67 +9,115 @@ import { Icon } from './Icon'
...
 
 Checkbox.propTypes = {
     onChange: propTypes.func.isRequired,
 
+    value: propTypes.string.isRequired,
     name: propTypes.string.isRequired,
     label: propTypes.string.isRequired,
 
+    onFocus: propTypes.func,
+    onBlur: propTypes.func,
```
varl: Maybe we should provide defaults to these to `null` so that React doesn't throw propType warnings when we omit them and they get sent as `undefined` to the actual `input`?

Mohammer5: Didn't notice that.. I didn't see an error in my console, but makes perfect sense.
Null will cause the same problem, so I just removed the `isRequired` from the internal input components

Mohammer5:
> > @varl Should we add a tabIndex property to our input components that defaults to 0?
> > Right now, the Checkbox and Radio will always have 0, which might be wrong if some apps make use of the tabindex property.
> 
> Not a bad idea.

@varl The only ones that are missing (and important) should be
* autocomplete
* autofocus
* readonly
* tabindex

I don't think we need to pass `form` and `list` through to the input. Agree?

varl:
>     * 
> 
> 
> I don't think we need to pass `form` and `list` through to the input. Agree?

Agreed.

varl:
Gonna let @cooper-joe have his say about the focus implementation before approving so dismissing my review.

Other than that, code looks good. 👍

cooper-joe:
Thanks for the heads up on this. I've taken a look and suggest the following focus styles:
focus
On focus:
- border color of the element is grey800 
- 1px of padding
- 1px border, blue600
This applies also when the element has been toggled on.

cooper-joe:
Also, what options do we have for styling the focus of the 'Switch' element? Would putting the same 1px blue border (with 1px padding) work here?

varl:
> Also, what options do we have for styling the focus of the 'Switch' element? Would putting the same 1px blue border (with 1px padding) work here?

I think we can do that there as well.

Mohammer5:
@cooper-joe Thanks for the styles! Where can I find the svg files? There's nothing (except colors) on invision anymore and the design-system repo only contains docs, can't find the assets.

Do we already have a focus style for the Switch?

cooper-joe:
> @cooper-joe Thanks for the styles! Where can I find the svg files? There's nothing (except colors) on invision anymore and the design-system repo only contains docs, can't find the assets.

For reference, @Mohammer5 and I discussed this in chat, focus styles will be implemented with a css border.

Mohammer5:
I've added focus styles to the `Switch`, `Checkbox` and `Radio` components.
@cooper-joe please let me know if the `Switch` focus style is ok like this

Mohammer5:
@cooper-joe 

I've changed the color code to blue600, for the Checkbox, Radio and Switch component.

The border is already set to 2px (see [here](https://github.com/dhis2/ui-core/blob/6bfd0e513125ae15912b4f08bc5556968e390d7b/src/Checkbox/Icon.js#L79) and [here](https://github.com/dhis2/ui-core/blob/6bfd0e513125ae15912b4f08bc5556968e390d7b/src/Radio/index.js#L150)).

The focus state is very hard to see for Switches that are checked:

image

varl:
@Mohammer5 If you zoom in on @cooper-joe's picture you can see that there should be 2px blue, then 1 px white, and then the form control.

2019-05-31-131130_867x389_scrot

Mohammer5:
@varl varl aka hawk-eye!
Fixed with commit [0842776](https://github.com/dhis2/ui-core/pull/230/commits/0842776de6e6be2929677ea697b020a0a135cfba)

varl:
2019-05-31-140254_428x207_scrot

It looks like the 1 px white border is transparent? I think it should be solid white.

varl:
> @varl varl aka hawk-eye!

Every pixel counts. :mag:

Mohammer5:
@varl @cooper-joe 
I replaced the transparent area with a white border

varl:
@Mohammer5 I cannot un-focus the switch though? It always has a blue border.

It also doesn't look aligned right in Firefox: 
image

There is less space to the top left of the circle than there is in the bottom right.

Mohammer5:
@varl thanks for checking that... Le fixed

Mohammer5:
@varl thanks for finding the focus bug. Fixed now

src/Checkbox/Input.js
```diff
@@ -1,29 +1,75 @@
...
+                    name={name}
+                    value={value}
+                    onChange={onChange}
+                    onFocus={onFocus}
+                    onBlur={onBlur}
+                    checked={checked}
+                    disabled={disabled}
+                    autocomplete={autocomplete}
+                    autofocus={autofocus}
+                    tabIndex={tabIndex}
```
varl: `tabIndex` is duplicated, L32 and L44: `0` and then `{tabIndex}`.

src/Checkbox/index.js
```diff
@@ -9,66 +9,124 @@ import { Icon } from './Icon'
...
-    error,
-}) => (
-    <label
-        className={cx('base', className, {
+class Checkbox extends Component {
+    constructor(props) {
+        super(props)
+
+        this.state = {
+            focus: props.focus,
```
varl: Do we need to copy the prop to state?

https://overreacted.io/writing-resilient-components/#dont-stop-the-data-flow-in-rendering

src/Radio/index.js
```diff
@@ -96,13 +168,24 @@ Radio.propTypes = {
...
     disabled: propTypes.bool,
     valid: propTypes.bool,
     warning: propTypes.bool,
     error: propTypes.bool,
+    autofocus: propTypes.bool,
+    readonly: propTypes.bool,
+}
+
+Radio.defaultProps = {
+    tabIndex: '0',
```
varl: Remove the default and let the browser handle it unless it is set explicitly.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/tabIndex

src/Switch/index.js
```diff
@@ -8,44 +8,84 @@ import { SwitchIcon } from '../icons/Switch.js'
...
+    onFocus = e => {
+        this.setState({ focus: true })
+
+        if (this.props.onFocus) {
+            this.props.onFocus(e)
+        }
+    }
+
+    onBlur = e => {
+        console.log('on blur!')
```
varl: Rogue log statement. 💀

src/Checkbox/index.js
Mohammer5: Yes, this is intentional. This way the input element can be focused without user interaction.
Contrary to `autofocus` this also works after the page has loaded.

I could add a `componentDidUpdate` to check whether the `focus` prop has changed and update the state accordingly, although I'm not certain we actually need this. As far as I know, none of our apps re-focus an input element, so I'd rather not do this until explicitly needed.

Mohammer5:
> We should avoid copying props to state on initialization (link provided).

@varl and I have talked about this and we need to keep this in order to be able to force an auto-focus on input elements after the page has loaded. We agreed to change the naming to `initialFocus` as that reflects the actual usage more precisely as well as [complies with the standard naming rules](https://reactjs.org/docs/react-component.html#constructor).

Mohammer5:
@cooper-joe I'm ready to deploy this one, could you have a look again to check if everything is ok right now?

There's one issue due to the way how browsers handle focus/blur events of form input elements:
When the `label` is clicked, the `input` gets focused.
When the label is clicked again, the `input` loses focus (because another element was clicked) but gets focused again right away because it's the associated label element that was clicked. So for a short time, the element loses the focus style.
Firefox has a bug (all other browsers behave like described above), in FF the input element loses focus when double clicked too quickly.

PR #232, opened 2019-05-22
feat(prop-types): mutually exclusive booleans

HendrikThePendric:
We discussed we need something like this. Not sure about two things:
1. I created a top-level folder called `prop-types`. Is this a good place to put this?
2. This works, but has to be set on each prop individually, so requires quite a bit of typing. It does fit in nicely with how you normally define prop-types though.

I could change the implementation a slightly and then use like this:
```javascript
const buttonSizeProp = createMutuallyExclusiveBoolProp('small', 'large')
const buttonTypeProp = createMutuallyExclusiveBoolProp('primary', 'secondary', 'destructive')

Button.propTypes = {
    small: buttonSizeProp,
    large: buttonSizeProp,

    primary: buttonTypeProp,
    secondary: buttonTypeProp,
    destructive: buttonTypeProp,
}
```

That would require a lot less typing, but is a bit less idiomatic.


Mohammer5:
Nice initiative! I prefer the version you already have, feels more like the standard behavior of prop-types.
There are a few things I'd like to discuss though:

### 1. Name of the function
`boolAndMutuallyExclusiveWith` is quite long, how about `xorPropWith`?

### 2. Rest params vs. array
I think we should provide the values as an array rather than collecting them with the rest operator:
1. It's the prop-types standard behavior (like `oneOf` or `oneOfType`)
2. It will allow us to provide a second param in the future (if ever needed) without having to refactor existing code. (Let's say we need a config object to defined that at least one must be provided, like a required state, but we don't want to print the message three times, so we specify that `isRequired` state on the first prop)

### 3. `error` variable
Although not really important in this case, but this is `mutable` style. You could just return the error inside the conditions and return `null` at the end of the function's body

### 4. "mutually exclusive" error message
I think the error message for when multiple values are given, should include which values are given. prop-types always includes that in the error message as well (expected `object` or `function`, but got `undefined`)

### 5. Checking for `bool`
Is this necessary? What's the problem with allowing other types as well? This could be a candidate for the config object (should be a function though so that type checking can be a bit more sophisticated than checking against `typeof` while the `boolAndMutuallyExclusiveWith` doesn't need to include that logic).

varl:
> ###
> 
> `boolAndMutuallyExclusiveWith` is quite long, how about `xorPropWith`?

I don't think the `With` makes sense. Just `xorProp` is fine to me. It does imply that we going with naming of logical ops like `nandProp` and `nxorProp` in the future. 😄 

I think aligning closer with the `PropTypes` API and calling it `mutuallyExclusive(...props)` would be fine.

As for the `prop-types.js` .. It's a bit too close to `PropTypes` and all we are doing is adding a validator, so maybe rename it to `prop-validators.js`?

`const buttonTypeProp = propValidators.mutuallyExclusive('primary', 'secondary', 'destructive')`

varl:
> 2. Rest params vs. array
>
> 4. "mutually exclusive" error message
> 
> 5. Checking for `bool`

I agree with these points @Mohammer5 raised.

HendrikThePendric:
Thanks for the input. Please see my reply below:
1. **Function name**: I like `mutuallyExclusive` but would like to capture the fact that the value needs to be boolean too. `mutuallyExclusiveBool` would be my preferred name. (also see 5)
2. **Rest params vs array**: I'll go with an array
3. **Returning the error**: I'll give that another shot. Can't see why it wouldn't work.... I do remember trying it yesterday and it not working. But later on I realised that to get the prop-type errors in storybook, I needed to do a page refresh all the time. I think the HMR in storybook bypasses the prop-types checks.
4. **"mutually exclusive" error message**: Sure I'll add it. Even though it doesn't add as much value as you might think. If multiple mutually exclusive props have a `true` value, this error will fire for each of them. That way, you already know exactly which ones are true.
5. **Checking for bool**: I disagree with you both on this one. Making this function more generic would make it usable in different scenarios, but at the same time the prop check looses its accuracy. We do expect boolean values, so we need to check if the value is boolean. Since you can't "chain" prop-type checks we to fully validate the prop in a single function (but I offer a solution in 6).
6. **Create a variable, or repeatedly call the function**: @Mohammer5 mentioned that he preferred my current implementation, but @varl used the alternative implementation in his example, which could mean he prefers the alternative. This also brings me back to point 5: if we use the alternative implementation, and define the exclusive prop-names as an array, we could make an alternative implementation that could look like this:
    ```javascript
    const buttonTypeProp = mutuallyExclusive(['primary', 'secondary', 'destructive'], propTypes.bool)
    ``` 
    So the function would be more generic, but you pass in a specific validator to make it accurate. I think that would be a nice solution, but have to check if and how you can manually run prop-validators. As I said, if we implement the function like this, I think we should declare a variable before the prop-types definition and use that. Because to repeatedly call this function with a more complex signature would be error-prone.
7. **Name and location**: going to put it in `src/prop-validators.js` as @varl suggested

Actually... After summarising all my thoughts above, I think I should proceed to implement this as I mention in point 6. Will start work now, so please let me know if you see things differently.

varl:
> 6\. Create a variable, or repeatedly call the function
> `const buttonTypeProp = mutuallyExclusive(['primary', 'secondary', 'destructive'], propTypes.bool)`

I like this.

varl:
>"mutually exclusive" error message

It it doesn't add value then it's just additional noise. If you already have the information and it's easy to reason about I don't care about it.

HendrikThePendric:
> I like this.

I'm glad because my last commit has implemented it that way :-)

HendrikThePendric:
I think my last commit implements all the input I received, and I'm quite happy with this implementation.

However, moving forward, we should actually go through all the components, identify mutually exclusive booleans, and add this new prop-validation...

I guess it makes sense to do that in the current branch too. Please let me know if the implementation is good now. And whether I should go through the components in this branch.

varl:
> Applying this new prop-type should be done in a separate branch.

Agreed.

PR #235, opened 2019-05-24
fix: styling adjustments

cooper-joe:
fix: color/opacity of ScreenCover
fix: color/opacity of ComponentCover
fix: added padding-right to menu labels, space needed between label and right arrow

### DIFF
--- a/src/ComponentCover/index.js
+++ b/src/ComponentCover/index.js
@@ -20,7 +20,7 @@ const ComponentCover = ({ children, className }) => (
                 width: inherit;
 
                 z-index: ${layers.applicationTop - 1};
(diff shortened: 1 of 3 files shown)


cooper-joe:
I'm unsure if adding padding-right to the label is the correct way of getting some padding between label and right chevron/arrow in menus. I wasn't sure if @Mohammer5 or @HendrikThePendric built the menu, but please take a look at my attempt!

varl:
I think it's OK. Could change it in the future if need be. Pushed a fix where we show ellipsis if the MenuItem label is very long to ensure that it doesn't run outside of the menu, and that the chevron is always visible.

PR #246, opened 2019-06-04
chore(release): v3.0.0

### LINKED ISSUE #236: Remove indeterminate linear progress
I have written up the specs for progress components today, I woud like to remove the indeterminate linear progress element. I think it would be a good idea to establish a pattern: circular loaders for indeterminate loading, linear for determinate. I think otherwise the amount of possiblities of different loader types end up diluting their meaning to just 'loading', rather than potentially communicating anything specific.

### LINKED ISSUE #237: Rename 'progress' components to 'loader'
Although progress bars are defined as `<progress>` in the spec, I think the system should define them as 'loaders'. This better communicates their intended usage. I think more often they will be used to communicate 'something is loading', rather than 'something is progressing', and therefore the name should reflect what potential devs/designers will search for. Of course, this is open for discussion!
Related: [Design docs for 'Loading indicators'](https://github.com/dhis2/design-system/blob/master/atoms/loading.md)

### LINKED ISSUE #238: Background for determinate linear progress/loader
I think the linear determinate loader/progress would communicate more clearly with a defined background that shows the 'track' of the loader/progress bar. It would then be clearer how much is left to load. I don't think linear loaders will always be used as 100% width of the parent, so this background track becomes necessary. Mockups use `background: rgba(110,122,138,0.20);`, and you can see examples of the mockups [here](https://github.com/dhis2/design-system/blob/master/atoms/loading.md)

### LINKED ISSUE #241: Adjustments to chip with icon
When an icon is present the chip label needs to have a 6px left margin instead of 12px. Would it be acceptable to have `margin-right:-6px` on the icon element? This works, but I don't know if it's invalid/gross.

### LINKED ISSUE #243: Add active state bg color to MenuItem - using pseudo element
A background color for Menu Item with class `active` was implemented here:  https://github.com/dhis2/ui-core/pull/240

However, this is not having any effect when I use the component in dashboards. When I added css for the pseudoelement:

```li:active {background-color: ${colors.grey400};}``` then I get the desired effect.

Could you add this as an additional css style?

### PR DESCRIPTION
**Pull request by varl**
Resolves #236.
Resolves #237.
Resolves #238.
Resolves #243.
Resolves #241.

* fix(loaders): rename progress components to loader components

* fix(linear loader): correct background color

* chore: update dependencies

* chore: update @dhis2/cli-style

* fix: go over focus states for inputs and fix some bugs

BREAKING CHANGE: The CircularProgress and LinearProgress have been renamed to CircularLoader and LinearLoader.
The LinearLoader does not have an indeterminate option any more.
Removes DOM attributes from component API: `autofocus`, `autocomplete`, `readonly`.

### CODE REVIEW

File: src/ButtonStrip/index.js [hunk 1]
```diff
@@ -14,7 +14,7 @@ const ButtonStrip = ({ className, children, compact, start, middle, end }) => (
 
 ButtonStrip.propTypes = {
     className: propTypes.string,
-    children: buttonArrayPropType.isRequired,
+    children: buttonArrayPropType,
```
varl: React throws an error when `isRequired` is used on this custom validator, claiming it to be `undefined`. I logged out `buttonArrayPropType` and it isn't `undefined`. Not sure what that's about.

src/Checkbox/Input.js
```diff
@@ -20,10 +20,7 @@ export class Input extends Component {
             value,
             onFocus,
             onBlur,
-            autocomplete,
```
varl: React does not like it when we use DOM attributes on React elements. I didn't expect this, but they aren't used anywhere so let's just backtrack and remove them.

src/Checkbox/Input.js
```diff
@@ -62,13 +56,11 @@ Input.propTypes = {
...
-    autocomplete: propTypes.string,
     tabIndex: propTypes.string,
 
     onFocus: propTypes.func,
     onBlur: propTypes.func,
     checked: propTypes.bool,
     disabled: propTypes.bool,
-    autofocus: propTypes.bool,
-    readonly: propTypes.bool,
+    focus: propTypes.bool,
```
varl: `focus` is used in `componentDidMount` and should be declared.

src/CircularLoader/index.js
```diff
@@ -5,7 +5,7 @@ import cx from 'classnames'
 
 import styles from './styles.js'
 
-const CircularProgress = ({ small, large, className }) => (
+const CircularLoader = ({ small, large, className }) => (
```
varl: Renamed to match the terms in https://github.com/dhis2/design-system/blob/master/atoms/loading.md

#237

src/LinearLoader/index.js
```diff
@@ -0,0 +1,65 @@
...
+                }
+
+                .determinate {
+                }
+            `}</style>
+        </div>
+    )
+}
+
+const LinearLoader = ({ amount, width, margin, className }) => {
```
varl: Renamed to match the terms in https://github.com/dhis2/design-system/blob/master/atoms/loading.md

#237

src/Radio/index.js
```diff
@@ -49,12 +49,12 @@ class Radio extends Component {
         super(props)
 
         this.state = {
-            focus: !!props.initialFocus,
+            focus: props.initialFocus,
```
varl: `initialFocus` is declared as a `bool`, so this cast shouldn't be required (it's not done in other components). There was a React error in this component, but that was related to: https://github.com/dhis2/ui-core/pull/246/files#diff-37738b57acc09fcc6cfc8af22d0933b5L124

`focus` isn't actually an attribute on an `input` element, so React threw an error for it. We use `componentDidMount` to set focus anyway so it isn't needed.

src/SelectField/Select.js
```diff
@@ -60,16 +60,6 @@ export class Select extends Component {
         }
     }
 
-    componentDidUpdate(prevProps) {
```
varl: Changed the implementation here to be more consistent with how the other fields work.

varl:
Added comments to make it a bit easier to review.

src/Checkbox/Input.js
HendrikThePendric: Well spotted. I did a commit that adds this to Icon too.

src/ButtonStrip/index.js
varl: Thanks.

src/Checkbox/Input.js
varl: Thanks. 👍

src/LinearLoader/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+const Progress = ({ amount }) => {
+    return (
+        <div>
+            <style jsx>{`
+                div {
+                    width: ${amount}%;
+                }
+            `}</style>
+            <style jsx>{`
```
Mohammer5: The two style tags can be combined to one

src/LinearLoader/index.js
```diff
@@ -0,0 +1,65 @@
...
+                div {
+                    background-color: ${theme.primary600};
+                    transition: width 0.3s linear;
+                    height: 4px;
+                }
+
+                .determinate {
```
Mohammer5: Empty, can be removed

src/LinearLoader/index.js
```diff
@@ -0,0 +1,65 @@
...
+            <style jsx>{`
+                div {
+                    display: block;
+                    overflow: hidden;
+                    overflow-x: hidden;
+                    overflow-y: hidden;
+                    background-color: rgba(110, 122, 138, 0.2);
+                }
+            `}</style>
+            <style jsx>{`
```
Mohammer5: Both `style` tabs cam be combined

src/LinearLoader/index.js
```diff
@@ -0,0 +1,65 @@
...
+                width: ${width};
+                margin: ${margin};
+            `}</style>
+        </div>
+    )
+}
+
+LinearLoader.defaultProps = {
+    amount: 0,
```
Mohammer5: I think the amount could/should be required

src/LinearLoader/index.js
HendrikThePendric: I think the current implementation is following the styled-jsx guidelines:
> That said when your CSS is mostly static we recommend to split it up in static and dynamic styles and use two separate style tags so that, when changing, only the dynamic parts are recomputed/rendered.

https://github.com/zeit/styled-jsx#dynamic-styles

src/LinearLoader/index.js
varl: True. I had a version that was backwards compatible. That's not a concern anymore so will force it!

src/LinearLoader/index.js
varl: Yes, dynamic styles should be in a separate `style` element. It's dynamic because it uses a `prop` and cannot be resolved through a constant.

src/LinearLoader/index.js
varl: What the.. Will fix.

src/LinearLoader/index.js
varl: No, the second one is dynamic since it uses `props`.

src/LinearLoader/index.js
varl: Agreed.

src/Chip/Icon.js
```diff
@@ -13,9 +13,16 @@ export const Icon = ({ icon }) => {
                     width: 24px;
                     height: 24px;
                     margin-left: 4px;
+                    margin-right: -6px;
                     border-radius: 50%;
                     overflow: hidden;
                 }
+
+                /* force icons to fit in our icon drawer */
+                span > :global(*) {
```
HendrikThePendric: If possible change this to a more specific selector, as css it's interpreted from right to left, so now is very expensive. (as @Mohammer5 pointed out to me a few weeks ago)

varl: I don't know how we can predict what will go in here?

varl: I'd be happy to change this if there are any alternatives. We can "guess" that most likely there will be an `` or `<svg>` tag passed in, but it's hard to guarantee it. We could just ignore it and let the sub element overflow?

It doesn't look too nice though.

On the performance side. I'm not advocating that we should make it slow, speed is a priority. But CSS selector speed is much, _much_, **much** faster than triggering layout reflows which [a lot of CSS properties do](https://csstriggers.com/). A short list of the expensive ones which trigger layout thrashing:

```
    border-radius: 5px;
    padding: 3px;
    box-shadow: 0 5px 5px #000;
    transform: rotate(10deg);
```
 
I do not think we need to worry too much about CSS selector speed as long as the selectors are kept reasonable:

E.g. I think `span > *` is OK while something like `div:hover[data-foo="bar"].name:nth-child(5)::after` would **not** be OK.

Some articles backing up my claims:

- https://benfrain.com/css-performance-revisited-selectors-bloat-expensive-styles
- https://calendar.perfplanet.com/2011/css-selector-performance-has-changed-for-the-better/
- http://perfectionkills.com/profiling-css-for-fun-and-profit-optimization-notes/
- http://www.stevesouders.com/blog/2009/03/10/performance-impact-of-css-selectors/

Note that these range from 2009-2014 so a lot of those benchmarks are done on _very_ slow browsers like IE6-9 and Firefox 3.0. I've re-run the benchmarks I could easily run and selector speed today is a lot faster than it was.

Layout reflows are still expensive.

varl: Thinking on this some more, it's not certain that the aspect ratio will be 1:1 for all elements passed in as `icon`, so I think we had it good enough with just the overflow. Added a docstring instead.

PR #247, opened 2019-06-05
chore: custom prop types

HendrikThePendric:
I've reworked our custom prop-type validators:
- They are now more generic
- They support `.isRequired`
- They have been moved into their own folder `./src/prop-validators`

Usage of `instanceOfComponent` is illustrated in the `src/ButtonStrip/index.js` line 18.

Usage of `mutuallyExclusive` would be like this:
```javascript
const buttonTypeProp = mutuallyExclusive(['primary', 'secondary', 'destructive'], propTypes.bool)
```
As discussed in https://github.com/dhis2/ui-core/pull/232#issuecomment-495137042)

And, while this PR would introduce these custom prop-types, we should also implement them where needed. Perhaps in the `next` branch?


HendrikThePendric:
Now ready to be merged into `next`

PR #249, opened 2019-06-06
feat: add spacing scale (and fix a few bugs)

varl:
### Spacing

This adds the spacing scale from: https://github.com/dhis2/design-system/blob/master/principles/spacing-alignment.md

I've implemented it for all the margin/padding declarations I could find. In some places I've had to adjust the amount to either one above or one below. E.g. if it said `20px` we have to jump to `24px` or drop to `16px`.

I chose `spacers` to match the name `layers`, any better options?

### Chip

This also re-adds the `span > :global(*)` selector for the Chip (but with `max-height` and `max-width`). I think it's fine, the reasoning is here: https://github.com/dhis2/ui-core/pull/246#discussion_r290871770

We need it because upon further testing I realised that if the Icon is too large, it actually breaks the layout, and we need to stop that from happening.

### Modal

The scrollable Modal was not behaving properly, the content expanded way beyond my viewport so the `overflow-y` rule never triggered. I have added a `max-height: 95vh` to the ModalCard next to the `height: auto` declaration.


varl:
> One minor thing: in a lot of places you are importing `from '../theme.js'` but this can just be `from '../theme'`.

I know, I use the extension to indicate that I am importing directly from a file, and not a directory through automatic `index.js` resolution to avoid confusing situations:

```
import { foo } from './bar'

// do we mean
import { foo } from './bar.js'

// or do we mean
import { foo } from './bar/index.js'
```

Using the `.js` extension for direct file imports avoids the ambiguity. For components the directory is usually uppercase, but we often have lowercase `utils/` and `helpers/`, and they are easy to mix up with `utils.js` and `helpers.js`. Even more confusing if you have both the file and the dir..

Mohammer5:
> Using the .js extension for direct file imports avoids the ambiguity. 

Just another reason why I don't like `index.js` files..
I still think that it's better to use `src/Button.js` instead of `src/Button/index.js`.

We've talked about this already and came to the conclusion that it's not worth the effort to rename all this stuff, just wanted to mention it again.

varl:
> Just another reason why I don't like `index.js` files..

Hehe, yeah. There's a talk by the NodeJS creator (he quit the project early when the community took it in directions he didn't like) where he revisits a lot the mistakes he thought he made when creating NodeJS.

Bare imports (no suffix) was one. Another was the `index.js` mechanism. Personally I don't mind the name `index.js`, but I do prefer importing e.g. `Button/index.js`. The `index.js` is handy to know what component is the actual entry-point, and the convention `Button/Button.js` is more silly to me than `Button/index.js`...

> We've talked about this already and came to the conclusion that it's not worth the effort to rename all this stuff, just wanted to mention it again.

Yeah, there are more pressing matters, like tests to do first.

Mohammer5:
>  the convention Button/Button.js is more silly to me than Button/index.js...

`Button/Button.js`: This is exactly what I wouldn't do! The `Button.js` should be in the same folder as the `Button/` folder:

```
|- src/
|--- Button/
|--- Button.js
```

`src/Button.js` makes sense. Everything related to the button namespace would be inside the button folder.

varl:
> 
> 
> But then we should improve consistency. There are still quite a few files that are importing `from '../theme'`.

OK. Done in a109b4d.

varl:
> `Button/Button.js`: This is exactly what I wouldn't do! The `Button.js` should be in the same folder as the `Button/` folder:

Aight, thanks for the clarification.

PR #257, opened 2019-06-06
Alert bar

HendrikThePendric:
So I have the `AlertBar` and `AlertStack` working according to the specs. I've also cleaned up all the unused vars we had in this project. 

Currently this is a draft PR, because:
- I don't like the way it looks when you dismiss 1 AlertBar, and let the other ones auto-hide. But I think this is quite likely just a result of the alert not being properly managed (i.e. removed when they are no longer needed etc). I'd like to make a POC AlertManager to see how it behaves in there.
- I still need write proper stories and include a decent readme
- I still need to work on disabling auto-hide under certain conditions

But there is already quite a lot there to review, so I thought I prepare this before I start my 4 day weekend :-)

CU in Oslo on Tuesday


varl:
They aren't removed from the DOM when they are hidden. They should be unmounted completely, and then they should stack nicely even if you remove one in the middle.

src/AlertBar/index.js
```diff
@@ -0,0 +1,112 @@
...
+        if (this.props.onHidden) {
+            this.onHiddenTimeout = setTimeout(
+                this.props.onHidden,
+                ANIMATION_TIME
+            )
+        }
+    }
+
+    show() {
+        setTimeout(() => {
```
varl: Why use `setTimeout` here? Wouldn't `requestAnimationFrame` be more appropriate?

HendrikThePendric: Yes, good suggestion.

HendrikThePendric:
> They aren't removed from the DOM when they are hidden.
Yeah, good point. I was thinking about the smart component already, and that it would remove it not just the DOM nodes, but also the component instance completely. But even just rendering `null` when done really helps here. You were spot-on: deleting alerts looks good now.

So... I've implemented your two suggestions, added some more logic for when to auto-hide, and the stories are good now. All that is missing is the readme files. I'll fix these up tomorrow.

src/AlertBar/index.js
```diff
@@ -0,0 +1,132 @@
...
+import { Message } from './Message'
+import { mutuallyExclusive } from '../prop-validators'
+
+class AlertBar extends PureComponent {
+    state = {
+        visible: false,
+        hidden: false,
+    }
+
+    componentDidMount() {
```
Mohammer5: I think this stuff should go into it's own function that's called both when the component did mount as well as when the component did update && the visible prop changed to true (in case someone using this library is rendering error messages (like for a form) and shows/hides/shows it (like a "you have to fill out all inputs" message for example)

src/prop-validators/arrayWithLength.js
```diff
@@ -0,0 +1,57 @@
...
+        return new Error(`${propName} is required.`)
+    }
+
+    if (arr && !Array.isArray(arr)) {
+        return new Error(`${propName} is not an array.`)
+    }
+
+    if (arr && arr.length > max) {
+        return new Error(
+            `${propName} array exceeds the maximum length of ${max}`
```
Mohammer5: Might be useful to know the actual length of the supplied array for debugging purposes

src/AlertBar/Icon.js
```diff
@@ -0,0 +1,49 @@
...
+import propTypes from 'prop-types'
+import { Valid, Warning, Error, Info } from '../icons/Status'
+import { spacers } from '../theme.js'
+
+const Icon = ({ icon, success, warning, critical }) => {
+    if (icon === false) {
+        return null
+    }
+
+    let IconComponent
```
Mohammer5: The naming doesn't really match the convention. The assigned value is not a component anymore but already the computed element (component would be `Error` while the element is `<Error />`), which is why I think this one needs to be camelCase (plus the name `Component` is confusing, I'd expect that the jsx code will do `<IconComponent />`.

React only enforces the first letter to be capitalized when using the `<... />` style, not when doing `{component}`.

PR #277, opened 2019-06-13
feat(alertbar): add the alert bar and alert stack

HendrikThePendric:
This is a second attempt to introduce the AlertBar and the AlertStack. I prematurely merged the first attempt in and this time I will wait until I have approval.


src/AlertBar/Icon.js
```diff
@@ -0,0 +1,49 @@
...
+import { Valid, Warning, Error, Info } from '../icons/Status'
+import { spacers } from '../theme.js'
+
+const Icon = ({ icon, success, warning, critical }) => {
+    if (icon === false) {
+        return null
+    }
+
+    let IconComponent
+    if (isValidElement(icon)) {
```
Mohammer5: Normally this would be perfectly fine to use it like this, but as there's also a `<Valid />` component, I'd like to see `React.isValidElement(icon)` here instead (just to reduce the confusion)

src/AlertBar/index.js
```diff
@@ -0,0 +1,146 @@
...
+    propTypes.bool
+)
+
+AlertBar.propTypes = {
+    className: propTypes.string,
+    children: propTypes.string.isRequired,
+    success: variantPropType,
+    warning: variantPropType,
+    critical: variantPropType,
+    icon: iconPropType,
```
Mohammer5: This is ambiguous, I'd prefer to have a 'noIcon` prop, but it's not too important. I'd reduce the "complexity" in the `Icon` component as well (complexity in quotes because it's not complex at all)

src/prop-validators/arrayWithLength.js
```diff
@@ -0,0 +1,59 @@
...
+    if (arr && arr.length > max) {
+        return new Error(
+            // prettier-ignore
+            `${propName} array has a length of ${arr.length}, but the maximum is ${max}`
+        )
+    }
+
+    if (arr && arr.length < min) {
+        return new Error(
+            // prettier-ignore
```
varl: Why are the ignores necessary?

src/AlertStack/styles.js
```diff
@@ -0,0 +1,19 @@
...
+export default css`
+    div {
+        position: fixed;
+        top: auto;
+        right: auto;
+        bottom: 0;
+        left: 50%;
+        transform: translateX(-50%);
+
+        z-index: 9999;
```
varl: Use `layers.alert` from `theme.js`: https://github.com/dhis2/ui-core/blob/master/src/theme.js#L120

src/AlertBar/index.js
HendrikThePendric: I see your point. However, @varl and I discussed this before. We are introducing some complexity here, but we thought it provides a very convenient API (see the Icons story for an illustration).

src/prop-validators/arrayWithLength.js
HendrikThePendric: I added these ignore comments because I thought that the way that prettier is breaking up these template strings makes them very hard to read. Prettier doesn't have a setting to change its behaviour in this regard https://github.com/prettier/prettier/issues/3280#issuecomment-345105820

We can remove it if you prefer, or I could try to write a more succinct message.

HendrikThePendric: For illustration, this is how prettier's formatting looks:
```javascript
`${propName} array has a length of ${
    arr.length
}, but the maximum is ${max}`
```

varl: Not too bad, I was wondering if there was a serious problem. You can keep the ignores.

PR #283, opened 2019-06-14
fix: remove spacing left to the checkbox icon in FF

Mohammer5:


src/Checkbox/Input.js
```diff
@@ -39,15 +39,13 @@ export class Input extends Component {
                 />
 
                 <style jsx>{`
-                    input {
-                        display: block;
-                        opacity: 0;
+                    div {
                         width: 0;
                         height: 0;
-                        outline: 0;
```
varl: Don't we need outline set to `0` for other browsers?

varl: I've tested chrome, edge, firefox and looks good.

PR #294, opened 2019-06-15
fix: ship the dev and prod builds

varl:
When `ui-widgets` (or another library) loads the production (`optimizeForSpeed=true`) components from `ui-core` that are built with `optimizeForSpeed` before loading the local components (in `ui-widgets`), styled-jsx first sees those speed optimized components and sets the entire styled-jsx singleton to parse all components as `optimizedForSpeed`.

This does not work for local components which are not pre-parsed and requires further string splitting for style declarations.

E.g. `span { display: block; } div { display: flex; }` in `optimizedForSpeed` will be treated as a single CSS rule, which causes the `CSSStyleSheet.insertRule()` to choke, since it expects a single rule at a time.

When styled-jsx is not `optimizedForSpeed`, the same line `span { display: block; } div { display: flex; }` would be split into two rules:

`span { display: block; }` and `div { display: flex; }`, and each would be passed to `insertRule` one at a time.

So when the styled-jsx singleton first sees a `optmizedForSpeed` component, it assumes that all of them are also optimized for speed, and if they are not it will crash when it encounters non-optimized components.

This should not be problem when [styled-jsx/#64](https://github.com/zeit/styled-jsx/issues/64) is resolved, until then, we can work around it by exposing both the development (non-optimized) and production (optimized) builds in the `ui-core` bundle.

This only changes the potential entry-points, everything behind the entry-point is just as it was so nothing should change for the implementations.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,5 @@
+if (process.env.NODE_ENV === 'production') {
+    module.exports = require('./prod/index.js')
+} else {
+    module.exports = require('./dev/index.js')
+}
--- a/package.json
+++ b/package.json
@@ -13,12 +13,25 @@
   },
   "scripts": {
     "prebuild": "rm -rf ./build/*",
-    "build:commonjs": "BABEL_ENV=commonjs babel src --out-dir ./build/cjs --copy-files --verbose",
-    "build:modules": "BABEL_ENV=modules babel src --out-dir ./build/es --copy-files --verbose",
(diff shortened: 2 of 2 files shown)


varl:
This is how React does it as well: https://github.com/facebook/react/blob/master/packages/react/npm/index.js

varl:
This breaks tree-shaking. :disappointed:

varl:
Will need to figure something else out which doesn't break tree-shaking.

PR #297, opened 2019-06-19
feat: extract and expose MenuList from Menu component

Mohammer5:


src/Menu/index.js
```diff
@@ -1,29 +1,24 @@
...
 const Menu = ({ children, className }) => (
     <Card>
-        <ul className={className}>{children}</ul>
-        <style jsx>{`
-            ul {
-                display: block;
-                position: relative;
-                width: 100%;
-                margin: 0;
+        <div>
```
varl: Should the `className` be added to the `div` or the `MenuList`? :thinking:

HendrikThePendric:
I feel like atm we have one too many layer here, for example:
DropMenu>Menu>MenuList>MenuItems

I guess the reasoning can be summarised like this:
1. DropMenu renders the menu in a Pop
2. Menu wraps the MenuList in a Card
3. MenuList ensures the correct list layout is guaranteed
4. MenuItem speaks for itself

Personally I would prefer it if:
1. The DropMenu wraps the Menu in a Card & renders everything in a Pop
2. Menu is removed, or MenuList is removed and its logic is moved to Menu

I think this many layers will just cause confusion.

Also, in the current implementation what menu does is just render a Card with padding. So the component-name doesn't match what the component does.

varl:
It's not a given that the DropMenu should include a Card though, and the design-system has a card-like wrapper around the Menu.

Mohammer5:
@HendrikThePendric 

> So the component-name doesn't match what the component does.

It matches the name of the design system (https://github.com/dhis2/design-system/blob/master/atoms/menu.md)

Which should probably answer all the other questions as well

HendrikThePendric:
I'm still under the impression we are adding layers of fluff here, and that this implementation of `Menu` does not match the design system:
> ... A menu provides user access to *options* that are available when clicking the menu toggle. ...

> ... Menu items are *options* or actions contained inside a menu. ...

So according to the design system the _Menu is the List_

HendrikThePendric:
Anyway, if you both feel strongly about going this, way: go for it. I will disagree and commit.

varl:
> Anyway, if you both feel strongly about going this, way: go for it. I will disagree and commit.

The strongest feeling I have to this is backwards compatibility. I don't think changing the `Menu` is important enough to warrant a breaking change version bump.

Other than that, what is driving this change is that the `Menu` currently does too much, so what we are doing is pulling out the list into its own component so it can be re-used instead of duplicated.

Any user that is using the `Menu` today can keep doing what they were doing with just `Menu>MenuItem`.

This just allows us to use the `MenuList` so that in certain scenarios we can do `MenuList>MenuItem` as well. Since the `MenuList` is an `ul` I think this is OK for now.

It would be nice to avoid two extra `div`s in this PR though. Not sure if i prefer extra `div`s or a `resolve`d class name passed to `MenuList`. 🤔

HendrikThePendric:
Fair enough. OK. 👍 

I would also like to see some prop validation here, so that the MenuList children should always be a single, or an array of MenuItem component instances.

src/MenuList/index.js
```diff
@@ -0,0 +1,27 @@
...
+                padding: 0;
+                user-select: none;
+            }
+        `}</style>
+    </ul>
+)
+
+MenuList.propTypes = {
+    className: propTypes.string,
+    children: propTypes.any.isRequired,
```
HendrikThePendric: only accept instance(s) of MenuItem?

varl: That would be a breaking change.

PR #304, opened 2019-06-24
feat: add dumb table component

Mohammer5:
This will add a dumb, but responsive table component.
It'll be necessary to build the data list table widget that's used for all the list view pages.

The reason why there's a `<TableCellText />` component is so that the implementing app can decide whether it wants to render plain text and use the previously mentioned component or render something custom.


Mohammer5:
@cooper-joe Please have a look at the responsive styles

cooper-joe:
Very nice @Mohammer5. Generally I think this looks good. I have some specific feedback:

- <400 breakpoint is disrupting the structure of the table too much, I think. All of the fields are split into individual boxes and it isn't clear what is a field and what is a row. I've done some testing and I think the <768 breakpoint actually works all the way down to 320. [I've pushed a fix](https://github.com/dhis2/ui-core/pull/310) which removes the <400 styles, but you may wish to take a look and see if I've messed some stuff up here.

- Responsive tables should be the default, but I think we should allow a non-responsive type too. Would this dumb wrapper be the place for this? Or should it be it's own component? On that non-responsive component it could take a fixed width, or inherit from the parent. Cells would condense to their minimum and then the table wrapper would scroll horizontally.

Mohammer5:
@cooper-joe I've merged your PR into the branch, looks good!
I'll add a `staticLayout` prop to prevent responsive styles.
Btw, there's a container around the table which will have a horizontal scrollbar if the table is too wide to be displayed. This way the page itself won't have vertical scrollbars and the layout of the page will stay unaffected.

Mohammer5:
@cooper-joe I added the `staticLayout` prop. Seems to work fine, story has been added as well

src/Table/TableCell.js
```diff
@@ -0,0 +1,106 @@
...
+                }
+            }
+        `}</style>
+    </td>
+)
+
+export const TableCell = ({ children, title, colSpan, rowSpan }) => (
+    <Consumer>
+        {({ staticLayout }) => {
+            const TableCell = staticLayout
```
amcgee: I'd recommend not overloading the parent component name here, use something like `TableCellComponent` maybe?

src/Table/TableCell.js
```diff
@@ -0,0 +1,106 @@
...
+                    display: block;
+                    padding: 0;
+                    min-height:32px;
+                }
+            }
+        `}</style>
+    </td>
+)
+
+export const TableCell = ({ children, title, colSpan, rowSpan }) => (
```
amcgee: I understand why we have a `title` property on a cell, but it makes for pretty verbose and unwieldy usage.  To make a simple 4x4 table with a header I need to duplicate 4x (and maintain those duplicates) of all 4 column names.  That's 16 redundant values, and it's much worse for more complicated situations.

src/Table/TableCellText.js
```diff
@@ -0,0 +1,57 @@
...
+
+const TableCellTextResponsive = ({ label }) => (
+    <span>
+        {label}
+        <style jsx>{tableCellTextStyles}</style>
+        <style jsx>{tableCellTextStylesResponsive}</style>
+    </span>
+)
+
+export const TableCellText = ({ label }) => (
```
amcgee: Is there ever a case where `TableCell` might be used without `TableCellText`?  If not, it would be nice to remove the boilerplate nexting of a `TableCellText` within every `TableCell`.  If there might be cases where it isn't used, maybe that could be encapsulated in props to the `TableCell`?

Additionally, wouldn't you use this same wrapper for non-text content like an image?  If we keep the explicit child it should probably be called `TableCellContent` or something more generic.

stories/Table.stories.js
```diff
@@ -0,0 +1,627 @@
...
+            <TableHead>
+                <TableRow>
+                    <TableCellHead>
+                        <TableCellText label="First name" />
+                    </TableCellHead>
+                    <TableCellHead>
+                        <TableCellText label="Last name" />
+                    </TableCellHead>
+                    <TableCellHead>
+                        <TableCellText label="Last name" />
```
amcgee: Minor nit-pick: all of the column headers are "LastName" instead of their intended values (Incident Date, Last Updated, etc.).  This also exposes the danger of duplicating the Title in multiple places - they will get out of sync.

src/Table/TableCell.js
Mohammer5: Oops.. that was a mistake.. thanks

src/Table/TableCell.js
Mohammer5: If you want a simple table, you can just use the table with `staticLayout`, then you don't need that one anymore

amcgee: Yeah but then it's static and not responsive...

src/Table/TableCellText.js
amcgee: I see your opening comment now, but my question still stands.  I think I would prefer a singular `TableCell` component with an optional `noPadding={true}` prop or something to skip the defaults (which should be default).

Mohammer5: > Is there ever a case where TableCell might be used without TableCellText?

Yes, it can be buttons, icons, `Chip`s, etc.

> Additionally, wouldn't you use this same wrapper for non-text content like an image? If we keep the explicit child it should probably be called TableCellContent or something more generic.

The component contains mostly css rules for text. I supposed it'd work for images (and other inline-block elements) as well.

I'll play around a bit and try to figure out if this can be incorporated into the table cell component

src/Table/TableCell.js
amcgee: You could accomplish this with `:nth-child` CSS rules on the `Table` component, set on the `Table` context by the order of the `Head` components

amcgee: (though that introduces some implicit dependence, it could also be accomplished independently on the `TableCell` component with some context voodoo)

stories/Table.stories.js
Mohammer5: > This also exposes the danger of duplicating the Title in multiple places - they will get out of sync.

True, we could do this dynamically with context as well.. Don't know if that's the right solution here, would be nice if we had these values without having to pass them to every `TableCell`. 

If context is the right thing, I'm wondering about how to change the thead / th components.
It's tempting to just pass an array with strings, but that won't work once one of the head cells needs to have a different rowSpan/colSpan. Extracting it from the children feels wrong..

amcgee: I'm not *sure* this is right, but it is possible.  It also removes the need for an explicit `TableCellHead` component, which if we keep it should probably be a `TableRowHead` with a bunch of `TableCell` children?

amcgee: Maybe title and label are the same thing but you could have anything you'd like...

amcgee: Don't know if this is getting too "centralized"

Mohammer5: If this table will be used for stuff like the pivot tables, then the thead will have multiple rows.
I think that the `ui-core` component should eventually be used for all apps, so I guess it's not really an option because then things will get very complicated.

@varl any thoughts on this?

amcgee: I mean nested / multi-row headers could also be supported `[ [ 'category', [ 'subCat1', 'subCat2' ] ] ]` (just as an example, maybe with some nice wrapper functions or objects to avoid lots of nested arrays)

Mohammer5:
@amcgee @HendrikThePendric @varl 
I've update the component according to the feedback. Please have a look again :)

Mohammer5:
@cooper-joe could you have a look at the table? I'm just waiting for some feedback

cooper-joe:
> @cooper-joe could you have a look at the table? I'm just waiting for some feedback

@Mohammer5 Looking good to me. Something I notice is missing is the hover style for table rows. Is this because of storybook? I couldn't find the relevant css added anywhere. The intended style is `background: blue050;` for hovered rows. 

 Or, where you thinking that this would be best defined by apps themselves? My thoughts were that hover style doesn't necessarily indicate 'clickable', it can be useful for highlighting rows for easy reading too. Clickable rows should get `cursor: pointer;` on hover, but I'm not sure how we'd define that in this component?

Mohammer5:
@cooper-joe I see, to me hover styled meant that it was clickable, but I get your point (so I thought the hover style should come from other components, like the data element list table component)

I can add the hover style!

src/Table/extractHeaderLabels.js
```diff
@@ -0,0 +1,71 @@
...
+            col.props.colSpan
+                ? total + parseInt(col.props.colSpan, 10)
+                : total + 1,
+        0
+    )
+
+const mapCellsToLabels = row => {
+    let labels = []
+
+    for (let i = 0, count = row.length; i < count; ++i) {
```
ismay: Maybe I'm missing something obvious, but why are you not comparing `i < row.length` directly here?

Mohammer5: It's just a habit: https://stackoverflow.com/questions/17989270/for-loop-performance-storing-array-length-in-a-variable

I guess as tables can get ridiculously long with dhis2, this still might be valid, but I don't really have an opinion on this

ismay: Ah ok. Yeah me neither. It did make me wonder what was going on there, but maybe that's just me.

ismay: Didn't know that that helped performance.

src/Table/Table.js
```diff
@@ -0,0 +1,68 @@
...
+const tableStylesResponsive = css`
+    @media (max-width: 768px) {
+        table {
+            display: block;
+            border: 0;
+        }
+    }
+`
+
+const TableRepsonsive = ({ children }) => (
```
ismay: Typo `TableRepsonsive`

varl:
What's the argument for a responsive Table component over one that is scrollable?

Example: https://codepen.io/varl/pen/pMpLyb

ismay:
> What's the argument for a responsive Table component over one that is scrollable?
> 
> Example: https://codepen.io/varl/pen/pMpLyb

If you ask me I'd say that with a responsive one the labels are always close to the data. Whereas with a scrollable one, the label for the cell you're looking at might be very far away.

Mohammer5:
> What's the argument for a responsive Table component over one that is scrollable?

Exactly what @ismay said. If that behavior is not wanted, there's a `staticLayout` prop on the `Table` to prevent this behavior and keep the table static

Mohammer5:
> I was thinking about the fixed background color. Wondering if it'd be nice if it would adapt to the background. But I guess that it isn't bad for it to set its own background.

@cooper-joe ?

> I see we're using values here for colors and spacing that aren't from the theme. Do we not want import everything from there? Seems nice to me to have everything come from there.

@ismay I've replaced the plain color values and some spacing values with values from `theme.js`
@varl and I have talked about this at some point and came to the conclusion that the spacing values make sense to add space between components while most components will use different values internally to match the specification

varl:
Good point.

What I am after is if we want to go down a breakpoint driven approach to support mobile devices, or something else?

There are other components which will need breakpoints, and from this PR and on (and retroactively), @cooper-joe will have to design everything with additional breakpoints in mind.

Using breakpoints is a well-proven method to support a range of devices, as such I'm OK with it, but there are non-technical consequences of introducing it.

If we wrap the table so it's viewable (sub-optimally) in a scrollbox on a mobile devices, we can defer commitment to using a breakpoint driven strategy for designing components.

Mohammer5:
> If we wrap the table so it's viewable (sub-optimally) in a scrollbox on a mobile devices, we can defer commitment to using a breakpoint driven strategy for designing components.

I'd like to get the layout components going. And then base all responsive decision around those component. So only apps using (or components inside) the layout components would have to think about responsive styles.
That way we'd have a consistent responsive style across apps as well.

For now, the table itself is already wrapped with a div that's `overflow: auto` for when using `staticLayout`, so we could either set that one to true by default or simply disable responsive styles for now until we have a proper solution how to handle responsive styles in a consistent and manageable way across apps (if I got your intention right?)

Side note:
From my perspective I just wanted to be a bit more responsive without really putting a focus on it now to make our apps usable at least to some extend on mobile devices (and the new maintenance app hopefully will be responsive, although it's the least important app in that regard)

cooper-joe:
> > I was thinking about the fixed background color. Wondering if it'd be nice if it would adapt to the background. But I guess that it isn't bad for it to set its own background.
> 
> @cooper-joe ?

I think keeping the fixed background is wise here, we want to make sure the data is always accessible. There are very few situations where a table shouldn't be using white as the bg. (Apart form aesthetics ;) )

> What I am after is if we want to go down a breakpoint driven approach to support mobile devices, or something else?

Good point, we do want to make sure we have agreed on an approach before committing to it.

> For now, the table itself is already wrapped with a div that's overflow: auto for when using staticLayout, so we could either set that one to true by default or simply disable responsive styles for now until we have a proper solution how to handle responsive styles in a consistent and manageable way across apps

This is my thinking too. 'staticLayout' would certainly be the default and I would encourage the use of static with scrolling in the design docs.

I think the 'responsive' styles are actually more useful for embedded tables than mobile, at least for the time being. Tables are often embedded in dashboard items/tracker widgets. These widgets are user resizeable. Of course static tables with scrollable overflow will work in these situations, but providing an option for responsive tables here could help gauge the need/requirements moving forward. This isn't exactly a well thought out strategy though, so I would be ok with disabling responsive styles and waiting until we've made some decisions.

varl:
> > If we wrap the table so it's viewable (sub-optimally) in a scrollbox on a mobile devices, we can defer commitment to using a breakpoint driven strategy for designing components.
> 
> I'd like to get the layout components going. And then base all responsive decision around those component. So only apps using (or components inside) the layout components would have to think about responsive styles.
> That way we'd have a consistent responsive style across apps as well.

Interesting. 👍 

> For now, the table itself is already wrapped with a div that's `overflow: auto` for when using `staticLayout`, so we could either set that one to true by default or simply disable responsive styles for now until we have a proper solution how to handle responsive styles in a consistent and manageable way across apps (if I got your intention right?)

Sure. The kicker is that it is harder to remove than to add, so if we aren't sure about the it, we shouldn't add it in the first place (🏏🐴).

> Side note:
> From my perspective I just wanted to be a bit more responsive without really putting a focus on it now to make our apps usable at least to some extend on mobile devices (and the new maintenance app hopefully will be responsive, although it's the least important app in that regard)

I get it. 😃 

I don't think that "without really putting a focus on it" is going to work for us when it comes to the **core** UI library since the consequences of what we do in this library already has widespread effects.

HISP India and Uganda (https://github.com/dhis2/ui-core/network/dependents) is already using **ui-core** for example.

**If we put something in _ui-core_ we should be convinced that it is the way to move forward for us.**

Since this PR adds breakpoints to the mix (this is new for **ui-core**) we should think deeply about it before merging it in.

varl:
In responsive mode, I get a label for the footer above the button:

image

That looks like a bug to me, since the label has nothing to do with the button.

ismay:
> Since this PR adds breakpoints to the mix (this is new for ui-core) we should think deeply about it before merging it in

Well main thing for me is that the breakpoint you want is going to depend on the content you're rendering in the table. If you're rendering a full width table with only two columns you might want it to break at a different point than when you're rendering 20 columns.

Our fixed breakpoint does not make sense to me in that respect.

varl:
> > What I am after is if we want to go down a breakpoint driven approach to support mobile devices, or something else?
> 
> Good point, we do want to make sure we have agreed on an approach before committing to it.

This is a question for designers as much as for the developers. Does creating breakpoints fit into your workflow?

I've experienced that when designers don't have a strategy for breakpoints in their workflows the breakpoints become arbitrary and divergent over time. That yields a misbehaving UI down the road.

From my perspective I'm happy with (pixel-based) breakpoints; they are well-proven enough to be deemed as a "work-horse" rather than a "show-horse", which is a tenet for **ui-core**.

varl:
> 
> 
> > Since this PR adds breakpoints to the mix (this is new for ui-core) we should think deeply about it before merging it in
> 
> Well main thing for me is that the breakpoint you want is going to depend on the content you're rendering in the table. If you're rendering a full width table with only two columns you might want it to break at a different point than when you're rendering 20 columns.
> 
> Our fixed breakpoint does not make sense to me in that respect.

Oh man, opening the API to allow for dynamic breakpoints is a terrifying prospect. 😰

I'd like to restrict the case to either you use the responsive one with our (consistently designed) breakpoints, or you go for the static one. This would be a trade-off we choose, in that case.

cooper-joe:
> > > What I am after is if we want to go down a breakpoint driven approach to support mobile devices, or something else?
> > 
> > 
> > Good point, we do want to make sure we have agreed on an approach before committing to it.
> 
> This is a question for designers as much as for the developers. Does creating breakpoints fit into your workflow?
> 
> I've experienced that when designers don't have a strategy for breakpoints in their workflows the breakpoints become arbitrary and divergent over time. That yields a misbehaving UI down the road.
> 
> From my perspective I'm happy with (pixel-based) breakpoints; they are well-proven enough to be deemed as a "work-horse" rather than a "show-horse", which is a tenet for **ui-core**.

Indeed, I've also experienced the same. I try to look at breakpoints as 'user driven'. For example, What is a user doing, where are they, what are their limitations when assigned to the `@<768` breakpoint? To set up a strategy that will survive the constant development of DHIS2 I think we need some input from management on the plans for mobile expansion (apps, PWA, none?). 

That said, I think most of the molecules/organisms in `ui-core` would benefit from fluid/adaptive resizing with most elements set to percentage widths of the parent. I think pixel-based breakpoints make sense in `ui-layout` (is that a thing?). Here we would define pixel-based breakpoints based on the user behaviour as I mentioned above. Within those breakpoints I imagine `ui-core` components fluidly adapting. But I am open to imagining how the workflow could be defined differently.

Mohammer5:
> Our fixed breakpoint does not make sense to me in that respect.

Ok, so I could add `showLabels` prop which will force the mobile view regardless of the viewport.
I can imagine that it makes because of this:

> I think the 'responsive' styles are actually more useful for embedded tables than mobile, at least for the time being. Tables are often embedded in dashboard items/tracker widgets. These widgets are user re-sizable.

By being enforceable the layout can be determined programmatically without having to depend on the viewport (which wouldn't work here for obvious reasons)

What we can't do is breakpoints based on container width, that's a technical limitation

Mohammer5:
> In responsive mode, I get a label for the footer above the button:

Fixed
I've also changed the default layout of the table to static.

varl:
> Indeed, I've also experienced the same. I try to look at breakpoints as 'user driven'. For example, What is a user doing, where are they, what are their limitations when assigned to the `@<768` breakpoint? To set up a strategy that will survive the constant development of DHIS2 I think we need some input from management on the plans for mobile expansion (apps, PWA, none?).

There are a lot of benefits to going with PWA, and we have had a prototype built converting Dashboards App into a PWA, with promising results. That seems like the best bet for now. I can run this up the chain and see if there is backing for it.

> 
> That said, I think most of the molecules/organisms in `ui-core` would benefit from fluid/adaptive resizing with many elements set to percentage widths of the parent. I think pixel-based breakpoints make sense in `ui-layout` (is that a thing?). Here we would define pixel-based breakpoints based on the user behaviour as I mentioned above. Within those breakpoints I imagine `ui-core` components fluidly adapting. But I am open to imagining how the workflow could be defined differently.

Media queries are on the device level, so figuring out the size of the device viewport is fine, but figuring out the parent element is not.

ismay:
To me it seems like a basic requirement to be able to set the breakpoint differently, based on what you're rendering in it.

Mr Brad Frost mentions it as well: http://bradfrost.com/blog/post/7-habits-of-highly-effective-media-queries/#content: `let content determine breakpoints`. We don't know how many columns will be rendered in the table, so we'd need different breakpoints depending on where the layout will break down right? I don't see why that's terrible.

cooper-joe:
> > That said, I think most of the molecules/organisms in `ui-core` would benefit from fluid/adaptive resizing with many elements set to percentage widths of the parent. I think pixel-based breakpoints make sense in `ui-layout` (is that a thing?). Here we would define pixel-based breakpoints based on the user behaviour as I mentioned above. Within those breakpoints I imagine `ui-core` components fluidly adapting. But I am open to imagining how the workflow could be defined differently.
> 
> Media queries are on the device level, so figuring out the size of the device viewport is fine, but figuring out the parent element is not.

I think media queries/breakpoints work for the layout components, but the components inside of them should behave fluidly (e.g. no breakpoints, all widths fixed or percentage based). This, of course, won't work for all components, but from what I've seen so far in DHIS2 I think this makes the most sense.

varl:
Given both what @ismay mentioned and the thoughts @cooper-joe has on how components and layouts interact, a way forward would be to yield the choice to the App and omit the media queries from **ui-core** completely.

Mohammer5:
I'll open a notes issue, laying out what I had in mind of how we could go forward (especially with the layout components & why I think they're so important).

ismay:
> Given both what @ismay mentioned and the thoughts @cooper-joe has on how components and layouts interact, a way forward would be to yield the choice to the App and omit the media queries from ui-core completely.

Yeah that, and @Mohammer5's suggestion of the flag to trigger a different layout sound preferable to me. Though I fully admit that that's not ideal, as handling it in css is preferable of course. But as a compromise I think it gives us the most usable component, which can be used by all. But yeah, let's talk about it in the notes issue, sounds good!

ismay:
> Some options:

> Introduce a dense prop which switches the Table component into the stacked view.

> Expose two Table components: Table and StackedTable, the name is debatable, just avoid "mobile".

Yeah both of those sound good to me. I was talking about this with @Mohammer5, and realistically we don't expect the screensize to vary a lot during use (unless it's a frontender 😋), so we don't have to worry as much about the rerendering cost of switching from one component to the other. Letting an app decide which component to render based on some logic it defines itself (like screen width, content, etc.) should be an ok compromise I think.

varl:
@Mohammer5 This has my blessing if we drop the media queries.

The compatible-with-[dhis2/notes#59](https://github.com/dhis2/notes/issues/59) way would be to only push the Table component with the scrollable wrapper.

You could pull out the code for the stacked view into a different branch where you add the separately add **StackedTable** component. We could refactor/merge that one based on the conclusions in dhis2/notes#59. I doubt the outcome of that issue would change the standard **Table** component.

src/Table/Table.js
```diff
@@ -0,0 +1,71 @@
+import { instanceOfComponent } from '@dhis2/prop-types'
+import React from 'react'
+import css from 'styled-jsx/css'
+import propTypes from 'prop-types'
+
+import { Consumer } from './tableContext'
```
varl: I would like us to move towards a more standards compliant way to import modules.

Only certain bundlers allow:

1. Omitting the suffix (`.js` in this case)
2. Automatic resolution of `index.js` files in directories

E.g. **Webpack** (ES6+NodeJS supported) supports it. **Rollup** (Only ES6+ supported) does not.

The reason for that is that this is **NodeJS Module Resolution** semantics. It is strictly speaking not valid for ECMAScript Modules.

I started digging in to this when I realised that Rollup did not accept the way our `import` statements are designed in **ui-core** since we do both **1.** and **2.** above. 🙅‍♂ 

The reason our ES6 bundle (`build/es`) works at all is because Babel saves our asses by fixing all our `import` statements to:

```js
import { Consumer } from './tableContext.js';
```

I don't think we should be mixing NodeJS semantics in pure ECMAScript libraries. It is confusing and increases our reliance on our tooling to save us from ourselves.

### Further reading

[1] https://nodejs.org/dist/latest-v10.x/docs/api/modules.html
[2] https://www.ecma-international.org/ecma-262/6.0/#sec-imports
[3] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Syntax

varl: We don't have to do this in this PR. We can do it for the entire library in one go and merge that in. We should keep it backwards compatible and at the same time think about re-organising the code from `src/${Component}/index.js` to e.g. `src/${Component}.js` + `src/${Component}/styles.js` (e.g.).

Mohammer5: > We should keep it backwards compatible and at the same time think about re-organising the code from src/${Component}/index.js to e.g. src/${Component}.js + src/${Component}/styles.js (e.g.).

Yeah, this is what I wanted to do!

> The reason for that is that this is NodeJS Module Resolution semantics. It is strictly speaking not valid for ECMAScript Modules.

So in the future we should append the `.js` suffix, did I get that right?

Mohammer5:
So for this PR I'll split the static table and the stacked table into two different components, right?
@varl @ismay

src/Table/Table.js
varl: > Yeah, this is what I wanted to do!

I remember. 👍

> So in the future we should append the .js suffix, did I get that right?

Yes.

Mohammer5:
@varl @ismay @cooper-joe 
I've split the component into two components, no more media queries

Mohammer5:
Hmm, something with the spacing went wrong, I'll fix that

Mohammer5:
done

varl:
Nice work on splitting the components. 👍 

I also meant pulling out the **StackedTable** to a different PR and only shipping the standard **Table** in this PR.

The reason is that if we ship both components in this PR we commit to a strategy, we effectively render the discussion dhis2/notes#59 obsolete. It can be closed with "_... but we have already done it this this way so..._".

In its current state we should hold off on this until we have discussed dhis2/notes#59.

I think that we need the **Table** anyway, and that base component doesn't impact the responsive strategy in dhis2/notes#59:

- If we decide to use a `prop` for switching the **Table** to a stacked mode, _that is not a breaking change_.

- If we decide to add a **StackedTable** component, _that is also not a breaking change_.

- If we decide to drop the **StackedTable** component though, _that **would be a breaking change**_ (this is why it forces a bias on the proposal in dhis2/notes#59).

One note on the name **StaticTable**:

The name for **StaticTable** should be **Table**, since by using the general name that signals that it:

- is the default,
- that it resembles the HTML table in scope and behaviour, 
- and avoids any confusion that "Static" adds as there is no linguistic relationship between "static" and "stacked" (the stacked table is also technically "static").

Mohammer5:
Closed, replaced by #377 & #378

PR #310, opened 2019-06-25
fix: <400 table styles 

cooper-joe:
fix: remove space, extra borders between table cells <400
fix: adjust spacing/padding tables cells <400


cooper-joe:
Sorry, I accidentally pushed straight to the branch. Reverted that and opened this PR.

PR #312, opened 2019-06-27
feat: add tabs component

HendrikThePendric:
This PR introduces the `Tabs` and `Tab` component. Some notes:
- I had a worked on these components before, but was then following the MUI specs, so I used that component as a starting point which I then refactored, and aligned with the specs.
- I managed to simplify the old code quite a bit, but I held on to two of the features that I think are quite nice, but not in the specs:
    - A sliding scroll indicator. Personally I like how it looks, but I am fine with removing it. If we did, we could probably get rid of a few lines in `Tabs` and the entire `TabIndicator` component.  And instead we could add some extra styles to the `Tab`*.
    - Two buttons that help with the scrolling if there are too many tabs in the container. I feel quite strongly about keeping these two. Not only does it improve the UX in a functional way, the presence of these buttons also makes it much more clear that more tabs are hiding away from view.
- I've added stories and docs already. Quite likely, the last story needs to be removed, before we merge this in, but it does demo the component in the best way


Mohammer5:
One thing that has nothing to do with code:
The commit messages should follow the following rules:
1. in the present
2. should start with a verb after the colon

i. e. "wip - applying theme styles" could be "apply theme styles partially"

src/Tabs/Tab.js
```diff
@@ -0,0 +1,109 @@
...
+
+const Tab = forwardRef(
+    ({ icon, onClick, selected, disabled, fixed, children }, ref) => (
+        <button
+            className={`${cx({
+                selected,
+                disabled,
+                fixed,
+            })}`}
+            onClick={disabled ? noop : onClick}
```
Mohammer5: I think it's better to change `onClick`'s default value to undefined (by removing it from the defaultProps), that way no function will be executed. This line would change to `onClick={disabled ? undefined : onClick}`

src/Tabs/Tab.js
```diff
@@ -0,0 +1,109 @@
...
+            ref={ref}
+        >
+            {icon && icon}
```
Mohammer5: Should be `{icon}`.
If the left-hand value is falsy, this expression will return the value of the left-hand variable, which is empty string.

src/Tabs/Tab.js
```diff
@@ -0,0 +1,109 @@
...
+    addTabRef: PropTypes.func,
+    selected: PropTypes.bool,
+    disabled: PropTypes.bool,
+    children: PropTypes.node,
+}
+
+Tab.defaultProps = {
+    active: false,
+    disabled: false,
+    onClick: noop,
```
Mohammer5: See comment above

src/Tabs/Tab.js
```diff
@@ -0,0 +1,109 @@
...
+    icon: PropTypes.string,
+    onClick: PropTypes.func,
+    addTabRef: PropTypes.func,
+    selected: PropTypes.bool,
+    disabled: PropTypes.bool,
+    children: PropTypes.node,
+}
+
+Tab.defaultProps = {
+    active: false,
```
Mohammer5: Not used, can be removed

src/Tabs/Tab.js
```diff
@@ -0,0 +1,109 @@
...
+    disabled: false,
```
Mohammer5: Default value of boolean props is falsy anyways, so this default value can be removed.

src/Tabs/Tab.js
```diff
@@ -0,0 +1,109 @@
...
+Tab.propTypes = {
+    icon: PropTypes.string,
+    onClick: PropTypes.func,
+    addTabRef: PropTypes.func,
+    selected: PropTypes.bool,
+    disabled: PropTypes.bool,
+    children: PropTypes.node,
+}
+
+Tab.defaultProps = {
```
Mohammer5: When accepting all suggestions below, this can be removed entirely

src/Tabs/TabBarScroller.js
```diff
@@ -0,0 +1,136 @@
...
+            onClick={scrollLeft}
+            className={cx('scroll-button', {
+                hidden: scrolledToStart,
+            })}
+        >
+            <ChevronLeft />
+        </button>
+        <div className="scroll-box-clipper">
+            <div className="scroll-box" ref={scrollBoxRef}>
+                <div className={'scroll-area'} ref={scrollAreaRef}>
```
Mohammer5: Can be `className="scroll-area"`

src/Tabs/TabIndicator.js
```diff
@@ -0,0 +1,90 @@
...
+    render() {
+        const { translateX, scaleX } = this.getTransformStyle()
+
+        return (
+            <span
+                className={cx({
+                    visible: this.props.visible,
+                    animated: this.state.animated,
+                })}
+                style={this.getTransformStyle()}
```
Mohammer5: Unnecessary due to the style tag below

varl:
Is it possible to hide the browsers scroll bar when this loads? https://deploy-preview-312--dhis2-ui.netlify.com/?path=/story/tabs--interactive-demo-with-scrolling

src/Tabs/Tabs.js
```diff
@@ -0,0 +1,243 @@
...
+import { animatedSideScroll } from './animatedSideScroll'
+import { Tab } from './Tab'
+import { TabBar } from './TabBar'
+import { TabIndicator } from './TabIndicator'
+
+class Tabs extends PureComponent {
+    container = createRef()
+    scrollBox = createRef()
+    scrollArea = createRef()
+    tabRefs = Children.map(this.props.children, createRef)
```
Mohammer5: This approach will fail if tabs are added dynamically

src/Tabs/Tabs.js
```diff
@@ -0,0 +1,243 @@
...
+
+Tabs.propTypes = {
+    className: propTypes.string,
+    selected: propTypes.number.isRequired,
+    fixed: TabBar.propTypes.fixed,
+    children: propTypes.arrayOf(instanceOfComponent(Tab)),
+}
+
+Tabs.defaultProps = {
+    fixed: false,
```
Mohammer5: When using `TabBar.propTypes.fixed` I guess it makes sense to also use `TabBar.defaultProps.fixed` here

HendrikThePendric:
Thanks for the swift review 👍 
I have fixed most things mentioned above. I'll just summarise what I have not addressed:

> I see you set a max-width, nice. Any chance we can truncate text on max-width and use a tooltip (either our ui-core or browser default is fine) to expose the rest of the title?

I've implemented the ellipsis (truncation), but adding the tooltip would be a bit complex. A developer could pass a component with a tooltip to the tab, i.e.
```
<Tab>
  <Tooltip content="long text">short text</Tooltip>
</Tab>
```
I think that should allow for whatever is needed...

> The Tabs component needs a onActiveChange (or so) callback so the app can change the content accordingly.

That could be a way to go. But I would like to keep this component dumb. If you want to use it in an app, you need a very slim wrapper around it, i.e. https://github.com/dhis2/ui-core/blob/304f74f81c533dfa6ea9b2eac9dbca4b3323b8aa/stories/Tabs.stories.js#L8-L35
@varl what do you think reg. the above?

> Generally I think this component is way to smart to be in ui-core and should be in ui-widgets

I agree it does quite a bit, but it isn't "smart" by our working definition:
1. It doesn't reach out to the server
2. It doesn't contain any domain or organisation specific business logic
So I think it's good to keep it in ui-core....

cooper-joe:
@HendrikThePendric I'm seeing some weird behaviour now on the scrollable tabs demo, especially when the last tab is selected. 
- The tabs aren't scrolling at all now, even though they should be.
- When I select the last tab I get a browser scrollbar:
Screenshot 20190628145555

HendrikThePendric:
@cooper-joe thanks for spotting this, I had indeed broken the scrolling and hiding the scrollbar because of some variable-renaming-gone-wrong... Fixed both issues in https://github.com/dhis2/ui-core/pull/312/commits/e2275aafa0da7d60a90a5acdfe30d3a0b3a51a05. Should be good to test now. Apologies for not spotting this myself.

cooper-joe:
@HendrikThePendric The other issues I raised are fixed, however I still see the scrollbar when selecting the last tab in a scrollable tab list. Firefox 67.0.4

image

Mohammer5:
> But I would like to keep this component dumb.

:+1: Agree, I thought you're keeping track of the active item inside the `Tabs` component, but you don't! Better this way

> I agree it does quite a bit, but it isn't "smart" by our working definition:
> 
> * It doesn't reach out to the server
> * It doesn't contain any domain or organisation specific business logic
>   So I think it's good to keep it in ui-core....

What I got from what @varl told me is that `ui-core` should contain our building blocks. The `Tabs` component is already very sophisticated and opinionated, it'd even expect to find it `ui-widgets`.

> @HendrikThePendric The other issues I raised are fixed, however I still see the scrollbar when selecting the last tab in a scrollable tab list. Firefox 67.0.4

I can't reproduce this in FF 67 on arch linux.

HendrikThePendric:
Reg. the issue that @cooper-joe spotted: this was only happening on retina screens in FF. I've fixed it in https://github.com/dhis2/ui-core/pull/312/commits/6e965999df19b4e0e04cf1e70a2dcfe119ea8dce.

> The Tabs component is already very sophisticated and opinionated, it'd even expect to find it ui-widgets.
OK, I don't have very strong opinions about this. If both you and @varl think it should be in `ui-widgets` then I'll move it. In any case, I think it would be good to finish the PR review here. Once everyone is happy with the component we can:
1. Close this PR
2. Open a PR to ui-widgets and link to this discussion
3. Merge it in

varl:
Roughly what I was thinking structure-wise:

```
src/ScrollBar/animatedSideScroll.js
src/ScrollBar/detectHorizontalScrollbarHeight.js
src/ScrollBar/index.js (renamed: TabBarScroller.js)

src/TabBar/index.js (renamed: InnerTabBar.js)

src/Tab/index.js
```

`TabBar` and `Tab` should be dead simple components, and all the complexity can be hoisted to `ScrollBar`. That way, if the `ScrollBar` isn't actively pulled in to the App, then it can be tree-shaken out, and in cases where only the `TabBar` and/or `Tab` is used, then we do not have to pay the price of binding the components to the DOM and event models.

HendrikThePendric:
@varl Thanks for the review. I am on board with all your suggestions and will implement them. I have a couple questions/remarks still:

> [...] a ScrollBar component which contains a child (basically overflow-x: scroll) and allows the user to scroll e.g by a set amount (50 px) on each arrow click.

Idea: We could base the the scroll distance on the width of the TabBar, i.e. scroll 50%/100% on every click

> ```src/ScrollBar/detectHorizontalScrollbarHeight.js (preferably deleted)```

I would like to get rid of this function too, but we could only do that if we allow scrolling using the buttons only. If we want to allow mouse scrolling too, we really need to clip the scrollBar and need this function to do so correctly...

Let me know what you think about the above two points, and I'll get to work

varl:
> Idea: We could base the the scroll distance on the width of the TabBar, i.e. scroll 50%/100% on every click

Nice, let's try that. 👌 

> I would like to get rid of this function too, but we could only do that if we allow scrolling using the buttons only. If we want to allow mouse scrolling too, we really need to clip the scrollBar and need this function to do so correctly...

Yeah, I get it. Leave it for now, the mouse scrolling is a really nice usability feature (and a11y).

HendrikThePendric:
@varl: I am working on the simpler implementation now, without having a ref on each tab. There is one downside to that:
- Suppose the active item is at the end of the tabbar when the scrollbar loads
- Now we have no way of determining if the tab is in view, so in this case the active tab will remain out of view...

HendrikThePendric:
@varl I've implemented your feedback and have ended up with a much simpler (set of) component(s). The only downside I see is what I mentioned before https://github.com/dhis2/ui-core/pull/312#issuecomment-509155820.

varl:
Looks good to me, if we run in to that situation we can figure out how to deal with that.

Mohammer5:
@HendrikThePendric @varl This could be achievable with a single ref on the `Tabs` component, couldn't it? `this.tabsRef.querySelector('.tab.active')`

src/Tabs/Tab.js
```diff
@@ -0,0 +1,121 @@
...
+                color: ${colors.grey600};
+                font-size: 14px;
+                line-height: 14px;
+
+                transition: all 150ms ease-in-out;
+                transition-property: color, background-color;
+                cursor: pointer;
+            }
+
+            :global(.fixed) > button {
```
Mohammer5: This definitely should be `.tabs.fixed`, otherwise `.fixed` class on a body/html element (to prevent scrolling) might cause this to behave differently

HendrikThePendric: I don't think your worries are valid, because this rule will end up like this:
`.fixed > button.jsx-123456` 
And please note the `>` [Child combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator) in the selector.
So:
- This style is only applied to buttons with this unique class name, so no risk of styles leaking out
- This style is only applied when these buttons are **immediate children** of an element with the `.fixed` class.

Actually I initially went with the `.tab-bar.fixed > button` selector, but then remembered your slack suggestion about potentially using the `Tab` element outside of the `TabBar`, so this slightly more generic, but still secure selector seemed more appropriate.

Mohammer5: Oops you're right, how could I miss that... my bad

HendrikThePendric: NP 👍

HendrikThePendric:
> Could we use the flex-wrap: nowrap; rule instead?

That worked a treat, applied in https://github.com/dhis2/ui-core/pull/312/commits/29996b267788ad7cd9e2d800799448780182288c

How about the suggestion @Mohammer5 made:
> This [finding the correct tab and scrolling to] could be achievable with a single ref on the Tabs component, couldn't it? this.tabsRef.querySelector('.tab.active')
This is very true. I could add a method in the ScrollBar that does just this... But this would obviously not work as soon as someone places another component into the scrollbar...

HendrikThePendric:
I've implemented the suggestion @Mohammer5 made in https://github.com/dhis2/ui-core/pull/312/commits/ed667be5786dc53b181d625445ca1efe30c350d8. Upon reflection, I think we should include this:
- It really benefits the UX and doesn't add too much complexity
- Fully leans on the assumption that there is a `.tab.active` element available, so that's not ideal. However, if the element is not there, this method simply does nothing.

PR #315, opened 2019-07-01
feat: add focus prop to button

teug91:
Adds the functionality to give focus to a button initially.


varl:
Thanks for adding this @teug91. :+1:

PR #317, opened 2019-07-02
Table header labels in context

Mohammer5:
With these changes, the title for all table cells has to be provided to the table head cells.
They're extracted by leveraging the `React.Children` api, coupled with stricter propTypes to ensure that only the correct components are used.

I've decided against using a config object because it was way harder to read. Now the table can be constructed like a normal table, with multiple rows in the `thead`.

The `TableCell` component has a `noTitle` prop if the content is special and doesn't need a title.

I've added examples in the storybook

UPDATE:
I could use the same approach for row-header labels (the vertical headers displayed in pivot tables for example)

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -1,7 +1,12 @@
 import React from 'react'
 import css from 'styled-jsx/css'
+import propTypes from 'prop-types'
 
 import { Consumer } from './tableContext'
+import { TableBody } from './TableBody'
+import { TableFoot } from './TableFoot'
+import { TableHead } from './TableHead'
+import { instanceOfComponent } from '../prop-validators/instanceOfComponent'
 
 const tableStyles = css`
     table {
@@ -48,3 +53,16 @@ export const Table = ({ children }) => (
         }}
     </Consumer>
 )
+
+const childPropType = propTypes.oneOfType([
+    instanceOfComponent(TableHead),
+    instanceOfComponent(TableBody),
+    instanceOfComponent(TableFoot),
+])
+
+Table.propTypes = {
+    children: propTypes.oneOfType([
+        childPropType,
+        propTypes.arrayOf(childPropType),
+    ]).isRequired,
+}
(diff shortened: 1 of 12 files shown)


amcgee:
@Mohammer5 nice, this looks much cleaner to me!  One question: is it necessary to specify `noTitle` when there's no title specified in the `head` for a given column?  For instance in the Story there's a `label=""` in the header for a column, and then each TableCell in that column needs to also specify `noTitle`?

If we keep the `noTitle` behavior, maybe it should be called `noLabel` for consistency?

Generally through really great, huge win to support responsive column/row labels without duplicating the label props!!

Mohammer5:
> is it necessary to specify noTitle when there's no title specified in the head for a given column? For instance in the Story there's a label="" in the header for a column, and then each TableCell in that column needs to also specify noTitle?

There is a title ("Actions"), it's just the second header row that doesn't have a title.
I'll remove the empty `label=""` from the header cell in the story thought

> If we keep the noTitle behavior, maybe it should be called noLabel for consistency?

I've moved the noTitle to the head cells which allows for more control. I've updated the second story, now the "Name" head cell in the first row has "noTitle" as well, so it's not shown in the responsive view.

I had to replace the "middle" responsive view (>400 - 768px) with the phone view (which is now from 0 - 768) as the html I came up with was invalid (td inside td as I needed the colSpan for my solution, due to this limitation it's not possible to do this with pure css as it doesn't support colspan...). I've talked to @cooper-joe though and he said that the phone view should still be fine.

amcgee:
I've only looked at the API, not the implementation, but this seems really nice and clean to me.  Nice work @Mohammer5 !

One thought - the `noTitle` property on a `<TableHead>` seems a little misleading, since there will presumably be a title in Desktop orientations (omitted in mobile contexts).  Maybe something like `noRowLabel` or `noResponsiveLabel` might be a little clearer?

Mohammer5:
@amcgee I've renamed it to `hideResponsiveLabel`.
Will merge into the `Table` branch once you've approved this PR (giving you the chance to look at the code if you want to. If you don't want to, there's still a chance to do it in the `Table` branch PR)

PR #318, opened 2019-07-04
feat(layout): add layout components & styles

Mohammer5:


HendrikThePendric:
Great initiative :-). I haven't really looked at the code yet, but just focussed on the component API and style. Some initial thoughts:

1) API-wise it would be nice if we could avoid having to add these props to the `LayoutContainer`. I haven’t really worked with CSS grids much yet, but I reckon with flexbox this could/would/should be possible (not sure)…

2) Maybe the component names could be a bit shorter, i.e. `LayoutAreaSidebar` => `Sidebar`? (what’s inside could be called `SideNav` or something)

3) Not quite sure about this, but perhaps we could make the names correspond more with the components location/style, as opposed to what its function is likely to be… I.e. instead of using `Navigation` we could use `TopBar` (i.e. analogue to SideBar). Actually, this is not a general comment, it just relates to Navigation...

4) Style wise, I think the container should have a min-height of 100%

Mohammer5:
> API-wise it would be nice if we could avoid having to add these props to the LayoutContainer. I haven’t really worked with CSS grids much yet, but I reckon with flexbox this could/would/should be possible (not sure)…

Flexbox doesn't play nicely with grids, it's a one directional flow of elements, so it wouldn't allow this:
```
|-------------------|
|     Navigaton     |
|-------------------|
| Sidebar | Content |
|-------------------|
```

We could expose different containers, but it'll be hard to name them properly (`ContainerSidebarTopbar`?)

> Maybe the component names could be a bit shorter, i.e. LayoutAreaSidebar => Sidebar? (what’s inside could be called SideNav or something)

Absolutely agree with that, the names are terribly long right now, although I'd like to keep a reference to `Layout` somehow to make it clear when looking at implementations that it's a layout component and has nothing to do with the actual content (`Sidebar` would imply to me that there's sidebar stuff going on? Dunno, haven't really figured the naming part out yet)

> Not quite sure about this, but perhaps we could make the names correspond more with the components location/style, as opposed to what its function is likely to be… I.e. instead of using Navigation we could use TopBar (i.e. analogue to SideBar). Actually, this is not a general comment, it just relates to Navigation...

`TopBar` sounds awesome

> Style wise, I think the container should have a min-height of 100%

This is not really possible atm. By simply adding `height: 100%;`, the layout breaks. By applying `height: content-fit` to the `TopBar`, the layout breaks.. This can be solved by the apps when needed (applying a specific height to the `TopBar` and `height: calc(100vh - ${specificHeight})` to the `Sidebar` and `Content`

Mohammer5:
> API-wise it would be nice if we could avoid having to add these props to the LayoutContainer

I thought about this a bit and I really like the `with...` props. They allow you to specify which areas you want without having to find the appropriate container component, it's very maintainable.
The container component is also extensible and we don't have to create long names for the various versions to describe their proper use case

> Maybe the component names could be a bit shorter, i.e. LayoutAreaSidebar => Sidebar? (what’s inside could be called SideNav or something)

I've renamed the follwing component:
* `LayoutAreaNavigation` to `LayoutTopBar`. 
* `LayoutAreaSidebar` to `LayoutSidebar`
* `LayoutAreaContent` to `LayoutContent`

I think it's important to keep `Layout` in the name to indicate their usage.
`SideNav` has the same problem as `Navigation`, so I'd rather keep it as `Sidebar`

HendrikThePendric:
> I thought about this a bit and I really like the with... props. They allow you to specify which areas you want without having to find the appropriate container component, it's very maintainable.

My main problem with these `withSomething` props, is that you are declaring what should go into the container both via the container props, and via the children components. It is essentially duplicate information. And it allows the developer to create invalid combinations very easily...

You are right in both things you are saying about flexbox:
1) Flexbox isn't specifically designed for grids and css grids are (no surprise there)
2) The only way to get a decent layout using flexbox would be to create an additional component, which would flip the flex direction from row to column...

This being said, it would enable us to get stuff to work without adding these properties to the container, i.e. we could do something like this:
```jsx
<LayoutFlexColumn>
    <LayoutTopBar />
    <LayoutFlexRow>
        <LayoutSideBar />
        <LayoutContent />
    </LayoutFlexRow>
</LayoutFlexColumn>
```

I would also prefer to get rid of all the `Layout` prefixes personally, but that's less of a biggie for me.

Mohammer5:
@HendrikThePendric You're right, it's redundant. I've found a simpler way: Checking what's inside `children`... No idea why I didn't do that before, I'm also doing that in the `Table` component proposal and it's worked fine there sofar

src/Layout/helper.js
```diff
@@ -0,0 +1,12 @@
+import { Children } from 'react'
+import { LayoutTopBar } from './LayoutTopBar'
+import { LayoutSidebar } from './LayoutSidebar'
+
+export const childrenContainsComponent = component => children =>
+    Children.toArray(children).reduce(
+        (found, child) => found || child.type === component,
+        false
+    )
```
HendrikThePendric: just returning this would a bit easier on the eye:
```javascript
!!Children.toArray(children).find(({ type }) => type === component)
```
(and less iterations, but that's not really relevant in this context...)

HendrikThePendric:
> And instead dynamically load the correct styles
Or just include all styles in a single stylesheet, and just toggle some classNames... Even simpler...

Mohammer5:
@HendrikThePendric I've replaced the container elements with styles, looks a lot cleaner now!

Regarding contextual semantics:
I like the idea to always use the correct semantic element here, I'd like to limit it though:
1. Content should always be `main`
1. Sidebar should always be either `section` (default) or `aside`
1. The TopBar should always be either `header`, `nav`, `section`, `aside` or `div` (default) as no other elements would make sense here.

Mohammer5:
@varl can be closed until we (maybe) decide to create a `ui-layouts` lib & we decide this one could be a part of it, right?

Issue #306, opened 2019-06-24
Rework text input (replace outlined style)


cooper-joe:
I made a mistake with the input fields. I was writing the documentation for filled/outlined styles and I realised I couldn’t really justify where one would be used over the other. I made the mistake of taking this from Material Design without doing due dilligence. 

To rectify this, I would like to replace the 'outlined' style with a regular, traditional, split label+input style. The 'filled' style can remain unchanged.

To be clear, the input types we would then have:
- filled
- traditional split (separate? Lacking a good name)

Apologies that this will undo some work already done to a high standard. I spoke to @varl and he thought we might be able to make these changes to the outlined style without making breaking changes. I think it may be possible, all the states remain the same functionally (error, disabled, dense etc.), it is visually that there are big changes. Ideally I would like to make this change soon, before apps start using the outlined style extensively.

Please grab me on slack to discuss if needed. Here are the specs: https://sketch.cloud/s/DwkDk/a/jO5k5b

CSS for the input box-shadow, for easy copy&paste:
```css
box-shadow: inset 0 0 0 1px rgba(102,113,123,0.25), inset 0 1px 2px 0 rgba(102,113,123,0.21);
```

Mohammer5:
> I spoke to @varl and he thought we might be able to make these changes to the outlined style without making breaking changes

I'm 100% sure that we'll be able to change it without changing the component's api.

> Here are the specs: https://sketch.cloud/s/DwkDk/a/jO5k5b

I like this one a lot more tbh! Looks neat

cooper-joe:
@Mohammer5 Sounds good! As I wrote above, the different states shouldn't require major changes; error still has the icon, help text still appears below, etc. The disabled state has a little bit different style, but that's it I think. The major change is the label position, and the separation of the label/placeholder (although I believe this is done in the current one already, as label and placeholder can be different?)

Let me know if you have any questions underway 👍

cooper-joe:
A note to say that this rework would also include the 'outlined' select style too.

Mohammer5:
@cooper-joe The help & error texts have a different font size than the label, placeholder and value texts.
What values should I use? In my branch I'm using 16px for label, placeholder and value right now.

cooper-joe:
> @cooper-joe The help & error texts have a different font size than the label, placeholder and value texts.
> What values should I use? In my branch I'm using 16px for label, placeholder and value right now.

[...]

PR #337, opened 2019-07-12
refactor: adjust styles of outlined input components

Mohammer5:


varl:
If the Label is too long, couldn't we just wrap it now and show it on multiple lines?

Should the dense Label be the same size as the non-dense one?

src/FieldLabel/Outlined/index.js
```diff
@@ -20,58 +19,42 @@ const styles = css`
...
-    .dense {
-        height: ${inputHeightDense + 10}px;
+    .content {
+        align-items: center;
+        box-sizing: border-box;
+        display: flex;
+        height: 42px;
+        left: 1px;
+        position: relative;
+        width: calc(100% - 1px);
```
varl: I'm not ecstatic about these hacks, can we put a comment about what the situation is that requires it so we know when they can be removed? :thinking:

stories/InputField.stories.js
```diff
@@ -57,7 +60,20 @@ function createStory(name, props) {
...
+                    value="This value produces an error"
+                    error
+                />
+                <Help error {...props}>
+                    This describes the error, if a message is supplied.
+                </Help>
+                <Help {...props}>
+                    This is some help text to advice what this input actually
+                    is.
+                </Help>
```
varl: Generally I would put the error text after the default help text to avoid more elements than necessary moving around.

varl:
Let me know when this isn't a WIP anymore.

Mohammer5:
We'll have to wait for @cooper-joe 
See [#306](https://github.com/dhis2/ui-core/issues/306#issuecomment-510884218)

Mohammer5:
> Should the dense Label be the same size as the non-dense one?

@varl Fixed

> If the Label is too long, couldn't we just wrap it now and show it on multiple lines?

@cooper-joe ?

cooper-joe:
> If the Label is too long, couldn't we just wrap it now and show it on multiple lines?

Yeah, can do. Just to note, it should be possible for the label to be longer than the input. So the label can wrap if the parent has a width, but it shouldn't necessarily wrap to the input width.

Screenshot; this should be possible:
image

varl:
The label doesn't wrap so when it overflows it is truncated, see [comment](https://github.com/dhis2/ui-core/pull/337#issuecomment-518199278) from @cooper-joe.

Expected: 
image

While this causes the label to be readable at all times, inputs are no longer guaranteed to be the same height.

Cropping however is dangerous from a i18n perspective.

Mohammer5:
Right now, the `text-overflow` is still set to `ellipsis, if I wanted to change it, I'd change the label's style to 

```css
label {
  padding: 0 0 4px;
  overflow: visible;
  text-overflow: clip;
  white-space: normal;

  /*
    I'd just remove the height,
    but this is needed in the storybook to override the existing height
  */
  height: auto;
}
```

So, you're stating:

> While this causes the label to be readable at all times, inputs are no longer guaranteed to be the same height.

I assume you're talking about inputs that are next to each other?
Then you're right, but there are scenarios where it'd make sense to wrap the label,
so we could add a `wrapLabel` prop?

> Cropping however is dangerous from a i18n perspective.

That's also true, so I'm not sure what to do right now (do you want me to crop it, not crop it or should we use the `wrapLabel` prop?).

I guess @cooper-joe wants to have a say in this as well.

<hr />

**Here's what I don't understand:**
The first screenshot is labeled as "expected", this means the second one is wrong?

How did you produce (it if not intentionally, by setting a specific width on the internal `input` component)? And if you set a specific width on the internal input component, it'd match what @cooper-joe wanted in his [comment](https://github.com/dhis2/ui-core/pull/337#issuecomment-518199278).

If you want the label and input to match in width, just set an explicit width on their container or wrap them with a container with a specific width.

varl:
> ****
> The first screenshot is labeled as "expected", this means the second one is wrong?

It was meant as "also expected".

> 
> How did you produce (it if not intentionally, by setting a specific width on the internal `input` component)? And if you set a specific width on the internal input component, it'd match what @cooper-joe wanted in his [comment](https://github.com/dhis2/ui-core/pull/337#issuecomment-518199278).

I set an explicit width on the internal component which arguably wouldn't _normally_ happen since we don't pass a `className` to the internal `Input` component so it's more of a "_if_ the input has a width, then it behaves as specced" situation.

_Although_ an app could override width on `input` elements through global CSS.

> If you want the label and input to match in width, just set an explicit width on their container or wrap them with a container with a specific width.

That's possible.

This is the change I'm talking about:

```diff
diff --git a/src/FieldLabel/Outlined/Label.js b/src/FieldLabel/Outlined/Label.js
index 3f63f07..37176fe 100644
--- a/src/FieldLabel/Outlined/Label.js
+++ b/src/FieldLabel/Outlined/Label.js
@@ -15,13 +15,9 @@ const styles = css`
     label {
         display: block;
         box-sizing: border-box;
-        height: 24px;
         font-size: 14px;
         line-height: 20px;
         padding: 0 ${spacers.dp16} ${spacers.dp4};
-        overflow: hidden;
-        text-overflow: ellipsis;
-        white-space: nowrap;
     }
```

Mohammer5:
> In short:
> 
> * do not truncate labels for outlined fields, always wrap.
> * Truncate labels for filled fields.

This will mean:

> While this causes the label to be readable at all times, inputs are no longer guaranteed to be the same height.

Are you against the `wrapLabel` prop idea?

varl:
> 
> 
> > In short:
> > 
> > * do not truncate labels for outlined fields, always wrap.
> > * Truncate labels for filled fields.
> 
> This will mean:
> 
> > While this causes the label to be readable at all times, inputs are no longer guaranteed to be the same height.
> 
> Are you against the `wrapLabel` prop idea?

Yes, I'm against it. I don't see a practical use case where you ever want to truncate the label for a field.  

There is no way for the user to read the full label if we do, and that leads to the user not understanding what the field is for and she might input bad data.

The only reason for truncating the label with ellipsis would be for aesthetics, that a form looks nicer when all the fields are the same height. That would be **form over function** though, so I don't think it's a reasonable argument.

The **filled** label needs to truncate though, since that floating label is already a hack and wrapping that label would further interfere with the usability of the field...

@cooper-joe would have to agree to "always wrap" the label if needed though.

---

👨‍🚒: _a label should probably not be so long as to wrap on a desktop monitor in the first place..._

Mohammer5:
> The only reason for truncating the label with ellipsis would be for aesthetics

Agreed. I'm on board! @cooper-joe confirmed?

cooper-joe:
> > The only reason for truncating the label with ellipsis would be for aesthetics
> 
> Agreed. I'm on board! @cooper-joe confirmed?

@Mohammer5 Agreed and confirmed that labels should never be truncated.

I am not sure if my original question/comment has been confirmed though:

- is it possible for a label to be wider than the input field? I'm not sure how, but it should be possible to say 'The component itself width is 100%(or 500px), but the input itself is 100px.'. This is important, setting width of an input is a useful way to encourage users to input the right content (a small input implies a small number should be used!), so this is a necessary feature.

image

Mohammer5:
@cooper-joe yes that's possible. The implementing app/lib can override the width of the input field with another value (relative or absolute)

cooper-joe:
> @cooper-joe yes that's possible. The implementing app/lib can override the width of the input field with another value (relative or absolute)

Great, thanks! In that case, LGTM

PR #345, opened 2019-07-18
chore(prop-types): switch to prop types from npm

HendrikThePendric:
- I identified a couple of mutually exclusive boolean props across the codebase:
    - status: valid | warning | error
    - buttonVariant: primary | secondary | destructive
    - size: large | small
- So I decided to put them in a central location and import them where needed
- I also manually scanned all the components and updated the prop-types where required


src/AlertBar/Icon.js
```diff
@@ -1,6 +1,8 @@
-import React from 'react'
 import propTypes from 'prop-types'
-import { Valid, Warning, Error, Info } from '../icons/Status'
+import React from 'react'
+
+import { statusPropType } from '../common-prop-types'
```
varl: General comment: Other imports use the `.js` suffix.

Mohammer5: Seems to be this file? I think there shouldn't be any `.js`

varl:
> It's a lot of repetition and all components will change if we ever add/remove/rename one of those prop types. But the it'll be disconnected from creating the class name with these props (we could extract that part and unify it everywhere as well?)

Not all repetition is automatically bad. Good repetition is when different things can be changed without affecting each other. For our **core** components that is a trait we should value.

src/AlertBar/Icon.js
varl: I see ... `common-prop-types` is a directory.

Mohammer5:
> Not all repetition is automatically bad. Good repetition is when different things can be changed without affecting each other. For our core components that is a trait we should value.

I absolutely agree. In this case:
* If this needs to be different on one place => can be adjusted where it needs to be different
* If this needs to be different in all places => all places have to be adjusted anyways
* If everything stays as it is => improves maintainability

HendrikThePendric:
I'm going to be off for the next three weeks. Feel free to move this forward in the meanwhile.

PR #359, opened 2019-07-29
fix(radio): set box-sizing to border-box

Mohammer5:


src/Radio/styles.js
```diff
@@ -16,6 +16,7 @@ export default css`
         height: calc(100% + 2px);
         top: -1px;
         left: -1px;
+        box-sizing: border-box;
```
varl: Is this necessary when we load in the [`CSSReset`](https://github.com/dhis2/ui-core/blob/master/src/CssReset/index.js#L15)?

Mohammer5: The import export app doesn't use the `CSSReset` and either we fix the component there or we make sure that the component can be used with and without the reset css

ismay: Would adding the CssReset there break things? Might also be a solution?

Mohammer5: That seems to work fine actually

Mohammer5:
`CssReset` should be used if possible

PR #377, opened 2019-08-16
feat: add stacked table component

Mohammer5:
This is the stacked component developed in #304 
Most of the commits are the same as in #378 , but as this will be "Squashed & merged", there shouldn't be any conflicts.

**Please do not approve until the commits have been squashed/grouped properly!** Just state that you're OK with the current state and wait for the commits to be adjusted if you'd normally accept this PR.


varl:
This depends on: https://github.com/dhis2/notes/issues/59

Mohammer5:
Waiting for response to [issue comment on #378](https://github.com/dhis2/ui-core/pull/378#issuecomment-522454269)

Mohammer5:
Keep open

Mohammer5:
I think we could add this component? We've agreed to leave out responsive styles from `ui-core` and with [this](https://github.com/dhis2/notes/issues/59#issuecomment-525167553) solutions, apps could reliably switch components.

I'd just have to remove the `TableCellText` component like I did in the `Table` component.

varl:
@Mohammer5 Took a look at this again now that 4.0 is out and yeah I agree, we can get this in if we align it with the Table implementation.

Mohammer5:
Ok, I think this one is ready for review.
@cooper-joe please have a look at the styles!

src/StackedTableBody.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import css from 'styled-jsx/css'
+import propTypes from 'prop-types'
+import { instanceOfComponent } from '@dhis2/prop-types'
+
+import { StackedTableRow } from './StackedTableRow.js'
+
+const tableBodyStylesResponsive = css`
+    tbody {
+        display: block;
```
cooper-joe: I notice in the story that the text content inside the table is rendering as browser default. It should be `grey900`. Is this just the case because it is an isolated component? I'm thinking that `grey900` should be the `table` color defnition. (Perhaps that is done in the non-stacked table and then inherited here?)

src/StackedTableCell.js
```diff
@@ -0,0 +1,100 @@
...
+
+        <style jsx>{`
+            .title {
+                display: block;
+                white-space: normal;
+                min-height: 24px;
+                line-height: 18px;
+                padding: 8px 0 0 0;
+                font-weight: bold;
+                white-space: nowrap;
```
cooper-joe: Right now the cell title appears more important than the content itself, I think this is the wrong way around. The content is important and the cell title is just a reference. I suggest adjusting the adding/adjusting the following styles for the title :
```
font-size: 13px;
line-height: 16px;
font-weight: 400;
padding: 8px 0 4px 8px;
color: ${colors.grey700};
```

src/StackedTableRow.js
```diff
@@ -0,0 +1,68 @@
...
+
+const tableRowStyles = css`
+    tr {
+        min-height: 45px;
+        display: block;
+        border: 1px solid #e8edf2;
+    }
+
+    tr:nth-child(even) {
+        background: none;
```
cooper-joe: change to `background: ${colors.white}`

src/StackedTableRow.js
```diff
@@ -0,0 +1,68 @@
...
+    :global(tbody) tr {
+        min-height: 36px;
+    }
+
+    tr + tr {
+        margin-top: 32px;
+    }
+
+    tr:nth-child(even) :global(td) {
+        background: #fbfcfd;
```
cooper-joe: I think we should turn off the zebra-striping for these stacked cells. They are separated by space and borders, so I think a change in background color would only be confusing.

src/StackedTableCell.js
Mohammer5: Sure can do, but the padding on the left looks very weird, breaks the vertical optical line.
Would `paddding: 8px 0 4px` be ok as well?

cooper-joe: Apologies, my mistake. `padding: 8px 0 4px` is correct, yes.

Mohammer5:
@cooper-joe I've addressed all the concerns
@dhis2/team-apps Please have a look at this as well

I still have to take care of the commit history, so please don't approve until it's appropriate & request a changes, but leave a comment if everything else is fine

cooper-joe:
Thanks for making those changes, can confirm they are all fixed. There is one more I noticed now:

- The bottom of the table has a double thickness border, 1px from the last cell and 1px from the table itself. Can this be fixed with some css selector wizardry?

varl:
> Thanks for making those changes, can confirm they are all fixed. There is one more I noticed now:
> 
>     * The bottom of the table has a double thickness border, 1px from the last cell and 1px from the table itself. Can this be fixed with some css selector wizardry?

cc: @Mohammer5

Mohammer5:
@varl thanks for pinging me, this one slipped me..
@cooper-joe I've added the footer to the default story. I think what you saw was the border of the storybook container and the border of the footer. I can't see it in the default story. Please have a second look

Mohammer5:
Commit have been squashed. This one is ready to be merged once approved

cooper-joe:
@Mohammer5 I'm still seeing the double border at the bottom of the table. The double border also appears on the footer area. The 2px border should only be 1px. Here is a zoomed in version:
image

Mohammer5:
@cooper-joe Ahh, you mean at the bottom of the rows (which appear like single tables now), sorry for that confusion. Fixed!

src/StackedTable.js
```diff
@@ -0,0 +1,68 @@
...
+    children: childrenPropType.isRequired,
+    className: propTypes.string,
+    dataTest: propTypes.string,
+}
+
+export const StackedTable = ({ children, className, dataTest }) => {
+    const headerLabels = extractHeaderLabels(children)
+
+    return (
+        <Provider value={{ headerLabels }}>
```
varl: We should pass the `headerLabels` as props to the children as well so that it is possible to use custom child components without relying on the provider (which is not exposed publicly).

cooper-joe:
Thanks, LGTM!

src/StackedTable.js
```diff
@@ -0,0 +1,68 @@
...
+import { StackedTableHead } from './StackedTableHead.js'
+import { colors } from './theme.js'
+
+const childPropType = propTypes.oneOfType([
+    instanceOfComponent(StackedTableHead),
+    instanceOfComponent(StackedTableBody),
+    instanceOfComponent(StackedTableFoot),
+])
+
+const childrenPropType = propTypes.oneOfType([
```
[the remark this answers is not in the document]
Mohammer5: Replaced by `propTypes.node`

src/StackedTable.js
```diff
@@ -0,0 +1,68 @@
...
+    childPropType,
+    propTypes.arrayOf(childPropType),
+])
+
+export const Table = ({ children, className, dataTest }) => (
```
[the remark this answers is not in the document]
Mohammer5: I've moved the `Table` to `src/StackedTable/Table.js`

src/StackedTable.js
Mohammer5: I've moved the context up to the row, which now supplies the cells with the `headerLabels`, which allows custom table cells (I've added a story).

Mohammer5:
@varl thanks for having a look at this again! I've addressed all issues, please have a look again

src/StackedTable/extractHeaderLabels.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react'
+import { StackedTableHead } from '../StackedTableHead.js'
+
+const isChildTableHead = child => child.type === StackedTableHead
```
varl: Will this work if I use a wrapped `StackedTableHead` component, or a custom component backed by a `thead`?

There's nothing special about the `StackedTableHead` so it should still work.

Mohammer5: I don't think this will be a problem, as these components will not be rendered, just used to extract the texts. The reason why these are components and not a prop is so you can configure colspans the "natural" way.

The `StackedTable` now accepts a `headerLabels` prop, so you could also just provide an array of strings instead, then this won't be used anymore, but it will only work for simpler cases. If we ever need to introduce rowspan in the header cells, it'd result in a very complex value we'd have to pass to the `StackedTable`.

If the app needs the header to be dynamic, you can still create the ui-core components through functions instead of components. It's not an ideal way, but other than that, I don't know how we should know which component is the actual header component.

varl: Ok, cool.

varl:
I have a question about usage.

Say I have a Table:

```jsx
<Table>
  <TableHead>
    <TableRowHead>
      <TableCellHead>
        First name
      </TableCellHead>
      <TableCellHead>
        Last name
      </TableCellHead>
      <TableCellHead>
        Incident date
      </TableCellHead>
      <TableCellHead>
        Last updated
      </TableCellHead>
      <TableCellHead>
        Age
      </TableCellHead>
      <TableCellHead>
        Registering unit
      </TableCellHead>
      <TableCellHead>
        Assigned user
      </TableCellHead>
      <TableCellHead>
        Status
      </TableCellHead>
    </TableRowHead>
  </TableHead>
  <TableBody>
    <TableRow>
      <TableCell>
        Obi
      </TableCell>
      <TableCell>
        Okafor
      </TableCell>
      <TableCell>
        06/07/1990
      </TableCell>
      <TableCell>
        01/03/2006
      </TableCell>
      <TableCell>
        28
      </TableCell>
      <TableCell>
        Dalakuru CHP
      </TableCell>
      <TableCell>
        Anatoliy Shcherbatykh
      </TableCell>
      <TableCell>
        Incomplete
      </TableCell>
    </TableRow>
  </TableBody>
  <TableFoot>
    <TableRow>
      <TableCell
        colSpan="8"
      >
        <TableFooterButton />
      </TableCell>
    </TableRow>
  </TableFoot>
</Table>
```

Now in my app, I want to dynamically switch this to a StackedTable.

I would expect that I would only have to switch out the top-level component, `Table` to `StackedTable` and it would just work, e.g.:

```
<StackedTable>
  <TableHead>
...
  </TableHead>
  <TableBody>
...
  </TableBody>
  <TableFoot>
...
  </TableFoot>
</StackedTable>
```

But what I actually have to do is switch out each component to a Stacked counterpart:

```
<StackedTable>
  <StackedTableHead>
    <StackedTableRowHead>
      <StackedTableCellHead>
        First name
      </StackedTableCellHead>
      <StackedTableCellHead>
        Last name
      </StackedTableCellHead>
      <StackedTableCellHead>
        Incident date
      </StackedTableCellHead>
      <StackedTableCellHead>
        Last updated
      </StackedTableCellHead>
      <StackedTableCellHead>
        Age
      </StackedTableCellHead>
      <StackedTableCellHead>
        Registering unit
      </StackedTableCellHead>
      <StackedTableCellHead>
        Assigned user
      </StackedTableCellHead>
      <StackedTableCellHead>
        Status
      </StackedTableCellHead>
    </StackedTableRowHead>
  </StackedTableHead>
  <StackedTableBody>
    <StackedTableRow>
      <StackedTableCell
        headerLabels={[]}
      >
        Onyekachukwu
      </StackedTableCell>
      <StackedTableCell
        headerLabels={[]}
      >
        Kariuki
      </StackedTableCell>
      <StackedTableCell
        headerLabels={[]}
      >
        02/06/2007
      </StackedTableCell>
      <StackedTableCell
        headerLabels={[]}
      >
        05/25/1972
      </StackedTableCell>
      <StackedTableCell
        headerLabels={[]}
      >
        66
      </StackedTableCell>
      <StackedTableCell
        headerLabels={[]}
      >
        Jawi
      </StackedTableCell>
      <StackedTableCell
        headerLabels={[]}
      >
        Sofie Hubert
      </StackedTableCell>
      <StackedTableCell
        headerLabels={[]}
      >
        Incomplete
      </StackedTableCell>
    </StackedTableRow>
  </StackedTableBody>
  <StackedTableFoot>
    <StackedTableRow>
      <StackedTableCell
        colSpan="8"
        headerLabels={[]}
        hideTitle
      >
        <Button
          primary
          type="button"
        >
          StackedTable footer button
        </Button>
      </StackedTableCell>
    </StackedTableRow>
  </StackedTableFoot>
</StackedTable>
```

Which I find very cumbersome in an application. Any way to reuse the existing Table components? I also think that wrapping the entire Table in a StackedTable component would be OK:

```
<StackedTable>
  <Table>
    <TableHead>
...
    </TableHead>
    <TableBody>
...
    </TableBody>
    <TableFoot>
...
    </TableFoot>
  </Table>
</StackedTable>
```

What do you think?

PR #378, opened 2019-08-16
feat: add static table component

Mohammer5:
This is the static component developed in #304
Most of the commits are the same as in #377, but as this will be "Squashed & merged", there shouldn't be any conflicts.


src/Table/TableCell.js
```diff
@@ -0,0 +1,37 @@
...
+
+    :global(tfooter) div {
+        min-height: 36px;
+    }
+`
+
+export const TableCell = ({ children, colSpan, rowSpan }) => (
+    <td colSpan={colSpan} rowSpan={rowSpan}>
+        <div>
+            <TableCellText>{children}</TableCellText>
```
varl: Why are we restricting ourselves to phrasing content here?

A `td` can contain flow content (e.g. `div`, `h1`, `p`) so the `children` should be valid for any [flow content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Flow_content).

As we force all the `children` into a `TableCellText`, which in turn puts the `children` inside of a `span`, we are restricting the `children` of `TableCell` to only be [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content). This is unnecessarily restrictive.

See permitted content table:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td

I think we should drop the `TableCellText` entirely.

src/Table/TableCellText.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
```
varl: See comment in `TableCell`, this component could be removed.

Mohammer5:
@varl I've removed the `TableCellText` component completely.
There's one thing that I'd like to talk about:

https://github.com/dhis2/ui-core/pull/378/files#diff-50b03b6ca5150712ec3246ba4fed07cfR18

In order to work without any overhead, the padding on the table cell needs to be `13px 0` for inline text in the cell and `9px 0` for other elements (like the button in the footer).

That was the reason why I created the `TableCellText`, so there's a wrapper provided.
@amcgee was [asking if that component is necessary](https://github.com/dhis2/ui-core/pull/304#discussion_r297697584) in the first place, because so far it has worked alright in the stories.

But I haven't thought about a case before where it might be necessary to display a button in a row, which will need `9px` vertical padding while all other cells in that row need `13px`.

These are the options that I see right now:
1. We introduce a wrapper for text which has additional padding
2. We always use `9px` and leave the rest to the user of the lib
3. We add a prop that can be used to increase the padding
4. We always use `13px` as text is the most likely content, leaving the user of the lib with the task to use negative margin on contents that need `9px`

I think 2. and 4. are not viable options. Using a component is for just one style property is a bit much and I have no idea how the prop (proposed in 3.) could be named (`shrink`, `grow`, `containsTextOnly`).

We can't depend on checking whether the content is `string` or react components as text could be wrapped with other inline elements to change the inline styles.

What do you guys think about this?

varl:
> 
> 
> @varl I've removed the `TableCellText` component completely.

Ok, cool.

> What do you guys think about this?

An alternative is that the `TableCell` and `TableFoot` components accept a `className` prop, which would allow the user to override the `padding` on the `td` and `tfoot` directly.

We would provide a sane default for `td` (13px) and `tfoot` (9px) in the respective components, so we do not have to use the `:global` modifier:

```
    :global(tbody) div {
        padding: 13px 0;
    }
    :global(tfooter) div {
        min-height: 36px;
        padding: 9px 0;
    }
```

ismay:
> An alternative is that the TableCell and TableFoot components accept a className prop, which would allow the user to override the padding on the td and tfoot directly.

I was talking to @Mohammer5 about this. Maybe something like a `dense` prop would make sense? For making the row a bit more compact? Conveys the intention well to the user for me.

ismay:
PR for the above comment here: https://github.com/dhis2/ui-core/pull/404

ismay:
@varl Let me know what you think of the changes.

varl:
@ismay why does the non-dense variant look more dense than the dense variant?

- Non-dense: https://deploy-preview-378--dhis2-ui-core.netlify.com/?path=/story/table--static-layout
- Dense: https://deploy-preview-378--dhis2-ui-core.netlify.com/?path=/story/table--static-layout-with-buttons-in-dense-cells

Another note, there's a massive amount of padding in the table footer, is that intentional?

ismay:
> @ismay why does the non-dense variant look more dense than the dense variant?

That's because of the buttons in the dense example. The dense cells have less padding, but they still grow to accommodate their content. And since the buttons are large the rows end up looking larger. We could omit the buttons from the example and just show default and default with dense. That'd illustrate the difference better.

(Reason for showing it with buttons was that dense seems like a prop you'd use to shrink rows that contain large content, like buttons, but it doesn't really help the example here)

ismay:
> Another note, there's a massive amount of padding in the table footer, is that intentional?

Yeah I don't think I touched that bit. Most of the whitespace there actually comes from what's rendered inside the footer (a div with padding and a button):

Screenshot 2019-09-19 at 10 05 19

button

Why the extra padding was added I don't know. We could remove it to better illustrate the padding the Table itself uses. This might give the idea that the footer has more padding than it actually does.

varl:
Thanks. I suspected as much though it was surprising from a reader POV.

On September 19, 2019 9:55:11 AM GMT+02:00, ismay <notifications@github.com> wrote:
>> @ismay why does the non-dense variant look more dense than the dense
>variant?
>
>That's because of the buttons in the dense example. The dense cells
>have less padding, but they still grow to accommodate their content.
>And since the buttons are large the rows end up looking larger. We
>could omit the buttons from the example and just show default and
>default with dense. That'd illustrate the difference better.
>
>(Reason for showing it with buttons was that dense seems like a prop
>you'd use to shrink rows that contain large content, like buttons, but
>it doesn't really help the example here)
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/dhis2/ui-core/pull/378#issuecomment-533013774

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

varl:
Should we merge this without jsdocs and add them to next or do it now? Other than that I think this looks good.

src/Table/TableCell.js
```diff
@@ -0,0 +1,32 @@
...
+export const TableCell = ({ children, colSpan, rowSpan, dense }) => (
+    <td colSpan={colSpan} rowSpan={rowSpan}>
+        {children}
+
+        <style jsx>{tableCellStyles}</style>
+        <style jsx>{`
+            td {
+                padding: ${dense ? '9px 12px' : '13px 12px'};
+                height: ${dense ? '36px' : '45px'};
+            }
```
varl: @ismay @Mohammer5 I would prefer a `dense` class here and using `classnames` to toggle it: `cx({ dense })` to avoid dynamic styles.

src/Table/TableCellHead.js
```diff
@@ -0,0 +1,33 @@
...
+export const TableCellHead = ({ children, colSpan, rowSpan, dense }) => (
+    <th colSpan={colSpan} rowSpan={rowSpan}>
+        {children}
+
+        <style jsx>{tableCellHeadStyles}</style>
+        <style jsx>{`
+            th {
+                padding: ${dense ? '9px 12px' : '13px 12px'};
+                height: ${dense ? '36px' : '45px'};
+            }
```
varl: Same here (`dense` class).

varl:
We should probably add `className` prop to all components as well.

varl:
Looks good all in all, I pushed a tweak to illustrate what I meant. If someone could implement JSDocs and add the `className` prop to all components in this PR this is good to go.

Mohammer5:
@varl Thanks, I'll add the doc blocks

Mohammer5:
@varl I've added the doc blocks. I need your approval so I can merge this one.
I'd like to mention, that I don't like `@prop {Node} children` for `propTypes.node`. It's confusing as `node` also includes strings and numbers

varl:
I agree with that @Mohammer5. We should tackle that before merging `next`.

varl:
@Mohammer5 was there a reason you didn't add the `className` prop to the components or just an oversight?

Mohammer5:
> @Mohammer5 was there a reason you didn't add the className prop to the components or just an oversight?

That slipped me..

Mohammer5:
@varl I've added the class name props

varl:
Thanks, looks good to me.

PR #379, opened 2019-08-16
fix: use correct disabled style & fix checkbox and radio disabled style

Mohammer5:
Fixes dhis2/design-system#6

This PR will change the theme's `disabled` color from `grey500` to `grey600`,
which will impact all components that are using `colors.disabled` of the theme.

### DIFF
--- a/src/Checkbox/Label.js
+++ b/src/Checkbox/Label.js
@@ -5,7 +5,7 @@ import cx from 'classnames'
 import { colors, theme, spacers } from '../theme.js'
 
 export const Label = ({ disabled, required, children }) => {
-    const className = cx('label', { disabled, required, disabled })
+    const className = cx('label', { disabled, required })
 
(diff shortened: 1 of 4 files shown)


src/Checkbox/Label.js
```diff
@@ -5,7 +5,7 @@ import cx from 'classnames'
 import { colors, theme, spacers } from '../theme.js'
 
 export const Label = ({ disabled, required, children }) => {
-    const className = cx('label', { disabled, required })
+    const className = cx('label', { disabled, required, disabled })
```
varl: Why are we duplicating `disabled` in the object?

Mohammer5: Oops, my bad, will remove it!

Mohammer5: Fixed

PR #381, opened 2019-08-21
feat(radio-group): introduce RadioGroup component

HendrikThePendric:
As discussed (extensively) on Slack, this is the styled version of the RadioGroup. This version is form-library agnostic and does not include help text.


src/RadioGroup/index.js
```diff
@@ -0,0 +1,118 @@
...
+                    display: flex;
+                    flex-direction: column;
+                }
+
+                .group.inline {
+                    display: flex;
+                    flex-direction: row;
+                }
+
+                .group-label.required::after {
```
Mohammer5: I haven't thought about this before either, but this won't work with RTL

src/RadioGroup/index.js
```diff
@@ -0,0 +1,118 @@
...
+
+                .valid {
+                    color: ${theme.valid};
+                }
+
+                .error {
+                    color: ${theme.error};
+                }
+
+                .warning {
```
Mohammer5: This should be before `error` so `error` has priority over `warning

src/RadioGroup/index.js
```diff
@@ -0,0 +1,118 @@
...
+                    color: ${theme.warning};
+                }
+            `}</style>
+        </>
+    )
+}
+
+RadioGroup.propTypes = {
```
Mohammer5: I think it's a better documentation to put the required props at the top

src/RadioGroup/index.js
```diff
@@ -0,0 +1,118 @@
...
+        })
+    ).isRequired,
+    required: propTypes.bool,
+    valid: statusPropType,
+    value: propTypes.any,
+    warning: statusPropType,
+}
+
+RadioGroup.defaultProps = {
+    required: false,
```
Mohammer5: Not necessary

src/RadioGroup/index.js
HendrikThePendric: I don't think this is an issue, as these are mutually exclusive props in the Button component. Will address it anyway.

src/RadioGroup/index.js
HendrikThePendric: As discussed on Slack: this is very true but also applies to many other components, so should be addressed in another branch

varl:
Why did we go for the `<RadioGroup options={...}/>` style API here when everything else uses composition?

```
<RadioGroup>
  <Radio />
  <Radio />
</RadioGroup>
``` 

There is no way for a consumer to use the `RadioGroup` component with e.g. a custom `<CustomRadio/>` component that I can see?

Mohammer5:
Fair point. @HendrikThePendric and I agreed, that this component is like a select, so it should accept options as well. But I just checked and saw that the `SelectField` also works with `<option />` as children, rather than using an `option` property.

I'm for changing it to allow more flexibility and to be consistent with our components' api design

HendrikThePendric:
Excellent point. Perhaps I should explain how this came about first, to provide some context: We were working on the ui-forms lib and concluded that all styles should be in ui-core. So we separated style from presentation and put it into ui-core. I think I focussed a bit too much on ui-forms and neglected to implement the ui-core component in a way that is consistent with the rest of the core-components.

**So, I would be very much in favour of reworking this component so it's more consistent with the rest too**

HendrikThePendric:
Below I will do some suggestions on how to implement this component in a way that is more in line with the rest.

In its final form (I am not talking necessarily about its implementation in any particular repo, but just about how this a radio-group will appear on the page) a radio group should have a few different parts:
- A group-label text, to appear above the `Radio`
- A set of radio inputs, to be arranged stacked or inline
- A help text

So to compose that, I suppose we would need the following components:
```jsx
<FormControl>
    <FieldLabel>Choose something</FieldLabel>
    <RadioGroup>
        <Radio />
        <Radio />
        <Radio />
    </RadioGroup>
    <Help />
</FormControl>
```

We have most of this. `RadioGroup` would need to be changed/simplified, and `GroupLabel` would have to be introduced.

@varl @Mohammer5 you agree with this approach?

HendrikThePendric:
And what is perhaps also relevant to consider is the following:
- @Mohammer5 and I decided that we should avoid having any styling in `ui-forms`, this makes a nice clean separation of concerns.
- But on the flip side, it will quite likely trigger the creation of many of these small styling components, such as `FieldControl` (already merged in) and now `GroupLabel`

I am undecided on whether this is a good or bad thing....

What do you think @varl

HendrikThePendric:
FYI: please note this PR: https://github.com/dhis2/ui-core/pull/388
I think that is a better solution...

varl:
> I am undecided on whether this is a good or bad thing....

Generally that's a good thing _if_ it means that if someone needs to create a forms library for some other form library (or lack thereof) they can reuse the small styling components (`FieldControl`, `GroupLabel`, etc.), which I assume is the case?

HendrikThePendric:
>  which I assume is the case?
Correct, thanks

PR #382, opened 2019-08-21
feat(form-control): introduce FormControl component

HendrikThePendric:
Behold the FormControl :-)


stories/FormControl.stories.js
```diff
@@ -0,0 +1,24 @@
...
+
+import markdown from './info/atoms/formControl.md'
+
+storiesOf('FormControl', module)
+    .addParameters({
+        notes: {
+            markdown,
+        },
+    })
+    .add('Default', () => (
```
Mohammer5: I think this story only makes sense when using 2 `FormControl`s

HendrikThePendric: True. I considered a wrapper with a border, but that would create as much confusion as it would clarify things. Two FormControls makes sense. Added them just now.

PR #384, opened 2019-08-21
feat(fileinput): introduce FileInput component

HendrikThePendric:
This is the FileField, which was pulled out of ui-forms


cooper-joe:
Hey, thanks for tagging me. I have some feedback regarding the functionality here, I am not sure if it has been discussed before or if it is out of scope of this component.

1. I think the filename and error text should be presented outside of the button. I think it is confusing for a user if the button text changes. (see screenshot below)
2. Uploaded file(s) should be removeable, in the current example there is not an intuitive way of handling this. I suggest having a 'remove' action next to the filename. (see screenshot below)
3. I've seen the need for multiple file upload in DHIS2, so I'm thinking it would be wise to include it here. This is another +1 for displaying the filename outside of the button, it allows for easy multi-file support.

Here is some examples of the states I am discussing above:
image

Mohammer5:
@cooper-joe Can we add a default text when no file has been selected, placed where the file names are listed? E. g. "No file(s) selected yet"

Most of the time only one file can be selected and with this little text, the rest of the form won't jump vertically

cooper-joe:
> @cooper-joe Can we add a default text when no file has been selected, placed where the file names are listed? E. g. "No file(s) selected yet"
> 
> Most of the time only one file can be selected and with this little text, the rest of the form won't jump vertically

Yes, we certainly can include that. I would like it to be an optional string, I think there could be times (using the component in a space dense area where it is not always used) where it be good to not have this.

Another thing I forgot to add in my notes above was the need for constraints: "Only PDF and JPG files allowed" etc. These can be included with the `help` component, but I think it makes sense to include them here by default so they are not easily forgotten!

HendrikThePendric:
Thanks @cooper-joe your design makes a lot of sense. I would prefer if we could also put a "label" above the button which would correspond with the field name, for example let's use a profile picture:
```
Profile picture (field name)
[Upload file] (button)
mypic.jpg [remove] (currently uploaded file)
There was an error (error message)
```

HendrikThePendric:
Since I am noticing that there is quite a bit to discuss reg. these form fields in general, I am going to start a Slack channel to discuss things further.

cooper-joe:
> Thanks @cooper-joe your design makes a lot of sense. I would prefer if we could also put a "label" above the button which would correspond with the field name, for example let's use a profile picture:
> 
> ```
> Profile picture (field name)
> [Upload file] (button)
> mypic.jpg [remove] (currently uploaded file)
> There was an error (error message)
> ```

Yes, totally agree that having a label above would make sense.

Mohammer5:
> 2. Please add a comment to the component that it's not a controlled component.
> This means that you can't set the value through the value prop directly.

I think this is still good to have in the component

HendrikThePendric:
I've completely reworked this component according to the design specs that @cooper-joe [made available here](https://github.com/dhis2/design-system/blob/file-input/atoms/fileupload.md#loading-progress).

I'll explain how I got to the current implementation below:
- I initially tried to implement this as a single component, but pretty quickly realised that, mainly because this component relies on quite a few strings, that it was getting way too many props.
- So, instead, I went for an approach with children. Initially I was hoping I could just keep the `FileUpload` very small, and then build the full `FileUploadField` in `ui-forms` using some of the form helper components that we already have (i.e. `Help`, `FormControl`, `Legend`)
- Unfortunately, this didn't work out, because of some subtle style differences, and the fact that `Help` only accepts a string (IMO a good thing). So I had to create 2 new components: one for the items in the file-list, and one for the placeholder text. Because these components are not really suitable to use anywhere else, will probably always be used alongside the main component, and are fairly small, I decided to use the name-spacing approach.

How I envisage using it, is probably best demonstrated in the last story: "Design system stacking order"

Let me know what you guys think.

HendrikThePendric:
And one more note that ties in with the remark @Mohammer5 made:
> Please add a comment to the component that it's not a controlled component.
This means that you can't set the value through the value prop directly.

I don't really feel like this remark is relevant given the current implementation. However, we could possibly expect some problems further down the line (i.e. in ui-forms), because we can't access the input element directly. For example, if we need to clear the files from the input element. However, I'd like to address this if and when it this becomes relevant. I haven't thought of an actual case that would not work.

HendrikThePendric:
> I don't really feel like this remark is relevant given the current implementation

Although, come to think of it, it could help explaining why this input is different from the other components...

src/FileInput/SelectedFile.js
```diff
@@ -0,0 +1,81 @@
...
+                </span>
+            )}
+        </span>
+
+        <style jsx>{`
+            p {
+                display: flex;
+                margin: 0;
+                padding-top: ${spacers.dp8};
+                line-height: 14px;
```
Mohammer5: Should be `line-height: 18px`.

I can see that @cooper-joe has not included line-heights in his design specs (please do that @cooper-joe).
But I can see a difference between the story and the design (the stories also seem quite dense in this regard)

src/FileInput/SelectedFile.js
```diff
@@ -0,0 +1,81 @@
...
+                flex-shrink: 0;
+            }
+            .text {
+                flex-grow: 1;
+                flex-shrink: 1;
+            }
+            .label {
+                font-size: 14px;
+                color: ${colors.grey900};
+                margin-right: 12px;
```
Mohammer5: I think this should be `margin-left` on the `.action`.
It's causing a margin that's outside of the `FileInput` box when the file name is displayed in multiple line.
This might cause that the `removeText` will have an oddly looking margin when it falls into its own line,
but I think an internal layout issue is less of a problem than one that can influence the elements in the document flow.

src/FileInput/SelectedFile.js
HendrikThePendric: Thanks for pointing that out to me. You comment has reminded me of what I agreed with Joe: some of the spacing in the specs was not really clear to me, so I would ask him to review once I was done...
@cooper-joe : could you have a look [here](https://deploy-preview-384--dhis2-ui-core.netlify.com/?path=/story/fileinput--design-system-stacking-order) to see what I need to change about the spacing

src/FileInput/SelectedFile.js
HendrikThePendric: > This might cause that the removeText will have an oddly looking margin when it falls into its own line
Yes that's why I changed this from a margin-left on the remove/cancel-link to a margin-right on the text.

I see the problem you are describing with the margin when I inspect the DOM... But if you inspect the parent elements, it's all correct. I don't think this is an actual problem. If it would be, we'd already be seeing it in the "File list" story.

I don't like having a margin-left on the link, for the reasons you mentioned, but I think I can fix it in another way, using the ::after pseudo element. That way we'll both be happy.

src/FileInput/SelectedFile.js
cooper-joe: Two changes:
- Padding under button increase +4px
- Padding between help/error and the file list +4px (not an increase in line height in the actual file list)

The line heights are correct, it's just those paddings between 'sections' of the component. I know what you mean with the specs looking a little different/less dense than in browser, I think this has a lot to do with browser/OS font smoothing. I've just measured pixel for pixel and these were the only changes haha.

Mohammer5: @cooper-joe could you also add line-height values to the specs? This will def. help in the future to use the correct values.
Btw, it's generally a bad idea to use
(in-depth info: https://iamvdo.me/en/blog/css-font-metrics-line-height-and-vertical-align)

* font-size to determine an elements height
font-size does not determine the height of an element, the font itself does (if there's no explicit line-height value).

* use the same value for both font-size and line-height
When using the same line-height as font-size, you might end up with touching characters.
We can solve this with padding of course, but as long as we have control over the font, we can also do this with line-height (which we should use anyways) without having to use padding

PR #385, opened 2019-08-21
chore(eslint): extend react-app in eslintrc and install required deps

### LINKED ISSUE #383: Linter setup is broken
```[Error - 1:03:32 PM] Cannot read config file: /Users/hendrik/apps/ui/.eslintrc.js Error: Cannot read property 'eslint' of undefined```

So the problem is that the `config` object exported by `@dhis2/cli-style` is undefined.

### PR DESCRIPTION
**Pull request by HendrikThePendric**
This fixes the linting setup as discussed on Slack. Some of the dependencies such as `eslint-plugin-react-hooks` might not be required for the current project, but I've just stuck to the [installation instructions](https://www.npmjs.com/package/eslint-config-react-app#usage-outside-of-create-react-app) for react-app.

Fixes #383.

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -34,8 +34,17 @@
     "@storybook/addons": "^5.1.8",
     "@storybook/components": "^5.1.10",
     "@storybook/react": "^5.1.11",
+    "@typescript-eslint/eslint-plugin": "1.x",
+    "@typescript-eslint/parser": "1.x",
     "babel-eslint": "10.x",
     "babel-loader": "^8.0.6",
+    "eslint": "6.x",
```
varl: This should not be in the dependency list, it is satisfied through `d2-style` and we do not want possible collisions.

package.json
```diff
@@ -34,8 +34,17 @@
     "@storybook/addons": "^5.1.8",
     "@storybook/components": "^5.1.10",
     "@storybook/react": "^5.1.11",
+    "@typescript-eslint/eslint-plugin": "1.x",
+    "@typescript-eslint/parser": "1.x",
```
varl: Why add TypeScript plugins/parser?

package.json
```diff
@@ -34,8 +34,17 @@
...
     "babel-eslint": "10.x",
     "babel-loader": "^8.0.6",
+    "eslint": "6.x",
+    "eslint-config-react-app": "^5.0.1",
+    "eslint-plugin-flowtype": "3.x",
+    "eslint-plugin-import": "2.x",
+    "eslint-plugin-jsx-a11y": "6.x",
+    "eslint-plugin-react": "7.x",
+    "eslint-plugin-react-hooks": "1.x",
```
varl: Not sure we should have this here.

package.json
```diff
@@ -34,8 +34,17 @@
     "@storybook/addons": "^5.1.8",
     "@storybook/components": "^5.1.10",
     "@storybook/react": "^5.1.11",
+    "@typescript-eslint/eslint-plugin": "1.x",
+    "@typescript-eslint/parser": "1.x",
     "babel-eslint": "10.x",
     "babel-loader": "^8.0.6",
+    "eslint": "6.x",
+    "eslint-config-react-app": "^5.0.1",
+    "eslint-plugin-flowtype": "3.x",
```
varl: Not sure we should have this either.

HendrikThePendric:
@varl Instead of replying to each individual comment, I will reply here...
I've had another go at this PR. What I did initially was:
- I simply installed `eslint-config-react-app` and followed the installation instructions.
- These instructions include installing all these dependencies, and not including them will result in a bunch of errors about missing dependencies.

But I fully agree it's not great to have to add all these dependencies. So for my second attempt, I switched to [eslint-plugin-react](https://www.npmjs.com/package/eslint-plugin-react). This doesn't require any other dependencies to be installed, so I think chances of getting this PR approved have increased a lot :-)

Mohammer5:
I'll leave this one to @varl as he's the chiefdom of this stuff

yarn.lock
```diff
@@ -5,7 +5,6 @@
 "@babel/cli@^7.5.5":
   version "7.5.5"
   resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.5.5.tgz#bdb6d9169e93e241a08f5f7b0265195bf38ef5ec"
-  integrity sha512-UHI+7pHv/tk9g6WXQKYz+kmXTI77YtuY3vqC59KIqcoWEjsJJSG6rAxKaLsgj3LDyadsPrCB929gVOKM6Hui0w==
```
varl: Why are all the integrity strings removed? Are you using an old version of Yarn?

HendrikThePendric: Thanks for spotting. Something weird had been going on on my local machine. I have been updating yarn, but it was actually failing to switch to the newer version. I've now managed to do so, and produced a new `yarn.lock` file

HendrikThePendric:
> Errors or Warnings?
@varl Warnings, but still.... You don't want a long list of them every time you run `yarn start`... I think the current solution is better/more lightweight than the first approach.

varl:
Aight, good solution until we figure out how to do the `project/react` group in `d2-style`.

varl:
Fixes #383.

PR #389, opened 2019-08-23
refactor(select,input): drop -field suffix on components

varl:
BREAKING CHANGE: **InputField** renamed to **Input**

**SelectField** renamed to **Select**


varl:
We also want to do the restructure (drop the automagic `index.js` resolution) but that's a different PR.

PR #392, opened 2019-08-26
chore: demo info addon with Button

Mohammer5:
This one looks a bit better than the one with Chapters

I've not been able to remove the `"Unknown" Copmonent` from the prop table


varl:
I prefer having the `props` and `jsx` in a separate pane and have the main pane be dedicated to the component.

Storybook shines as a development tool and testing, I do not think it does a great job as user facing documentation, and that we should have manually curated docs for that.

We should create something like @amcgee has done for the `app-runtime` for this library as well: https://runtime.dhis2.nu/#/hooks/useConfig

PR #395, opened 2019-08-27
chore(icons): clean up and improve consistency

HendrikThePendric:
## Consistency improvements
- All icons now use the same shared style (apart from Loading)
- Only SVG icons now live in the `./icons` folder (Switch was moved)
- All SVGs now share the same attributes: `xmlns`, `viewBox` and `className`. I removed `height` and `width` because we are controlling this via CSS.
- There is now also a shared prop type for icons. Currently it only contains the className property

## Switch refactor
- The icon wasn't an SVG so I moved it into the component folder
- I moved the icon styles to the `styles.js` file, which now exports `styles` and `switchIconStyles`
- The `SwitchIcon` component without the styles was a very small component, so I moved it to the main index.js file, same as we find the `Input` component there

This isn't a breaking change, but I decided to start this from the `next` branch anyway, since there are quite a few changes.


src/icons/Checkbox.js
```diff
@@ -1,43 +1,45 @@
...
             className={className}
         >
             <path d="M38 6H10c-2.21 0-4 1.79-4 4v28c0 2.21 1.79 4 4 4h28c2.21 0 4-1.79 4-4V10c0-2.21-1.79-4-4-4zM20 34L10 24l2.83-2.83L20 28.34l15.17-15.17L38 16 20 34z" />
+            <style jsx>{styles}</style>
         </svg>
     )
 }
+Checked.propTypes = iconPropType
 
 export function Indeterminate({ className }) {
```
varl: This is broken with the refactor: 
image

src/icons/Status.js
```diff
@@ -12,27 +14,15 @@ export function Valid({ className }) {
...
         <svg
             xmlns="http://www.w3.org/2000/svg"
-            width="48"
-            height="48"
             viewBox="0 0 48 48"
             className={className}
         >
             <path d="M24 4C12.95 4 4 12.95 4 24c0 11.04 8.95 20 20 20 11.04 0 20-8.96 20-20 0-11.05-8.96-20-20-20zm-4 30L10 24l2.83-2.83L20 28.34l15.17-15.17L38 16 20 34z" />
-            <style jsx>{`
-                svg {
```
varl: Status icons no longer have the correct colour:
image

Also true for the other statuses.

src/icons/Status.js
```diff
@@ -132,9 +90,7 @@ export function Loading({ className }) {
     return (
         <svg
             xmlns="http://www.w3.org/2000/svg"
-            width="48"
-            height="48"
-            viewBox="22 22 44 44"
+            viewBox="0 0 48 48"
```
varl: Loader status is also broken:

broken-loader

varl:
I do not want the changes for making the icons share more code, though moving the Switch makes sense. Could we split that out to a new PR and close this?

src/icons/Status.js
HendrikThePendric: Aha... yes I removed them too... a little too much cleaning up, sorry. I'm going to fix it in the same way as it was. 

However, in `./src/icons/style.js` I've applied the rule `fill: inherit;`. This is quite handy as it enables us to apply colour to the SVG from a parent element very easily...

For example, this produces a box with a green border and green SVG.
```jsx
const Demo = () => (
    <div>
        <ArrowUp />
        <style jsx>{`
            div {
                border: 1px solid green;
                padding: 24px;
                fill: green;
            }
        `}</style>
    </div>
)
```
So... if we like this type of approach, perhaps we could utilise it here too. IMO it's a more flexible approach then hard-coding the styles into the SVGs.

src/icons/Status.js
```diff
@@ -12,27 +14,15 @@ export function Valid({ className }) {
...
             viewBox="0 0 48 48"
             className={className}
         >
             <path d="M24 4C12.95 4 4 12.95 4 24c0 11.04 8.95 20 20 20 11.04 0 20-8.96 20-20 0-11.05-8.96-20-20-20zm-4 30L10 24l2.83-2.83L20 28.34l15.17-15.17L38 16 20 34z" />
-            <style jsx>{`
-                svg {
-                    fill: ${theme.valid};
-                    width: 24px;
-                    height: 24px;
-                    margin-right: ${spacers.dp4};
```
HendrikThePendric: Seems I didn't just remove the fill style but also the `margin-right: ${spacers.dp4};`. I don't actually think that this style belongs here. If I remember it correctly, we want to end up with an icon repo too. Components that use these icons should expect them to all behave the same, and it should be possible for these components to control things like fill, margin and size.

src/icons/Status.js
HendrikThePendric: Oops this one should have had a custom viewBox.

src/icons/Checkbox.js
HendrikThePendric: I fixed this now, but not quite as I would have liked... For all these icons I've been getting the production svg at 48px, however for `Indeterminate` [this seems to be unavailable](https://github.com/google/material-design-icons/tree/master/toggle/svg/production). So I've had to resort to putting `viewBox="0 0 24 24"`.

HendrikThePendric:
> I do not want the changes for making the icons share more code, though moving the Switch makes sense. Could we split that out to a new PR and close this?

If you really insist, sure. However, with the exception of the Status icons, I am convinced that the inconsistencies on the icons were simply due to the fact that they were added quickly at different times. For example: some had some style and others didn't, some accepted a className prop and others didn't... 

I did introduce a few regression on Checkbox and the status icons, but these were pretty quick to fix. If you don't mind, I'd like to discuss this PR tomorrow before closing it. I don't object to removing the shared parts, but at least these icons should be implemented in a consistent way and prior to this PR that was not the case.

HendrikThePendric:
> I really prefer that changing one component only affects that component and accept that we have minor duplication of code to facilitate the much more important isolation between components. This was an early design goal of ui-core that hasn't changed.

Your comment above is line with the suggestion I did in my last comment:
- You prefer to not have shared code, and prefer some duplication in this case
- I want a consistent implementation

So if I update this PR as follows, I think we should both be happy:
- Remove shared styles and replace by styles per icon-component (but implement for all icons this time)
- Remove shared prop-type and replace by prop-type definitions per component

HendrikThePendric:
@varl I think https://github.com/dhis2/ui-core/pull/395/commits/951bb4a442328343ada8b37d0bc5ada9cec0f7d2 addresses your concerns

varl:
Cool, thanks @HendrikThePendric. I'll take another look.

src/icons/Status.js
```diff
@@ -74,23 +74,27 @@ export function Error({ className }) {
...
                 svg {
-                    fill: ${theme.error};
-                    width: 24px;
+                    fill: inherit;
                     height: 24px;
+                    width: 24px;
+                    vertical-align: middle;
+                    pointer-events: none;
                     margin-right: ${spacers.dp4};
+                    fill: ${theme.error};
```
varl: `fill` is duplicated between line 83 and 89, both `inherit` and a colour. Is that intentional?

varl: It's the same in other components.

HendrikThePendric: Nope that just highlights I shouldn't have tried to make these changes quickly and very late at night... Fixed it now

PR #396, opened 2019-08-28
refactor: remove compact from button strip

varl:
BREAKING CHANGE: Removes the `compact` prop and styles from the ButtonStrip.


varl:
This just moves the `compact` styles to the SplitButton so that component does not have a dependency on ButtonStrip, and ButtonStrip does not have a compact mode.

PR #397, opened 2019-08-30
docs: introduce user docs

varl:


varl:
This publishes the user facing docs at the root URL and the demo site under `/demo`:

- docs: https://deploy-preview-397--dhis2-ui-core.netlify.com
- demo: https://deploy-preview-397--dhis2-ui-core.netlify.com/demo

varl:
`props` tables are generated by putting e.g. the string `REPLACE_WITH_PROPS_TABLE ` somewhere in the `button.md` file and then running:

```sh
npx props-table src/Button/index.js --replace REPLACE_WITH_PROPS_TABLE --output docs/components/button.md
```

The `api.md` file is simpler and can be overwritten freely:

```sh
npx props-table src/**/*.js --output docs/api.md
```

.gitignore
```diff
@@ -12,6 +12,7 @@
 # production
 /build
 /dist
+/dest
```
Mohammer5: The name is a bit off, maybe `stories` suits this one better?

varl: Agreed, I'll change it to something else. It's not `stories` though, the stories are under `dest/demo`.

varl:
Now with more automation to pull out the api docs.

amcgee:
Nice!  Looks like there are still some manually-compiled docs (under `docs/components/`) - is the idea to keep those?  If so they should be listed in the sidebar (`docs/_sidebar.md`), if not they should probably be removed for now?

varl:
They'll be removed for now. I'm working on adding jsdocs for all components that we have.

varl:
Aight, I've added basic jsdocs for all components, will merge this into `next` so we can add more documentation strings in a self-contained manner.

PR #401, opened 2019-09-02
refactor: reorganize and move away from node module resolution hacks

### LINKED ISSUE #394: Allow outlined inputs without labels
Outlined/regular (whatever the not-filled style is called...) text inputs are valid without labels. Currently the label is required. When a label is not included the height of the component should not include an empty label area.

varl:
Fixes #394.


.storybook/css-reset-decorator.js
```diff
@@ -1,6 +1,6 @@
 import React, { Fragment } from 'react'
 
-import { CssReset } from '../src'
+import { CssReset } from '../src/index.js'
```
amcgee: The `.js` extension is probably superfluous here - doesn't really hurt, but a bit cleaner and easier to refactor as a subdirectory later if you use `../src/index` (bad example for the refactoring case... `./AlertBar/Action` below is better)

amcgee: It looks like you added it intentionally though?

varl: I did.

It is to align closer to the ECMAScript spec than to specific module resolver implementations. Node and Webpack have implemented the module resolution algorithm based on [RequireJS](https://requirejs.org/docs/api.html#jsfiles). The ECMAScript Module definition is much stricter and is based on paths instead of module names.

Given the file `./src/index.js` RequireJS allows these to be equivalent:

- `./src/index.js`
- `./src/index`
- `./src`

Whereas in ESM only the `./src/index.js` path would be valid. I prefer aligning with ESM in our libraries.

I think `api-index.js`, `external-api.js`, `main.js`, or `api.js` is perhaps a less confusing name than `index.js` even though it is technically an index in the literal sense.

varl: This particular import is in Storybook though so it doesn't matter. 🤷‍♂

PR #404, opened 2019-09-04
feat: add dense prop to static table

ismay:
Main goal with this PR is to add the `dense` prop to the cells and rows of the static table (or show what I meant with it at least).

Another small change is that instead of using a `div` to set a `min-height` on the table-cell, this sets a `height` on the table-cell. Which effectively does the same as the cell can still grow larger in height, and it doesn't require an extra element. This way we don't need the global styles.

I've also removed the `min-height` settings from the table row, as they were overriding the table cell heights (when set to `height`). We could put it back in the same manner as the cells (with a `height` instead of `min-height`, and a `dense` prop), but that does mean that setting `dense` on the row will override the cells, which might be confusing.

Btw. I marked my commits as chores since it's a pr for a pr. I'm not sure what the convention is here so let me know if I've done it wrong.

### DIFF
--- a/src/Table/TableCell.js
+++ b/src/Table/TableCell.js
@@ -5,35 +5,28 @@ import css from 'styled-jsx/css'
 const tableCellStyles = css`
     td {
         border-bottom: 1px solid #e8edf2;
-        padding-left: 12px;
-        padding-right: 12px;
-        height: 45px;
         font-size: 14px;
         line-height: 18px;
     }
-
-    :global(tbody) td {
-        padding-top: 13px;
-        padding-bottom: 13px;
-    }
-
-    :global(tfoot) td {
-        height: 36px;
-        padding-top: 9px;
-        padding-bottom: 9px;
-    }
 `
 
-export const TableCell = ({ children, colSpan, rowSpan }) => (
+export const TableCell = ({ children, colSpan, rowSpan, dense }) => (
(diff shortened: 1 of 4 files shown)


Mohammer5:
Looks good to me already, just one request:
Could you adjust the 'dense' story so that one column is not dense and contains a button?
Otherwise it wouldn't be clear to me why you'd have to add the dense prop on the cell level instead of row-, tbody or table level

ismay:
> Could you adjust the 'dense' story so that one column is not dense and contains a button? Otherwise it wouldn't be clear to me why you'd have to add the dense prop on the cell level instead of row-, tbody or table level

Yeah that's a good point. The difference with and without `dense` is actually fairly small. I've pushed a commit that demonstrates it (this is not how I'd want to merge this, but just for the sake of demonstrating). You'll see that the entire top row is `dense`. And there are two rows with buttons, one without, and one with `dense`. So when you include a button, the `dense` prop makes sense to me.

For an entire row, I think it'd be easier to set it on the row (or in fact the table). Though the api might then be confusing to a user as it can override the cell ("Why is this dense prop on the cell not having any effect?"). Maybe it'd be good to discuss how we want to approach that.

ismay:
Updated the `dense` story to have buttons in all rows, and the dense prop only on the cells with buttons in them.

PR #414, opened 2019-09-09
Feat/file input

HendrikThePendric:
This could be a potential solution to the next branch becoming out-of-date for reasons we don't understand yet (at the time of writing).

This is draft PR, because I'd prefer to find a proper solution....


varl:
I'm working on the solution. I rebased `next` on `master` to bring it up to date. I was not aware additional work had been merged. Sorry about that.

PR #419, opened 2019-09-11
fix: checkbox value should be optional

HendrikThePendric:
I believe that the `value` attribute on a checkbox element is not mandatory, so the corresponding prop on our component shouldn't be either.

Source:
> If the value attribute was omitted, the default value for the checkbox is on , so the submitted data in that case would be subscribe=on.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value

### DIFF
--- a/src/Checkbox.js
+++ b/src/Checkbox.js
@@ -119,7 +119,7 @@ class Checkbox extends Component {
 Checkbox.propTypes = {
     onChange: propTypes.func.isRequired,
 
-    value: propTypes.string.isRequired,
+    value: propTypes.string,
     name: propTypes.string.isRequired,
     label: propTypes.string.isRequired,
     tabIndex: propTypes.string,
--- a/src/Checkbox/Input.js
+++ b/src/Checkbox/Input.js
@@ -1,4 +1,4 @@
-import React, { Component, Fragment, createRef } from 'react'
+import React, { Component, createRef } from 'react'
 import propTypes from 'prop-types'
 
 export class Input extends Component {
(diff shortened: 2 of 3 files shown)


varl:
You're right that a HTML checkbox has an optional `value` property.

Not sure why we made the `value` required, we did the same for Radio.

Perhaps the value can always be optional instead, I would like that consistency:

```sh
src/Select.js
109:    value: propTypes.string,

src/Input.js
106:    value: propTypes.string,

src/Radio.js
177:    value: propTypes.string.isRequired,

src/SplitButton.js
97:    value: propTypes.string,

src/MenuItem.js
102:    value: propTypes.any,

src/Checkbox.js
122:    value: propTypes.string.isRequired,

src/Button.js
71:    value: propTypes.string,

src/Select/InternalSelect.js
106:    value: propTypes.string,

src/FieldLabel/Filled.js
159:    value: propTypes.bool,

src/FieldLabel/Regular/Label.js
54:    value: propTypes.string,

src/Input/InternalInput.js
91:    value: propTypes.string,

src/FieldLabel/Regular.js
138:    value: propTypes.bool,
```

HendrikThePendric:
@varl : I agree with what you wrote and have removed `isRequired` from the radio. However, there were some linter issues in the file, which I had to solve before I could commit. 

Do you agree with how I solved the lint problems with the `Input`?
Since all the props were just put straight onto the `input` element, I thought in this case just spreading the props, and thus avoiding to have to declare prop-types, was not a problem. The prop-types have already been defined on the component that is being exported.

PR #421, opened 2019-09-13
feat(css-variables): add css variable injection component

ismay:
A sketch for a way to inject css variables for users that don't use a css-in-js approach in their project:

Screenshot 2019-09-13 at 17 07 29


stories/CssVariables.stories.js
```diff
@@ -0,0 +1,17 @@
...
+import { storiesOf } from '@storybook/react'
+import { CssVariables } from '../src'
+
+const App = ({ children }) => <div>{children}</div>
+
+storiesOf('CssVariables', module).add('Default', () => (
+    <App>
+        <CssVariables />
+
+        <p>
```
Mohammer5: I think this text should be in the jsdoc block of the component.
Rather than explaining here what you could do with the `CssVariables` component,
I'd be expecting a show-case, maybe something like this:

```jsx
<p>
  This {' '}
  <span style="color: var(--colors-blue500);">text</span> {' '}
  contains snippets with different {' '}
  <span style="color: var(--colors-red500);">colors</span>
</p>
```

src/CssVariables/index.js
```diff
@@ -0,0 +1,34 @@
...
+ * Maps over the exported sections of the theme
+ */
+const renderTheme = () => {
+    const sections = Object.entries(theme)
+
+    return sections.map(renderThemeSection).join('\n')
+}
+
+/*
+ * Inject theme values as CSS variables for users that are using vanilla css.
```
Mohammer5: When using comments, jsdoc block comments would be best (we're moving towards them anyways, see the components in the `next` branch, e. g. the [`Help`](https://github.com/dhis2/ui-core/blob/next/src/Help.js) component)

src/CssVariables/index.js
```diff
@@ -0,0 +1,34 @@
...
+    const variables = Object.entries(section)
+
+    return variables.map(([ varKey, varVal ]) => (`--${sectionKey}-${varKey}: ${varVal};`)).join('\n')
+}
+
+/**
+ * Maps over the exported sections of the theme
+ */
+const renderTheme = () => {
+    const sections = Object.entries(theme)
```
Mohammer5: Can we make this a parameter? Then all the functions, except the component are 100% pure (and the theme could eventually be made a parameter of the `CssVariables` component (if we ever want that)

stories/CssVariables.stories.js
ismay: Yeah I just based it on the text of the CssReset component. I'll modify it to demonstrate what it does directly 👍

src/CssVariables/index.js
ismay: Np, I can move it over to a jsdoc comment. I do find the jsdoc comment on the proptypes a bit redundant by the way. Why are we doing that?

src/CssVariables/index.js
ismay: That sounds like a good idea. That way it's just a generic component that injects css variables, and it's not tied too much to our implementation. So people could also use it to inject some extra variables while they're at it for example.

stories/CssVariables.stories.js
ismay: We could make this text more specific, let me know what you think. Alongside this we could also add a story which doesn't rely on the default value for `variables` but sets its own. Just to demonstrate how that'd work. What do you think?

ismay:
I noticed by the way that `master` currently has linting errors. They aren't caught because linting isn't run on travis. I think we should modify the travis script to run our style checks as well, otherwise we'd only notice these errors if we discover them whilst committing or working on sth.

src/CssVariables/index.js
Mohammer5: For generating markdown prop type tables

varl:
> 
> 
> I noticed by the way that `master` currently has linting errors. They aren't caught because linting isn't run on travis. I think we should modify the travis script to run our style checks as well, otherwise we'd only notice these errors if we discover them whilst committing or working on sth.

I've been cleaning that up on `next` as I have touched almost all of the components while adding JSDocs. We don't need to do that on `master` as well.

ismay:
> I've been cleaning that up on next as I have touched almost all of the components while adding JSDocs. We don't need to do that on master as well.

@varl Ah nice. So you're planning on adding linting on CI there as well?

src/CssVariables/index.js
ismay: So those jsdoc comments are used with this lib then: https://www.npmjs.com/package/storybook-addon-react-docgen?

Mohammer5: No, they're used by [`react-docgen`](https://www.npmjs.com/package/react-docgen) (not the storybook addon, see @varl's PR that has been merged into the `next` branch already):

https://github.com/dhis2/ui-core/pull/397/files?file-filters%5B%5D=.json#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

stories/CssVariables.stories.js
Mohammer5: Yeah, we should have one story for every possible scenario

src/CssVariables/index.js
```diff
@@ -1,16 +1,19 @@
...
  * @module
- * @description Injects our theme as CSS custom properties
+ * @description Injects the variables prop as CSS custom properties
  * @param {CssVariables.PropTypes} props
  * @returns {React.Component}
  * @example import { CssVariables } from @dhis2/ui-core
  * @see Live demo: {@link /demo/?path=/story/cssvariables--default|Storybook}
  */
 const CssVariables = ({ variables }) => {
+    if (!variables) {
```
Mohammer5: I think this is unnecessary when marking the prop as required.

stories/CssVariables.stories.js
ismay: I resolved this because CssVariables and ThemeCssVariables both have a story. Maybe we need to update the wording, but to me it seems all right.

stories/CssVariables.stories.js
```diff
@@ -7,13 +7,13 @@ const App = ({ children }) => <div>{children}</div>
 
 storiesOf('CssVariables', module).add('Default', () => (
     <App>
-        <CssVariables />
+        <CssVariables variables={{ 'color-red': '#a83c32' }} />
```
Mohammer5: Can we use a name here that doesn't look like you'd be overwriting a theme value?
Right now it looks like the intended use-case is do customize the theme while that's exactly what we don't want

src/CssVariables/index.js
ismay: I wanted to return early here because the mapping will crash if there is no variable prop.

stories/CssVariables.stories.js
ismay: Yeah no problem. Though they wouldn't be overriding the theme. The internals of our components aren't affected by anything you set here. I think Viktor's concerns were on using css-variables internally which users would then be able to override.

And if users want to import the theme and then override it there's not much you can do about that.

src/CssVariables/index.js
Mohammer5: If the variables are required, crashing would be a "desired" behavior, wouldn't it?

varl:
I haven't had a chance to look at this yet, so if you open a PR on it, please allow me to go through it.

src/CssVariables/index.js
ismay: I do find it a bit extreme. This is **without** handling a lack of variables:

Screenshot 2019-09-16 at 14 59 59

What will be displayed depends on the error handling the app does of course. And alongside this error there are also various other error messages in the console (that are a bit more helpful).

And this is **with** handling a lack of variables (so just a console message):

Screenshot 2019-09-16 at 15 00 35

I guess it's not incorrect to crash when a component's API isn't correctly used. I do always try to neatly handle any potential errors though, and just the prop-type validation failure does seem a bit more civil to me. But I don't have a strong opinion on this, I can see the point of both approaches.

ismay: On second thought, just letting it throw doesn't seem too bad either. I'll leave it to what @varl prefers, I'm fine with either direction.

ismay:
> Why is there both a CssVariables and a ThemeCssVariables component?

> To me it seems they should be combined into one. We should not encourage users to insert arbitrary CSS variables through our components. If they want their own CSS variables they can use whatever CSS loading technology to do that already. If they are using CSS-in-JS then they have access to our JavaScript variables already.

Originally we had that, a single CssVariables component that just inserted the theme variables on mount. Not everyone might want to insert all the theme variables though, so that's why we separated it out into two. Users can insert part of the theme that way, and it seemed like a useful component to have in its own right.

ismay:
The abstraction seems nice to me, but if you think that users should only be allowed to insert the entire theme or not, we could also just export the ThemeCssVariables component for example.

src/CssVariables/index.js
varl: This entire problem can be sidestepped by having a single component that does not accept any props.

varl:
To insert a partial theme a user would have to do:

```js
const { colors, CssVariables } from '@dhis2/ui-core'

const MyTheme = () => (
    <CssVariables variables={
        'color-red': colors.red050,
    } />
)

const MyComponent = () => (
   {* this should be in a class *}
    <span style={'color': 'var(--color-red)'}>foobar</span>
)
```

What's the utility here?

Even if we don't expose this functionality externally, I don't see the raison d'être for it internally either. What do you have in mind for it?

ismay:
> Even if we don't expose this functionality externally, I don't see the raison d'être for it internally either. What do you have in mind for it?

So they could just do:

```
const { colors, CssVariables } from '@dhis2/ui-core'

const MyTheme = () => (
    <CssVariables variables={colors} />
)

const MyComponent = () => (
   {* this should be in a class *}
    <span style={'color': 'var(--color-red)'}>foobar</span>
)
```

varl:
> So they could just do:

Ah! I understand now.

We can cover this use case with a single component, and lock down the API so a user cannot use our component to inject any CSS variables:

```js
CssVariables.propTypes = {
    colors: propTypes.bool,
    theme: propTypes.bool,
    spacers: propTypes.bool,
}
```

The default could be to inject everything, and then other parts can be turned off.
 
Or, default is to inject nothing, and force the users to take a stance:

```
<CssVariables colors theme spacers />
```

It matches the principles driving the component API in ui-core better as well.

varl:
We discussed this on Slack and settled on:

- A single `CssVariables` component
- The component accepts props for e.g. `colors` and `theme` to specify what should be injected
- Props are `false` by default forcing users to make a choice

ismay:
All right, I've refactored it to what we agreed on, and also cleaned it up a bit. We could potentially combine the mapping stages to a single pass. But to me this seemed more clear to someone who's not familiar with the component's internals.

src/CssVariables/index.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
```
varl: We have moved from the directory structure of `src/{Component}/index.js` to `src/{Component}.js` in `next`. So either we merge this to `master` as is, and then refactor it immediately for `next`, or we use `src/CssVariables.js` in this PR. The discrepancy internally doesn't matter (in this case) since it is exported from `src/index.js` anyway.

Mohammer5: I'd prefer `src/CssVariables.js`, it will cause less work (no need for another PR).
Also `next` is soon to be merged anyways, so it doesn't matter if there's a bit on inconsistency internally for a short amount of time. The users of this lib won't notice it

ismay: All right, that's fine for me as well. @varl?

varl: Sure.

ismay: Done, should be good to go now 👍

src/CssVariables.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+import propTypes from 'prop-types'
+import * as theme from './theme'
```
varl: This is a contentious topic, but we have standardised on being explicit about the file extension. I don't care now, but to follow the conventions in the project it is good to be aware of.

ismay: Ah I didn't catch that. Wait, I'll add it.

ismay: Done

ismay:
@Mohammer5, you all right with the current state of the PR as well?

src/CssVariables.js
```diff
@@ -0,0 +1,70 @@
...
+ * @returns {React.Component}
+ * @example import { CssVariables } from @dhis2/ui-core
+ * @see Live demo: {@link /demo/?path=/story/cssvariables--default|Storybook}
+ */
+const CssVariables = props => {
+    const variables = Object.keys(props)
+        // Filter all props that are false
+        .filter(prop => props[prop])
+        // Map props to corresponding theme section and prefixes keys with section name
+        .map(toPrefixedThemeSection)
```
Mohammer5: This will throw an error when providing a prop that doesn't exist on `theme`.
If we want to use this "generic" method, I'd filter out everything that's not in `Object.keys(CssVariables.propTypes)`.

What I like about react is that I can spread object containing the required props on a component while all irrelevant ones are being ignored.

Mohammer5: Discussed with @ismay, it makes sense to throw an error here to detect typos

PR #427, opened 2019-09-16
refactor: kill the filled option for selectfield and inputfield

varl:
@cooper-joe This removes the Filled options for InputField and SelectField, leaving one master style. ⚔️


varl:
> 
> 
> The `InputField` includes the `Help` text, which makes sense I assume (I think this touches on @HendrikThePendric is doing in `ui-forms`, maybe he wants to say sth specific to this)
> 
> The `SelectField` does not have a `Help` text, I think we should make that consisten

Yeah, the `SelectField` did not get the same treatment on Friday, though it has now in e6f0715.

src/Help.js
```diff
@@ -12,13 +12,12 @@ import { spacers, theme } from './theme.js'
  * @example import { Help } from @dhis2/ui-core
  * @see Live demo: {@link /demo/?path=/story/help--default|Storybook}
  */
-const Help = ({ children, valid, error, warning, className, indent }) => (
+const Help = ({ children, valid, error, warning, className }) => (
```
varl: The indentation for the `Help` was mainly for the `filled`, so let's get rid of it.

varl:
@cooper-joe There might be some styling issues that we should clean up, are there any obvious ones in `InputField` and/or `SelectField` you see? Like should the `Label` have a dense option?

cooper-joe:
@varl These are the minor style changes I can see need to be made. I had a play around but wasn't sure whether they'd be applied to the `input` file or the `inputField`/`selectField` file, so I'll let you take care of that!

These changes apply to all input fields and select fields, in all states, unless otherwise noted:

- `border-radius: 3px` replaces `border-radius: 4px`
- Default border color changes to `grey500` from `theme.default` (this should not be themeable).
- disabled fields get `background-color: grey100`
- All fields get `box-shadow: inset 0 0 0 1px rgba(102,113,123,0.15), inset 0 1px 2px 0 rgba(102,113,123,0.10);`
- padding on `input` changes to `padding: 13px 0 11px 11px;`

That's everything I see, thanks!

varl:
Awesome @cooper-joe. I've made the mods and thanks for the concrete list.

varl:
> By keeping the Help out of the input fields in ui-core we keep our components "feel" much more lightweight and composable.

@HendrikThePendric The `InputField` molecule is composed from `Field`, `Input`, `Help`, and `Label`. 

All of those building blocks are also exposed through `ui-core`, so you are free to use the same atoms to create different molecules somewhere else (e.g. `ui-forms`), but a fully formed `InputField` molecule should be exposed through `ui-core` as well for simpler use cases.

varl:
> So the naming convention will be that components ending with `Field` are the fully formed ones?

Yes.

> This will trigger the migration of some ui-forms components to ui-core. 

Right. :+1: 

One more thing, I'm moving _all_ the styles in the `{Input,Select}Field` components into the lower level building blocks `{Input,Select}` to ensure that the `-Field` component is 100% composable.

PR #430, opened 2019-09-17
chore(release): cut version 4.0.0

varl:
# 4.0.0 (https://github.com/dhis2/ui/compare/v3.12.0...v4.0.0) (2019-11-20)                                                                  
                                                                                                                                             
### Bug Fixes                                                                                                                                
                                                                                                                                             
* alert-stack: revert changes so it just uses a fixed z-index (e9c043d (https://github.com/dhis2/ui/commit/e9c043d))                         
* checkbox: align focus/blur with change and remove disabled check (2c8abbe (https://github.com/dhis2/ui/commit/2c8abbe))                    
* checkbox,radio: value should be optional (#419 (https://github.com/dhis2/ui/issues/419)) (b22a9d0 (https://github.com/dhis2/ui/commit/b22a9d
0))                                                                                                                                          
* chip remove icon: fix position cross browser (ae7bc0f (https://github.com/dhis2/ui/commit/ae7bc0f))                                         * css-variables: improve robustness and developer warnings (#449 (https://github.com/dhis2/ui/issues/449)) (6422f3e (https://github.com/dhis2/
ui/commit/6422f3e))                                                                                                                           * correct handler usage for filterable menu (8e35c5d (https://github.com/dhis2/ui/commit/8e35c5d))
* file-field: consistently use FileField instead of File + fix typo (6fc35f7 (https://github.com/dhis2/ui/commit/6fc35f7))                   
* file-input: adjust component names - FormControl to Field (48bf08d (https://github.com/dhis2/ui/commit/48bf08d))                            * file-input: correct line height and add className prop (01d10e3 (https://github.com/dhis2/ui/commit/01d10e3))
* file-input: fix Edge text overflow behavior (e28b751 (https://github.com/dhis2/ui/commit/e28b751))                                         
* file-input: implement PlaceHolder as child of FileList (814f007 (https://github.com/dhis2/ui/commit/814f007))                              
* file-input: implement spacing between text and link via ::after el (4894747 (https://github.com/dhis2/ui/commit/4894747))                  
* file-input: improve children prop-type to allow multiple files (34d7d2c (https://github.com/dhis2/ui/commit/34d7d2c))                      
* file-input: introduce FileList to solve spacing issues (a870efb (https://github.com/dhis2/ui/commit/a870efb))                               * file-input: tweak CSS to fix incorrect total height (6732e36 (https://github.com/dhis2/ui/commit/6732e36))
* input: align handleFocus/handleBlur with handleChange (5afc697 (https://github.com/dhis2/ui/commit/5afc697))
* input: allow types specified in design-system (1416d5e (https://github.com/dhis2/ui/commit/1416d5e))
* input: correct prop-type for type (445e69a (https://github.com/dhis2/ui/commit/445e69a))
* input: remove duplicate story and improve "No label" (#480 (https://github.com/dhis2/ui/issues/480)) (8e1b045 (https://github.com/dhis2/ui/c
ommit/8e1b045))
* input-field: add required onChange prop to all stories (5f83475 (https://github.com/dhis2/ui/commit/5f83475))
* input-field: adjust focus props and label story (#447 (https://github.com/dhis2/ui/issues/447)) (d85eda7 (https://github.com/dhis2/ui/commit
/d85eda7))                                                                                                                                   
* layer: children must be a function and required (4d74a9e (https://github.com/dhis2/ui/commit/4d74a9e))                                     
* layer context: simplify and correct calculations (afb27bd (https://github.com/dhis2/ui/commit/afb27bd))                                    
* layer-context: add support for layers upon layers (cccd8ad (https://github.com/dhis2/ui/commit/cccd8ad))                                   
* prop-types: adjust statusPropType and use custom one for AlertBar (4eae18f (https://github.com/dhis2/ui/commit/4eae18f))                   
* radio: align handleFocus/handleBlur with handleChange (afd8bb2 (https://github.com/dhis2/ui/commit/afd8bb2))                               
* select: add missing prop types (dfe2105 (https://github.com/dhis2/ui/commit/dfe2105))                                                      
* select: add missing tab index default (e2e36dd (https://github.com/dhis2/ui/commit/e2e36dd))                                               
* select: align onfocus and onblur with other cbs (83becac (https://github.com/dhis2/ui/commit/83becac))                                     
* select: align select with callback style changes (f237557 (https://github.com/dhis2/ui/commit/f237557))                                    
* select: fix overlap on specifying input width for flex children (de02c2d (https://github.com/dhis2/ui/commit/de02c2d))                     
* select: fix positioning issues (f615d46 (https://github.com/dhis2/ui/commit/f615d46))                                                      
* select: fix select closing on chip removal (298b1dd (https://github.com/dhis2/ui/commit/298b1dd))                                           * select: make filtering case insensitive (95d6ff4 (https://github.com/dhis2/ui/commit/95d6ff4))
* split-button: remove unused styles (f66bc5a (https://github.com/dhis2/ui/commit/f66bc5a))                                                  
* switch: align handleFocus/handleBlur with handleChange (dfc4efc (https://github.com/dhis2/ui/commit/dfc4efc))
* switch: change input type to checkbox because switch is not valid (b42258c (https://github.com/dhis2/ui/commit/b42258c))
* tab-bar: enforce that children are instances of the Tab component (c8d6c9f (https://github.com/dhis2/ui/commit/c8d6c9f))
* text-area: align handleFocus/handleBlur with handleChange (4a73b78 (https://github.com/dhis2/ui/commit/4a73b78))
* adjust menu position after option selection (a31adcf (https://github.com/dhis2/ui/commit/a31adcf))
* export Label in index.js (69498d0 (https://github.com/dhis2/ui/commit/69498d0))
* introduce layer-context to deal with overlapping elements + story (e4f7169 (https://github.com/dhis2/ui/commit/e4f7169))
* make name prop optional instead of required (0a58802 (https://github.com/dhis2/ui/commit/0a58802))
* use layer-context in alert-stack and drop-menu (abc171a (https://github.com/dhis2/ui/commit/abc171a))
* text-area: only set height when value changes (35ba2f7 (https://github.com/dhis2/ui/commit/35ba2f7))
* toggle-group: remove .isRequired from value prop (deed091 (https://github.com/dhis2/ui/commit/deed091))
* toggles: remove position relative from toggles (04953cf (https://github.com/dhis2/ui/commit/04953cf))
        
### chore                              
                                                   
* help: remove indent props and examples from stories (#471 (https://github.com/dhis2/ui/issues/471)) (bba5ba9 (https://github.com/dhis2/ui/co
mmit/bba5ba9))
* upgrade react to support hooks (27c852f (https://github.com/dhis2/ui/commit/27c852f))
                                                
### Code Refactoring                                              
        
* checkbox: align 'on' function handlers (7cc2f6b (https://github.com/dhis2/ui/commit/7cc2f6b))
* fileinput: align 'on' function handlers (0a94f88 (https://github.com/dhis2/ui/commit/0a94f88))
* formcontrol: rename to field (#402 (https://github.com/dhis2/ui/issues/402)) (84edf35 (https://github.com/dhis2/ui/commit/84edf35))
* input: align 'on' function handlers (8aeaf64 (https://github.com/dhis2/ui/commit/8aeaf64))
* input,textarea: remove option to pass in number for width (4a4a2e7 (https://github.com/dhis2/ui/commit/4a4a2e7))
* modal: align modal with component structure (1cafa84 (https://github.com/dhis2/ui/commit/1cafa84))
* modal: remove open prop (e616bd9 (https://github.com/dhis2/ui/commit/e616bd9))
* modal: remove unnecessary aside element wrapper (ace92d1 (https://github.com/dhis2/ui/commit/ace92d1))
* radio: align 'on' function handlers (f37501c (https://github.com/dhis2/ui/commit/f37501c))
* select: update component to new design (a21e426 (https://github.com/dhis2/ui/commit/a21e426))
* select,input: drop -field suffix on components (#389 (https://github.com/dhis2/ui/issues/389)) (ed56586 (https://github.com/dhis2/ui/commit/
ed56586))
* splitbutton: separate component from buttonstrip (#396 (https://github.com/dhis2/ui/issues/396)) (b33d4f4 (https://github.com/dhis2/ui/commi
t/b33d4f4))                                                      
* implement the Radio component consistently with other inputs (9387656 (https://github.com/dhis2/ui/commit/9387656))
* kill the filled option for selectfield and inputfield (#427 (https://github.com/dhis2/ui/issues/427)) (73ab1be (https://github.com/dhis2/ui/
commit/73ab1be))                                    
* toggle input components (e5bfb32 (https://github.com/dhis2/ui/commit/e5bfb32))
* switch: align 'on' function handlers (36707da (https://github.com/dhis2/ui/commit/36707da))
* tabbar: align component structure (427d7cb (https://github.com/dhis2/ui/commit/427d7cb))
* textarea: align 'on' function handlers (267b088 (https://github.com/dhis2/ui/commit/267b088))
* toggle-group: split into composable components (#388 (https://github.com/dhis2/ui/issues/388)) (a5d17e1 (https://github.com/dhis2/ui/commit/
a5d17e1))    
                                                    


### Documentation

* introduce user docs (#397 (https://github.com/dhis2/ui/issues/397)) (9f7cd2b (https://github.com/dhis2/ui/commit/9f7cd2b))

### Features

* field: add className prop (32d3638 (https://github.com/dhis2/ui/commit/32d3638))
* file-field: introduce FileField component (023d0db (https://github.com/dhis2/ui/commit/023d0db))
* file-input: add disabled prop and button sizes (83fce79 (https://github.com/dhis2/ui/commit/83fce79))
* input atom: add ellipsis to overflowing text (0cc8155 (https://github.com/dhis2/ui/commit/0cc8155))
* label: add className prop (4cb579c (https://github.com/dhis2/ui/commit/4cb579c))
* modal: add className prop (663d717 (https://github.com/dhis2/ui/commit/663d717))
* node: add className prop to node (82e921c (https://github.com/dhis2/ui/commit/82e921c))
* screencover: add transparent prop (70ade82 (https://github.com/dhis2/ui/commit/70ade82))
* select: add input max height prop (47a6516 (https://github.com/dhis2/ui/commit/47a6516))
* select: add inputWidth property to control width (81d67f1 (https://github.com/dhis2/ui/commit/81d67f1))
* status-icon: add info and defaultTo props (b972bce (https://github.com/dhis2/ui/commit/b972bce))
* tab components: add className props (8512eec (https://github.com/dhis2/ui/commit/8512eec))
* textarea: introduce TextArea and TextAreaField with docs and stories (#456 (https://github.com/dhis2/ui/issues/456)) (3d85a6c (https://github.com/dhis2/ui/commit/3d85a6c))
* theme: expose spacers, elevations and layers constants (d82f1f4 (https://github.com/dhis2/ui/commit/d82f1f4))

### BREAKING CHANGES

* modal: Removed the aside element wrapper
* select,input: InputField renamed to Input
  SelectField renamed to Select
* chore: trigger netlify
* input,textarea: Input and TextArea now requires the input width
  override to be passed in as a string.
* Now requires React >= 16.8.
* fileinput: onChange is called with two parameters:
  First parameter an object with properties for checked, name, and
  value.
  The second is the React event object.
* textarea: onChange is called with two parameters:
  First parameter an object with properties for checked, name, and
  value.
  The second is the React event object.
* input: onChange is called with two parameters:
  First parameter an object with properties for checked, name, and
  value.
  The second is the React event object.
* checkbox: onChange is called with two parameters:
  First parameter an object with properties for checked, name, and
  value.
  The second is the React event object.
* radio: onChange is called with two parameters:
  First parameter an object with properties for checked, name, and
  value.
  The second is the React event object.
* switch: onChange is called with two parameters:
  First parameter an object with properties for checked, name, and
  value.
  The second is the React event object.
* tabbar: This replaces the need to nest a TabBar in a ScrollBar,
  and instead allows TabBar.propTypes.scrollable to toggle if a TabBar
  is scrollable.
* tabbar: This deprecates external use of ScrollBar.
* modal: This deprecated use of Modal.Actions. Instead, use
  ModalActions.
* modal: This deprecated use of Modal.Content. Instead, use
  ModalContent.
* modal: This deprecated use of Modal.Title. Instead, use
  ModalTitle.
* tab-bar: TabBar will now only accept Tab instances
* modal: Removing the "open" prop from the Modal.
* select: SelectField deprecated in favor of SingleSelectField,
  see the docs for the API.
* select: The new Select components do not accept a standard HTML option
  element, instead they require use of SingleSelectOption, MultiSelectOption,
  or a component with a matching API.
* select: The Select's onChange no longer returns a standard Event.target,
  instead returns the selected Object with properties: label and
  string, or an array of those for the MultiSelects
* select: The Selects all require a selected prop, instead of a value.
  Where selected is an object with a label and a value string, or an array of
  those for the MultiSelect.
* select: Since the Selects do not use a native input element, the name
  prop has been dropped from the Select components.
* The icon and required props have been removed from
  the Checkbox and Switch. For Switch, the label prop has become
  required.
* RadioGroup has a new implementation and drops the
  following props: label, inline, options, and helpText.
* The icon and required props have been removed from the Radio component and the value and label props have become required.
* help: Removes indent prop from Help. Use className to
  override if necessary.
* Remove the Filled field style for SelectField and
  InputField.
* MenuItem value prop is more strict, requires String
  instead of any.
* formcontrol: Renames FormControl to Field.
* splitbutton: remove the compact prop from buttonstrip
* toggle-group: Renames ToggleGroup to FieldSet to match intended
  usage.
* select: Select deprecated in favor of the SingleSelect, refer
  to the docs for the API.


Mohammer5:
It looks like not all PRs have been merged with `Squash and merge`.
Not a biggie, because they'll be `Squashed and merge`d into master, but it makes it harder to read what happened @HendrikThePendric @varl @ismay 

If all PRs had been merged with 'Squash and merge`, this PR could even be merged with `Rebase and merge`, then we'd have all the feature commits on master as well

varl:
I tried rebasing `next` on `master`, which was a colossal fustercluck so merged `master` into `next` and did the cleanup myself.

We do not do "rebase and merge" (it's disabled almost everywhere, and should be disabled _everywhere_) so everything will be squashed and merged with a commit message that describes _all_ the breaking changes and additions, which is going to be quite involved in writing.

varl:
#### ⚠️ Early heads up

Once all the existing branches have been merged to `next` and this branch is due for release, I am going to rebase it on `master`, fix any commit messages that _are not_ conventional commit messages, and then force-push `next` branch.

Mohammer5:
Let's add "Fix storybook in Edge" to the todo list

src/DropMenu.js
```diff
@@ -58,15 +87,23 @@ class DropMenu extends PureComponent {
...
 DropMenu.propTypes = {
     className: propTypes.string,
-    /** The component to use as the dropdown component */
     component: propTypes.element,
-    /** Function to trigger when click happens outside of the DOM element */
     onClose: propTypes.func,
-    /** Decides if the menu should call the onClose function or not */
     stayOpen: propTypes.bool,
-    /** DOM node to position itself against */
+    /**  */
```
ismay: Can be removed right?

varl: Yeah.

src/Help.js
```diff
@@ -0,0 +1,69 @@
...
+ * @prop {boolean} [valid] - `valid`, `warning`, and `error`, are mutually exclusive
+ * @prop {boolean} [warning]
+ * @prop {boolean} [error]
+ */
+Help.propTypes = {
+    className: propTypes.string,
+    children: propTypes.string.isRequired,
+    error: statusPropType,
+    valid: statusPropType,
+    warning: statusPropType,
```
ismay: Possible breaking change: `<Help>` no longer accepts an `indent` prop

src/MenuItem.js
```diff
@@ -96,12 +106,26 @@ const MenuItem = ({
...
+ * @prop {boolean} [dense]
+ * @prop {boolean} [active]
+ * @prop {boolean} [disabled]
+ */
 MenuItem.propTypes = {
-    label: propTypes.oneOfType([propTypes.string, propTypes.object]).isRequired,
+    label: propTypes.oneOfType([propTypes.string, propTypes.node]).isRequired,
 
-    value: propTypes.any,
+    value: propTypes.string,
```
ismay: Possible breaking change: this prop has become more strict.

src/Help.js
HendrikThePendric: Agreed

src/MenuItem.js
HendrikThePendric: agreed

src/RadioGroup.js
```diff
@@ -0,0 +1,45 @@
...
+ * @prop {boolean} [error]
+ *
+ * @prop {boolean} [dense]
+ */
+RadioGroup.propTypes = {
+    ...ToggleGroup.propTypes,
+    children: propTypes.arrayOf(propTypes.instanceOfComponent(Radio))
+        .isRequired,
+    value: propTypes.string,
+}
```
HendrikThePendric: BREAKING CHANGE: RadioGroup has been implemented differently. It now doesn't have:
- `label`
- `inline`
- `options`
- `helpText`

src/Tab.js
```diff
@@ -109,13 +119,23 @@ const Tab = ({ icon, onClick, selected, disabled, children }) => (
...
+ * @prop {function} [onClick]
+ * @prop {boolean} [selected]
+ * @prop {boolean} [disabled]
+ * @prop {Node} [children]
+ * @prop {string} [className]
+ */
 Tab.propTypes = {
-    icon: PropTypes.element,
-    onClick: PropTypes.func,
-    addTabRef: PropTypes.func,
```
ismay: Possible breaking change, addTabRef was dropped.

src/Menu.js
```diff
@@ -0,0 +1,46 @@
...
+ * @typedef {Object} PropTypes
+ * @static
+ *
+ * @prop {*} [children]
+ * @prop {string} [className]
+ */
+Menu.propTypes = {
+    children: MenuList.propTypes.children,
+    className: propTypes.string,
+}
```
HendrikThePendric: BREAKING CHANGE: Menu was implemented differently and has lost virtually all of its props

src/TableCellHead.js
```diff
@@ -46,15 +46,13 @@ export const TableCellHead = ({
...
  * @static
- * @prop {string} [label]
  * @prop {string} [colSpan]
  * @prop {string} [rowSpan]
  * @prop {bool} [dense]
  * @prop {Node} [children]
  * @prop {string} [className]
  */
 TableCellHead.propTypes = {
-    label: propTypes.string,
```
ismay: Possible breaking change: TableCellHead no longer accepts a label

HendrikThePendric:
I've compared the props in the storybook-prop-table@master with the-docs@next. Got down to RadioGroup and left a few comments. I'm not sure how we should describe the breaking changes in Menu, because I didn't take time to look into it. There a comment to flag up the breaking changes though...

src/ButtonStrip.js
```diff
@@ -0,0 +1,58 @@
...
+ * @prop {string} [className]
+ * @prop {Array.<Button>} [children]
+ * @prop {boolean} [middle] - The props `middle`, and `end` are
+ * mutually exlusive
+ * @prop {boolean} [end]
+ */
+ButtonStrip.propTypes = {
+    className: propTypes.string,
+    children: propTypes.arrayOf(propTypes.instanceOfComponent(Button)),
+    middle: alignmentPropType,
```
ismay: So maybe technically not a breaking change since `start` has become the default, but the `start` prop has been dropped.

src/DropMenu.js
ismay: https://github.com/dhis2/ui-core/pull/552

src/TableCellHead.js
varl: No, it was an unused prop so ESLint threw a warning about it.

src/Menu.js
varl: No difference from `master`: https://github.com/dhis2/ui-core/blob/master/src/Menu/index.js

src/Tab.js
varl: It was unused so ESLint complained.

varl:
I have gone over all the comments noted, and scanned the APIs to make them more consistent. I found that in some places we say that onChange functions are required, and in other places they are optional.

I have applied a strict definition that `isRequired` indicates that the component basically will crash if the property is not passed in, for example the `name` property is assumed to exist in the code. If it is not supplied, the component will throw an error, so that's a required prop.

For the `onChange` we always check if we have an onChange function before calling it, so that's not a strictly required property and as such I have loosened it so that all `on` handlers are optional. This makes the components easier to use in a prototype situation.

Naturally, the component wont' work _properly_ without an `onChange` handler, but I think that should be documented in the user docs instead of enforcing it with props.

varl:
One thing that @edoardo found is that when wrapping a component in another, the `instanceOf` propType checks throw errors, for example when using a Redux connected component. The same problem applies in any situation where a component is wrapped though so even this throws a propType validation error:

```
const MyButton = () => <Button>Foobar</Button>

<ButtonStrip>
    <MyButton />
    <Button>Save</Button>
    <Button>Save</Button>
    <Button>Save</Button>
</ButtonStrip>
```

```
Warning: Failed prop type: Invalid prop `children[0]` supplied to `ButtonStrip`, expected an instance of `Button` but found an instance of `function MyButton() {
  return _ref;
}`.
```

We should remove the `instanceOf` prop type checks from all the externally available components.

Mohammer5:
@varl I see the point, I'm not sure it makes sense to remove these from all components. The `Table` component (`TableRow`, etc) might need that (and I assume the tables themselves will be wrapped, not their rows or cells, although it could make sense in rare cases, I'd still prefer to make sure the majority of the cases work seamlessly)

ismay:
> I have gone over all the comments noted, and scanned the APIs to make them more consistent. I found that in some places we say that onChange functions are required, and in other places they are optional.
>
> I have applied a strict definition that isRequired indicates that the component will crash if the property is not passed in, for example the name property is assumed to exist in the code. If it is not supplied, the component will throw an error, so that's a required prop.
>
>For the onChange we always check if we have an onChange function before calling it, so that's not a strictly required property and as such I have loosened it so that all on handlers are optional. This makes the components easier to use in a prototype situation.
>
>Naturally, the component won't work properly without an onChange handler, but I think that should be documented in the user docs instead of enforcing it with props.

That makes sense, seems like a developer friendly approach 👍

About the `name` prop. I've noticed in the scheduler app that the ui-core components require a name prop, where I don't believe I used it myself in the app. Also, including or omitting the name prop seems like a valid way to include or omit fields for natively submitting forms. By requiring it we're making that choice for the user. Would there be a way to loosen that requirement, so that it becomes optional again?

HendrikThePendric:
> the ui-core components require a name prop
Well spotted @ismay! I agree name should be optional everywhere and currently there seems to no pattern in where this has been implemented as required vs optional. As far as I can tell, there is no form element that requires this attribute. I'll prepare a PR.

varl:
I think we should settle the event handler pattern before releasing.

HendrikThePendric:
> I think we should settle the event handler pattern before releasing.

I'm on board with the current implementation following our discussion on Slack. How about you @ismay @Mohammer5?

Mohammer5:
I'm ok with the `on[CBName](obj, evt)` style. @varl suggested to use that for **all** callbacks, also onFocus/onBlur/etc for consistency reasons, is everyone on board with that as well?

ismay:
> I'm ok with the on[CBName](obj, evt) style. @varl suggested to use that for all callbacks, also onFocus/onBlur/etc for consistency reasons, is everyone on board with that as well?

I still have concerns but I've stated those yesterday, so I'll defer to what the majority prefers.

> We definitely should handle all event handlers consistently. And currently this is not the case, we only implement the (obj, event) pattern in the handleChange methods. So this will need to be amended.

So what do we do if there's no changed data. Do we pass just some basic info, like the following:

```
// The Chip's onRemove cb does not have any direct data associated with the event for example
onRemove({ name: 'chipname' }, e)
```

Same goes for the Button's onClick, etc. There are event handlers that do not have data associated with the event. We should decide how to handle that.

I do think that passing an object as the first arg introduces an extra thing that the user has to know. Not only is the cb style different, but they also have to know if the value they want lives under selected, value, checked, etc. I understand that going with an object provides extra flexibility for future additions, but it does introduce an extra thing you have to know.

varl:
> Same goes for the Button's onClick, etc. There are event handlers that do not have data associated with the event. We should decide how to handle that.

We would call the handler with what makes sense for the component, we do have a `value` prop for the Button as well, so we'd call it back with that.

> I do think that passing an object as the first arg introduces an extra thing that the user has to know. Not only is the cb style different, but they also have to know if the value they want lives under selected, value, checked, etc. I understand that going with an object provides extra flexibility for future additions, but it does introduce an extra thing you have to know.

Agreed. I don't think there is away around the "extra thing to know" problem; as such I don't think that is a deal-breaker.

One _fundamental flaw_ that has been discovered is that by changing the callback style we break interoperability with standard `onEvent` handlers on components. The most consistent solution is to state that if you pass an `onEvent` handler to a UI component it is going to be called with the signature `object, event`.

#### Use a CustomEvent with the `details` property
  Another way is to wrap the generated event in a new CustomEvent, and use the `detail` property to provide an object:

```js
const event = new CustomEvent("uiEvent", {
  detail: {
    name: this.props.name,
    value: this.props.value,
  }
})
```

Which would be accessed through `event.detail`. There is still "another thing to learn", and for e.g. the **Select** the expected `event.target.value` is not going to work, even though it works in other components like the **Input**.

The user would have to know that she should look under `event.detail` to find the "real" values. :thinking: 

At that point it is more consistent, and more discoverable, to pass an object as the first parameter that has the actual values that the user _wants_.

[1] https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent

ismay:
> Use a CustomEvent with the details property

Yeah I was thinking about that as well. I thought maybe we could create a custom synthetic event, but that's possible apparently. Using native events instead will probably be a lot more confusing, so I agree with your sugggestion 👍 I've refactored the select PR to match it.

varl:
The `Backdrop` between the Select menu and the rest of the page is intentional. It is not _optimal_ but it works around using an event listener attached to the document as well as provide a consistent UX.

Again, it's not optimal but _it is_ good enough.

Before we had the backdrop that stopped the event from propagating, the `onClick` handler on the document would fire before the click on the `x` in a Chip, causing the menu to close. There is no way to stop a native event from within React's synthetic event world, and they generally recommend _not_ using `addEventListener` as it breaks the React event model by bypassing it.

If we use the `addEventListener` method, we can maintain the more traditional UX of being able to click on an element outside the list, but to what advantage?

This way the user does not have to find a "safe" area to click when she wants to close the menu, but can just click _anywhere_ on the page. I'd consider that a positive. Especially as a misguided click in our (or more likely, external) apps can lead to destructive actions.

Even if we figure out another way (maybe move the entire Select, menu-and-all, to a portal and have it over the Backdrop?) this would not be a breaking change and can be released as an incremental improvement. Most important is that it _works_.

varl:
Ref. the **LayerContext**, I've added a rudimentary feature file here: https://github.com/dhis2/ui-core/pull/593 (no tests).

varl:
Let's fly. 🚀

PR #435, opened 2019-09-18
ci: lint source on ci as well

ismay:
One thing I noticed whilst going through the propType warnings that I didn't address yet:

It seems Input and Select have a `focus` prop, but the components are reporting receiving a function instead of a bool. In the Field story I'm not seeing a function being passed in for `focus` to the rendered Input component explicitly, so I'm not sure what's going on here exactly.

I looked at this with Hendrik, but we didn't have enough time to really figure out what was going wrong. It does seem like there's some funky stuff going on here. Input has an initialFocus prop as well for example (and uses it), and InputField defines one, but doesn't use it anywhere.

Wwe should enable react/no-unused-prop-types, to catch these kinds of things:

Screenshot 2019-09-19 at 17 40 46

Some false positives, but might be useful

### DIFF
--- a/src/Button.js
+++ b/src/Button.js
@@ -97,7 +97,7 @@ Button.defaultProps = {
  * state
  */
 Button.propTypes = {
-    children: propTypes.string,
+    children: propTypes.node,
     onClick: propTypes.func,
 
     className: propTypes.string,
--- a/src/Button.js
+++ b/src/Button.js
@@ -76,8 +76,7 @@ Button.defaultProps = {
  * @typedef {Object} PropTypes
  * @static
  *
- * @prop {string} [children] The string to show as the label text for a
- * button
+ * @prop {Node} [children] The children to render in the button
  * @prop {function} [onClick] The click handler
  *
  * @prop {boolean} [small] - `small` and `large` are mutually exclusive
--- a/src/InputField.js
+++ b/src/InputField.js
@@ -23,6 +23,8 @@ class InputField extends React.Component {
         const {
             className,
             onChange,
+            onFocus,
+            onBlur,
             type,
             dense,
             required,
@@ -54,8 +56,8 @@ class InputField extends React.Component {
 
                 <Input
                     focus={focus}
-                    onFocus={this.onFocus}
-                    onBlur={this.onBlur}
+                    onFocus={onFocus}
+                    onBlur={onBlur}
                     onChange={onChange}
                     name={name}
                     type={type}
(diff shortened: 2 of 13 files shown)


src/Button.js
```diff
@@ -97,7 +97,7 @@ Button.defaultProps = {
  * state
  */
 Button.propTypes = {
-    children: propTypes.string,
+    children: propTypes.node,
```
ismay: For `<DropdownButton>`, which renders more than just strings in the button (also an icon).

src/InputField.js
```diff
@@ -54,8 +56,8 @@ class InputField extends React.Component {
 
                 <Input
                     focus={focus}
-                    onFocus={this.onFocus}
-                    onBlur={this.onBlur}
+                    onFocus={onFocus}
```
ismay: Let me know if this was intentional, it seemed like a typo to me. In storybook before this change, it was throwing a TypeError (`this.props.onFocus is not a function`). I assume the intention here was to allow the user of this component to specify a callback for the onFocus/onBlur events.

src/MenuList.js
```diff
@@ -38,7 +38,7 @@ const MenuList = ({ children, className }) => (
  */
 MenuList.propTypes = {
     className: propTypes.string,
-    children: propTypes.element.isRequired,
+    children: propTypes.node.isRequired,
```
ismay: MenuList accepts an array (or at least it receives an array in the storybook demo), which element does not allow (it expects a single react element), but node does.

stories/CssReset.stories.js
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { CssReset } from '../src'
 
+// eslint-disable-next-line react/prop-types
```
ismay: I'm ignoring prop-types for storybook inline components, seems excessive to check those as well.

src/InputField.js
ismay: Ah, so maybe the intention here was to refactor this in the same manner as the Checkbox? With class methods for onFocus and onBlur. I guess it got lost somewhere. So we'll have to implement this here just like with Checkbox?

varl: Good catch, @ismay. The `InputField` used to have class methods for `onFocus` and `onBlur`. Since a `InputField` should only ever be a composition of smaller components, the class methods were pushed down to `Input`: https://github.com/dhis2/ui-core/blob/next/src/Input.js#L94-L102

I forgot to change `this.onFocus` and `this.onBlur` to `props.onFocus` and `props.onBlur` for the `InputField` component. I think your change here is correct!

varl:
Let's merge this and look at it holistically in `next`.

PR #436, opened 2019-09-18
build: fix docs build

ismay:
As a follow up to yesterday's changes on 2b764abce54713b52043105a762b3d7e80d47efa, I assume it was also the intention to remove this line? This line still being there doesn't crash the build, but does crash the storybook app with `Uncaught ReferenceError: withPropsTable is not defined`

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -6,7 +6,6 @@ import { FrameStylesWrapper } from './frame-styles-decorator'
 import 'typeface-roboto'
 
 addDecorator(jsxDecorator)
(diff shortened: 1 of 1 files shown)


varl:
Yeah, I missed removing that line, but it's done in 7c19c9c58ef17e2cd45395edd054fcc5b2798d29.

varl:
We can close this.

ismay:
Haha, ah you beat me to it. I checked out next before that. Well ok, fixed 👍

PR #439, opened 2019-09-19
test: add test framework

ismay:
Somehow snapshots aren't cooperating nicely with enzyme-to-json in combination with styled-jsx. Styled-jsx is rendering a lot of noise in our snapshots.

There's this article: https://kevinjalbert.com/jest-snapshots-reducing-styled-jsx-noise and the styled-jsx/babel-test plugin, but I've not yet been able to find something that works well.

Alongside that I think some test would be more at home in the browser than in jsdom. I'd like to be able to do basic tests, like verify that the CssVariables are inserted correctly (which jsdom can't verify atm.), or make sure that a disabled Checkbox does not call onChange when clicked (which is not possible in enzyme as you have to pass the change event directly), etc. So we'd need to see what we want to test with jest exactly, alongside fixing the snapshot issue.


varl:
Interesting. Perhaps the case for Cypress/Selenium here is stronger than I thought. :thinking:

ismay:
> Interesting. Perhaps the case for Cypress/Selenium here is stronger than I thought

Yeah it would allow us to test things that aren't testable at the moment with jest/jsdom. Plus the tests would be more realistic.

Though I can still imagine tests where the overhead of cypress wouldn't be worth it. Whatever we decide though, I do think that for a library as central to our ecosystem as this, we need good test coverage.

ismay:
Should we just get started with cypress as that seems like a decent option, and then if the need arises add jest when needed?

varl:
Closing this as we merged Cypress setup in #476. Let's see if we need Jest in the future as noted in @ismay's  [comment](https://github.com/dhis2/ui-core/pull/439#issuecomment-534012949).

PR #438, opened 2019-09-19
refactor: align file-input with others

HendrikThePendric:
- FileInput no longer includes the label
- FileInputField includes label, help and error
- Other inputs have a tabIndex property so we added that to FileInput(Field) too. However, since the visible element on the form is a button, the Button component needed to get a tabIndex
- We are not longer exporting name-spaced components, FileList, FileListItem and FileListPlaceholder are exported separately
- Stories were updated
- Small tweak to ESLint config so it would allow me to commit

PR #446, opened 2019-09-23
ci: add no unused props rule

ismay:
We have a couple of components with props that aren't used. I've added a linting rule to catch this in the future and fixed some of the linting errors. However, for some of the linting errors I'm not sure if the props can be safely removed, or if this is a mistake and we actually want to use these props:

Screenshot 2019-09-23 at 11 15 30

So @Mohammer5, @HendrikThePendric can the above props be safely removed?

### DIFF
--- a/src/CssVariables.js
+++ b/src/CssVariables.js
@@ -61,13 +61,11 @@ CssVariables.defaultProps = {
  * @prop {boolean} [elevations]
  */
 CssVariables.propTypes = {
-    /* eslint-disable react/no-unused-prop-types */
     colors: propTypes.bool,
     theme: propTypes.bool,
     layers: propTypes.bool,
     spacers: propTypes.bool,
     elevations: propTypes.bool,
-    /* eslint-enable */
 }
 
 export { CssVariables }
--- a/src/CssVariables.js
+++ b/src/CssVariables.js
@@ -61,11 +61,13 @@ CssVariables.defaultProps = {
  * @prop {boolean} [elevations]
  */
 CssVariables.propTypes = {
+    /* eslint-disable react/no-unused-prop-types */
     colors: propTypes.bool,
     theme: propTypes.bool,
     layers: propTypes.bool,
     spacers: propTypes.bool,
     elevations: propTypes.bool,
+    /* eslint-enable */
 }
 
 export { CssVariables }
--- a/src/FileInput.js
+++ b/src/FileInput.js
@@ -43,7 +43,6 @@ class FileInput extends PureComponent {
             large,
             disabled,
             tabIndex,
-            name,
         } = this.props
 
         return (
--- a/src/FileInput.js
+++ b/src/FileInput.js
@@ -43,6 +43,7 @@ class FileInput extends PureComponent {
             large,
             disabled,
             tabIndex,
+            name,
         } = this.props
 
         return (
@@ -100,6 +101,7 @@ class FileInput extends PureComponent {
  * @typedef {Object} PropTypes
  * @static
  *
+ * @prop {string} name
  * @prop {function} onChange
  * @prop {string} [buttonLabel]
  * @prop {string} [className]
--- a/src/FileInput.js
+++ b/src/FileInput.js
@@ -97,7 +97,6 @@ class FileInput extends PureComponent {
  * @typedef {Object} PropTypes
  * @static
  *
- * @prop {string} name
  * @prop {function} onChange
  * @prop {string} [buttonLabel]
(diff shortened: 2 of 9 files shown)


src/CssVariables.js
```diff
@@ -60,11 +60,13 @@ CssVariables.defaultProps = {
  * @prop {boolean} [elevations]
  */
 CssVariables.propTypes = {
+    /* eslint-disable react/no-unused-prop-types */
```
ismay: Disabled the rule here because these props are used, but just not directly.

HendrikThePendric: Suppose we'd change the implementation of `CssVariables` a bit:
```javascript
const CssVariables =({ colors, theme, layers, spacers, elevations }) => {
    const variables = Object.keys({ colors, theme, layers, spacers, elevations })
    // ....
```
This is quite verbose, but it does prevent potential problems: in the current implementation, whenever a prop is added of a non-existent theme key, the component throws an error, i.e.
```html
<CssVariables colors theme layers spacers elevations someThingElse />
```
This will result in the following error:
`Uncaught TypeError: Cannot convert undefined or null to object`

ismay: Yeah we decided to leave that behaviour in to fail clearly when there are spelling errors in the props. Otherwise you won't be explicitly notified (there'll be a prop-type warning in development, but that's it). We could do it this way as well, I'm fine with both approaches.

I don't think it's a problem to have to disable a rule by the way, as long as it doesn't happen 
that often. It cannot catch certain prop usage, but weighed against the benefit of being warned about unused props is worth it for me. On 124 files having to add the disable to 1 file isn't too bad in my opinion.

HendrikThePendric: > Yeah we decided to leave that behaviour in to fail clearly

IMO this isn't that clear... The error is quite cryptic. I would have preferred to have seen `PropTypes.exact` being used to give a clear message, and the code not failing.

HendrikThePendric: BTW: sorry, I should've gotten involved in the `CssVariables` PR instead of suggesting changes here.

ismay: > IMO this isn't that clear..

I agree, I also don't think it's that clear. We discussed it here: https://github.com/dhis2/ui-core/pull/421#discussion_r324646233

If we only want warnings in development we could go with your suggestion. And as you say, we could use propTypes.exact to catch non-existent props. That seems good to me.

It will however fail silently in production then. If we also want to fail clearly in production I think we should throw a clear error on non-existent props.

ismay: I'd prefer using your suggestion (propTypes.exact and destructuring) and failing clearly in production as well. It doesn't add a lot of extra code or complicated logic and I like that better than failing silently. At that point you can argue that the prop-types don't add much, but that's another discussion, I'm all for optimizing that as well if there's a nicer solution.

Mohammer5: How about explicitly throwing an error? Then we'd have control over the error message as well

HendrikThePendric: I'll work on a PR. Let's discuss further there.

ismay:
Rebased on next and removed the CssVariables eslint disable comments. So the only thing that we need to resolve now is the above question in the top comment.

varl:
@ismay The `initialFocus` prop on `SelectField` and `InputField` should be passed down to the `Select` and `Input` components. This may have been resolved in another PR.

The `addTabRef` is not used anywhere in any stories, so looks safe to remove.

The `FileInput` props look like they belong to `FileInputField`. I think this was resolved in another PR as well, but I haven't verified (cc: @HendrikThePendric).

HendrikThePendric:
> The FileInput props look like they belong to FileInputField. I think this was resolved in another PR as well, but I haven't verified

Correct, this was merged into `next` https://github.com/dhis2/ui-core/pull/438

src/FileInput.js
```diff
@@ -97,7 +97,6 @@ class FileInput extends PureComponent {
  * @typedef {Object} PropTypes
  * @static
  *
- * @prop {string} name
```
varl: In `Input.js` we use `name` and `id`. Not sure if they apply to a `FileInput` but if they do the consistency would be nice:

https://github.com/dhis2/ui-core/blob/next/src/Input.js#L141-L142

ismay: All right, I'm not sure what's applicable here. @Mohammer5 @HendrikThePendric?

HendrikThePendric: yes, plz add them.

PR #449, opened 2019-09-23
fix(css-variables): improve robustness and developer warnings

HendrikThePendric:
The thinking behind this is:
1. Because we now loop through the object keys of `allowedKeys` we prevent nasty errors
1. Since the previous change cause things to fail silently (i.e. developer adds prop `shape` which isn't in the `allowedKeys` so it's simply ignored), I added some props-validation that would let the developer know he has added a not-allowed prop.

Currently this PR has draft status because it is dependent on https://github.com/dhis2/prop-types/pull/46. Once that's been merged, we still need to update the version of `@dhis2/prop-types` here....

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,9 +3,16 @@ const { config } = require('@dhis2/cli-style')
 module.exports = {
     parser: 'babel-eslint',
     extends: ['eslint:recommended', 'plugin:react/recommended', config.eslint],
+    env: {
+        es6: true,
+    },
     settings: {
         react: {
             version: '16.3',
         },
+        propWrapperFunctions: [
+            'forbidUnknowProps',
+            { property: 'forbidUnknowProps', object: 'propTypes' },
+        ],
     },
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,9 +7,5 @@ module.exports = {
         react: {
             version: '16.3',
         },
-        propWrapperFunctions: [
-            'whitelistProps',
-            { property: 'whitelistProps', object: 'propTypes' },
-        ],
     },
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,8 +8,8 @@ module.exports = {
             version: '16.3',
         },
         propWrapperFunctions: [
-            'forbidUnknowProps',
-            { property: 'forbidUnknownProps', object: 'propTypes' },
+            'whitelistProps',
+            { property: 'whitelistProps', object: 'propTypes' },
         ],
     },
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,7 @@ module.exports = {
         },
         propWrapperFunctions: [
             'forbidUnknowProps',
-            { property: 'forbidUnknowProps', object: 'propTypes' },
+            { property: 'forbidUnknownProps', object: 'propTypes' },
         ],
     },
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,9 +3,6 @@ const { config } = require('@dhis2/cli-style')
 module.exports = {
     parser: 'babel-eslint',
     extends: ['eslint:recommended', 'plugin:react/recommended', config.eslint],
-    env: {
-        es6: true,
-    },
     settings: {
         react: {
             version: '16.3',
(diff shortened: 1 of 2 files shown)


src/CssVariables.js
```diff
@@ -59,12 +60,12 @@ CssVariables.defaultProps = {
  * @prop {boolean} [spacers]
  * @prop {boolean} [elevations]
  */
-CssVariables.propTypes = {
+CssVariables.propTypes = propTypes.forbidUnknowProps({
```
varl: Should be: `forbidUnknownProps`

varl:
Why are we being clever about this at all?

We can pick the exact props we want from both the `theme` and the `props`, and if someone wants to pass in `foobar` we just ignore it.

varl:
OK I found the discussion that sparked this here: https://github.com/dhis2/ui-core/pull/446#discussion_r327044059

HendrikThePendric:
> OK I found the discussion that sparked this

Yeah you beat me to it. Damn. I would personally be fine just letting it fail silently. If we chose to do so, we could also consider just closing: https://github.com/dhis2/prop-types/pull/46

varl:
If we assume that:

- We close dhis2/prop-types#46
- We skip the `forbidUnknownProps`

Then:

> * But I would like to keep working with a notion of `allowedProps`, because I didn't like how that was implemented before

How would you change this PR to align it with that remark?

HendrikThePendric:
Like this: https://github.com/dhis2/ui-core/pull/449/commits/38af6c54a154cbb1bab5ae0178fca27c25c2a106

.eslintrc.js
```diff
@@ -7,5 +7,9 @@ module.exports = {
         react: {
             version: '16.3',
         },
+        propWrapperFunctions: [
+            'whitelistProps',
+            { property: 'whitelistProps', object: 'propTypes' },
+        ],
```
varl: Why is this necessary and will it be required everywhere we want to use the function?

varl:
> > How would you change this PR to align it with that remark?
> 
> Like this: [38af6c5](https://github.com/dhis2/ui-core/commit/38af6c54a154cbb1bab5ae0178fca27c25c2a106)

All right.

varl:
Fat fingers and touchpad.. Re-opened.

ismay:
Since the prop-types are a bit more involved than just using propTypes.exact (it seems that's a bug), why not just keep CssVariables as it is currently?

It's simple and it'll fail if you use it wrong. Seems fine to me.

varl:
> Since the prop-types are a bit more involved than just using propTypes.exact (it seems that's a bug), why not just keep CssVariables as it is currently?
> 
> It's simple and it'll fail if you use it wrong. Seems fine to me.

If we change it according to 38af6c5, then we do not need to ignore the ESLint rule `no-unused-props` here: https://github.com/dhis2/ui-core/pull/446/files/48a487a3f07835cf011ebe00126be7548c852593#diff-723be8d36b324ae9efbe3f705b379236R63.

.eslintrc.js
varl: Based on the last commit this should also be removed.

ismay:
> > Since the prop-types are a bit more involved than just using propTypes.exact (it seems that's a bug), why not just keep CssVariables as it is currently?
> > It's simple and it'll fail if you use it wrong. Seems fine to me.
> 
> If we change it according to [@HendrikThePendric's commit](https://github.com/dhis2/ui-core/pull/449/commits/2de77900a09f9ed7f7e2e847298da9929d1a0ebc#diff-723be8d36b324ae9efbe3f705b379236R26), then we do not need to [ignore the ESLint rule `no-unused-props`](https://github.com/dhis2/ui-core/pull/446/files/48a487a3f07835cf011ebe00126be7548c852593#diff-723be8d36b324ae9efbe3f705b379236R63).

That's true. I'll defer to what you and @HendrikThePendric prefer.

varl:
I agree with the `allowedProps` change just to be able lint it properly without ignores.

HendrikThePendric:
OK, just a short summary:
- We're keeping the change in `CssVariables` that introduced `allowedProps`, because it removes the need for ignoring the lint rule and prevent the function from throwing errors on non-existent theme variables.
- We're dropping the prop-types validation (`whitelistProps`) part, because the solution is a bit too involved for such a small problem.
- So, that also means https://github.com/dhis2/prop-types/pull/46 is now closed

PR #450, opened 2019-09-23
chore(prop-types): switched to our own @dhis2/prop-types

HendrikThePendric:
What I've done is:
1. Removed all named imports from `@dhis2/prop-types`
1. Prefixed calls to these named functions with `propTypes.`
1. Switch to importing from `@dhis2/prop-types` which includes all of our custom methods on the default exported object

What remains is:
1. What to do with the `package.json`? Currently we have `prop-types` as a peer dependency and `@dhis2/prop-types` as a dependency. I don't think this is really correct, but not sure what would be best....

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,12 +45,11 @@
     "typeface-roboto": "^0.0.75"
   },
   "peerDependencies": {
-    "prop-types": "^15",
     "react": "^16.3",
     "react-dom": "^16.3"
   },
   "dependencies": {
-    "@dhis2/prop-types": "^1.1.0",
+    "@dhis2/prop-types": "^1.1.1",
     "classnames": "^2.2.6",
     "styled-jsx": "^3.2.2"
   },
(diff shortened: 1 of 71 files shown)


HendrikThePendric:
PS: merge-base here is the `next` branch. This isn't really a breaking change, but since it touches many files I'd prefer to introduce it in `next`

varl:
>     1. What to do with the `package.json`? Currently we have `prop-types` as a peer dependency and `@dhis2/prop-types` as a dependency. I don't think this is really correct, but not sure what would be best....

I think the best way is that `@dhis2/prop-types` has a straight dependency to `prop-types` and `@dhis2/ui-{core,widgets}` has a dependency to `@dhis2/prop-types`.

In the platform we can mark `@dhis2/prop-types` as external and guarantee we only end up with a single version of it.

HendrikThePendric:
> I think the best way is that @dhis2/prop-types has a straight dependency to prop-types and @dhis2/ui-{core,widgets} has a dependency to @dhis2/prop-types.

OK, have implemented it that way now, see https://github.com/dhis2/prop-types/pull/50 and https://github.com/dhis2/ui-core/pull/450/commits/12dfd559af590cefa154d5acd63674c24acb8707

This is now ready to be merged in. Marked the PR as "ready for review".

PR #456, opened 2019-09-26
feat(textarea): introduce TextArea and TextAreaField with docs and stories

HendrikThePendric:
I think this has been implemented according to the specs, and in-line with the other components. There is only one small point to note:
In the design system the `TextArea` has a `width` property, which has been implemented on the `TextArea` itself. However, on the `TextAreaField` I haven't included the width property, because this is not supported by `Field`. If I were to add it as a prop on the `TextAreaField` and then just forward it to `TextArea`, we'd get a narrow textarea within a full-width container. Not sure what we want to do with all of that.

### DIFF
--- a/src/Tabs/detectScrollbarSize.js
+++ b/src/Tabs/detectScrollbarSize.js
@@ -1,4 +1,4 @@
-let horizontalScrollbarHeight
+let scrollbarSize
 const className = '__vertical-scrollbar-height-test__'
 const styles = `
     .${className} {
@@ -13,9 +13,11 @@ const styles = `
     }
 `
 
-export function detectHorizontalScrollbarHeight() {
-    if (horizontalScrollbarHeight) {
-        return horizontalScrollbarHeight
+// This function assumes horizontal and vertical scrollbars
+// are always the same for a given browser/OS combo
+export function detectScrollbarSize() {
+    if (scrollbarSize) {
+        return scrollbarSize
     }
 
     const style = document.createElement('style')
@@ -27,10 +29,10 @@ export function detectHorizontalScrollbarHeight() {
     document.body.appendChild(style)
     document.body.appendChild(el)
 
-    horizontalScrollbarHeight = el.offsetHeight - el.clientHeight
+    scrollbarSize = el.offsetHeight - el.clientHeight
 
     document.body.removeChild(style)
     document.body.removeChild(el)
 
-    return horizontalScrollbarHeight
+    return scrollbarSize
 }
--- a/src/Tabs/detectScrollbarSize.js
+++ b/src/Tabs/detectScrollbarSize.js
@@ -1,6 +1,6 @@
 let scrollbarSize
 const className = '__vertical-scrollbar-height-test__'
-const styles = `
+const baseCSS = `
     .${className} {
         position: absolute;
         top: -9999px;
@@ -8,6 +8,9 @@ const styles = `
         height: 100px;
         overflow-x: scroll;
     }
+`
+const hiddenScrollbarCSS = `
+    ${baseCSS}
     .${className}::-webkit-scrollbar {
         display: none;
     }
@@ -15,13 +18,13 @@ const styles = `
 
 // This function assumes horizontal and vertical scrollbars
 // are always the same for a given browser/OS combo
-export function detectScrollbarSize() {
+export function detectScrollbarSize(tryHiding) {
     if (scrollbarSize) {
         return scrollbarSize
     }
 
     const style = document.createElement('style')
-    style.innerHTML = styles
+    style.innerHTML = tryHiding ? hiddenScrollbarCSS : baseCSS
 
     const el = document.createElement('div')
     el.classList.add(className)
--- a/src/TextArea.js
+++ b/src/TextArea.js
@@ -110,7 +110,8 @@ export class TextArea extends PureComponent {
                     placeholder={placeholder}
                     ref={this.textareaRef}
                     value={value}
-                    disabled={disabled || readOnly}
+                    disabled={disabled}
+                    readOnly={readOnly}
                     tabIndex={tabIndex}
                     onFocus={onFocus}
                     onBlur={onBlur}
--- a/src/TextArea.js
+++ b/src/TextArea.js
@@ -103,7 +103,7 @@ export class TextArea extends PureComponent {
         const textareaWidth = typeof width === 'number' ? `${width}px` : width
 
         return (
-            <div className={cx('container', className)}>
+            <div className={cx('textarea', className)}>
                 <textarea
                     id={name}
                     name={name}
@@ -116,7 +116,7 @@ export class TextArea extends PureComponent {
                     onBlur={onBlur}
                     onChange={onChange}
                     rows={rows}
-                    className={cx('textarea', className, {
+                    className={cx({
                         dense,
                         disabled,
                         error,
@@ -137,7 +137,7 @@ export class TextArea extends PureComponent {
 
                 <style jsx>{styles}</style>
                 <style jsx>{`
-                    .textarea {
+                    textarea {
                         width: ${textareaWidth};
                         height: ${height};
                         resize: ${resize};
--- a/src/TextArea.js
+++ b/src/TextArea.js
@@ -185,6 +185,36 @@ TextArea.defaultProps = {
     onFocus: () => {},
 }
 
+/**
+ * @typedef {Object} PropTypes
+ * @static
+ *
+ * @prop {string} name
+ * @prop {function} onChange
+ * @prop {function} [onBlur=() => {}]
+ * @prop {function} [onFocus=() => {}]
+ * @prop {string} [className]
+ * @prop {string} [placeholder]
+ * @prop {string} [value]
+ * @prop {string} [tabIndex]
+ *
+ * @prop {boolean} [disabled]
+ * @prop {boolean} [dense] - Compact mode
+ * @prop {boolean} [initialFocus]
+ *
+ * @prop {boolean} [valid] - `valid`, `warning`, `error`, and `loading`
+ * are mutually exclusive
+ * @prop {boolean} [warning]
+ * @prop {boolean} [error]
+ * @prop {boolean} [loading]
+ *
+ * @prop {boolean} [autoGrow]
+ * @prop {boolean} [readOnly]
+ * @prop {string} [resize=vertical] - rezise attribute on the textarea,
+ * one of `none`, `both`, `horizontal`, `vertical`
+ * @prop {number} [rows=4]
+ */
+
 TextArea.propTypes = {
     name: propTypes.string.isRequired,
 
--- a/src/TextArea.js
+++ b/src/TextArea.js
@@ -1,144 +1,49 @@
 import propTypes from '@dhis2/prop-types'
 import React, { PureComponent } from 'react'
-import css from 'styled-jsx/css'
 import cx from 'classnames'
 
 import { statusPropType } from './common-prop-types.js'
-import { theme, colors } from './theme.js'
 import { StatusIconNoDefault } from './icons/Status.js'
 import { detectScrollbarSize } from './Tabs/detectScrollbarSize.js'
 
-const styles = css`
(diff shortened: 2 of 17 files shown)


src/Tabs/detectScrollbarSize.js
```diff
@@ -1,36 +1,41 @@
...
         display: none;
     }
 `
 
-export function detectHorizontalScrollbarHeight() {
-    if (horizontalScrollbarHeight) {
-        return horizontalScrollbarHeight
+// This function assumes horizontal and vertical scrollbars
+// are always the same for a given browser/OS combo
+export function detectScrollbarSize(tryHiding) {
```
Mohammer5: I think it would be nice to have a jsdoc block explaining the `tryHiding` param, because the name doesn't tell me anything about it's purpose, I had to read the code (which wasn't complicated, but will be an extra effort, everything I'm scanning this file)

Also using verbs for non-function vars is not so good, I think. I assumed it's a function, so maybe something like `withHiddenScrollbar` is more appropriate

src/TextArea.js
```diff
@@ -0,0 +1,246 @@
...
+        this.state = {
+            hasScrollbar: false,
+            height: 'auto',
+        }
+
+        this.scrollbarSize = detectScrollbarSize()
+        this.textareaDimensions = { width: 0, height: 0 }
+        this.userHasResized = false
+
+        this.setTextareaDimensions = this.setTextareaDimensions.bind(this)
```
Mohammer5: Why not just declaring the method like this:

```js
setTextareaDimensions = () => { ... }
```

Then you wouldn't have to bind functions in the constructor anymore and when reading through the method I know immediately that it's bound to the classes' instance.

src/TextArea.js
```diff
@@ -0,0 +1,246 @@
...
+
+import { styles } from './TextArea/styles.js'
+
+export class TextArea extends PureComponent {
+    constructor(props) {
+        super(props)
+
+        this.textareaRef = React.createRef()
+
+        this.state = {
```
Mohammer5: I think all the properties could be declared as properties instead of in the constructor (they don't depend on props).
With the other change in the comment below, we could get rid of the constructor.

src/TextArea.js
```diff
@@ -0,0 +1,246 @@
...
+
+                <style jsx>{styles}</style>
+                <style jsx>{`
+                    .textarea {
+                        width: ${width};
+                        height: ${this.state.height};
+                        resize: ${resize};
+                    }
+                    .status-icon :global(svg) {
+                        right: ${this.getIconRightPosition()};
```
Mohammer5: Although this is not a complicated calculation, I think this value should be put into the state instead of calculating it on every render

cooper-joe:
Looking good to me. The only issue I see Viktor has already mentioned, the icon overlapping the text:
image


I thought this might be an issue that needs a little 'massaging' to work 😅 Let me know how it goes. We do need the icon somehow (a red border is not an accessible error signal), and I think putting the icon outside the textfield would cause even more layout headaches. But let me know if we need to discuss other options.

src/TextArea.js
```diff
@@ -0,0 +1,246 @@
...
+        )
+    }
+}
+
+TextArea.defaultProps = {
+    rows: 4,
+    width: '100%',
+    resize: 'vertical',
+    onBlur: () => {},
+    onFocus: () => {},
```
Mohammer5: Is this necessary? I think it should be ok to just leave this out, as it's just passed down to the `textarea`, which has its own default prop for `onBlur` and `onFocus`

HendrikThePendric:
> We can apply the same technique by resorting to the resize CSS3 rule to resize the parent div

This suggestion is very promising @varl 👍. Quite early on in development I decided that "we need to apply the border to the TextArea because this needs to be resizable". Later on I did happen to read that it is possible to make most elements resizable (I wasn't aware of that until then), but it never occurred to me that this offered opportunities to implement the textarea in a way that is consistent with the other inputs. The current CSS works, but is a bit hacky. Following your suggestion I can implement cleaner CSS and be more consistent with the other inputs. I'm not 100% sure yet how things will work out with the scrollbar and the icon position. I expect it will actually make things easier there and solve the overlapping text and icon position issue too. Will work on implementing it like this.

HendrikThePendric:
@varl @cooper-joe @Mohammer5 could you please also reply to my comment about the `width` property? I'm not quite sure how I should deal with that...
> In the design system the TextArea has a width property, which has been implemented on the TextArea itself. However, on the TextAreaField I haven't included the width property, because this is not supported by Field. If I were to add it as a prop on the TextAreaField and then just forward it to TextArea, we'd get a narrow textarea within a full-width container

cooper-joe:
> @varl @cooper-joe @Mohammer5 could you please also reply to my comment about the `width` property? I'm not quite sure how I should deal with that...
> 
> > In the design system the TextArea has a width property, which has been implemented on the TextArea itself. However, on the TextAreaField I haven't included the width property, because this is not supported by Field. If I were to add it as a prop on the TextAreaField and then just forward it to TextArea, we'd get a narrow textarea within a full-width container

From my perspective: if I design a `textAreaField` with a defined width, that is because I also want to define the width of the label and help/error text. If I define a width only for `textArea` then I wouldn't expect that width to apply to the label/help/error text. In theory I would want to be able to define these independently. (`textArea` would inherit from `textAreaField` if not set). 

I'm not sure how that fits in with the way the composition works currently. This concept also applies to `input`, `select` and any other composed `Field` component.

Here is an illustrated example:
textArea width

HendrikThePendric:
Thanks for providing illustrated examples @cooper-joe, I'll first provide a response to the examples, and then will discuss some general implications:
1. `textArea width: 280px` This is basically the behaviour we'd see if I just introduce the `width` property on the `TextArea` component. I think I would also let `TextAreaField` accept `width` and then forward it to the `TextArea`. 

2. `textAreaField width: 280px`: This touches on some other components too, i.e. the `Label` and the `Help`. If we want to go this way, we should implement it on all input components.

3. `textArea width: 200px + textAreaField width: 280px`: This is problematic because we'd have to expose an additional prop on the `TextAreaField`, i.e. `width` and `textAreaWidth`. I would say, if a developer needs something like this, he/she should just build a custom component from the atoms we expose.

So, assuming the last example is not relevant, we only need to deal with point 1 and 2.

I can implement 1, but would like to know why this is needed. I can imagine it is to do with readability / preventing lines from becoming too long?

Reg. 2: I would argue that we actually don't want to implement control on the width of a `Field` (and its children) at all because there are better ways of dealing with these types of situations. Most likely, when we want to reduce the width of an entire field, this is because:
- We think the entire form is too wide
- We want to display some other input next to the field with the reduced width.
For both of these cases, I think it would be much better if the developer implements some form of a layout in the form, i.e. using `rows` and `columns`. By not introducing `width` on the `Field` component, we prevent dodgy form styles.

In summary I think this:
- Introducing `width` on the `TextAreaField` and `TextArea` could be justified to reduce line length, but the width should only be applied to the actual `TextArea` and not the label and help fields
- I don't think we should introduce width anywhere else because we should prefer our devs to implement layout grids instead.

HendrikThePendric:
> @HendrikThePendric and I talked about the icon that's rendered in the input a while back.

@ismay: while I think this is a useful discussion to have, please see the comment @varl left https://github.com/dhis2/ui-core/pull/456#pullrequestreview-294144007. I think for now, we should focus on consistency across our components. And then we can discuss an alternative approach at a later stage and in another branch....

cooper-joe:
> In summary I think this:
> 
>  Introducing `width` on the `TextAreaField` and `TextArea` could be justified to reduce line length, but the width should only be applied to the actual `TextArea` and not the label and help fields
> 
>  I don't think we should introduce width anywhere else because we should prefer our devs to implement layout grids instead.

I can live with this. I agree with your points that defining the width of the whole `Field` component would be best done with a proper layout.

> I can implement 1, but would like to know why this is needed. I can imagine it is to do with readability / preventing lines from becoming too long?

Controlling the input width is important because it lets you send a subtle hint to the user that 'this is the content this input expects'. A 750px wide input implies the user should write a whole sentence, a 100px input implies a single word is appropriate. It is very use to be able to control the `input` width independently of the whole `Field` (perhaps the label is long, but the expected input is short, so you'd want a wide `Field` and a short `input`, for example. That would be possible with your example, just passing width to the `input`.).

ismay:
> @ismay: while I think this is a useful discussion to have, please see the comment @varl left #456 (review). I think for now, we should focus on consistency across our components. And then we can discuss an alternative approach at a later stage and in another branch....

@HendrikThePendric and I talked a bit on slack about the above, conclusion: first implement in line with the existing components, talk about revising the way the icons and borders are handled in our components at a later stage. Which sounds good to me 👍

HendrikThePendric:
> Controlling the input width is important because it lets you send a subtle hint to the user that 'this is the content this input expects'. A 750px wide input implies the user should write a whole sentence, a 100px input implies a single word is appropriate.

Aha, that hadn't occurred to me, but does make a lot of sense. So I suppose the `Input` and `InputField` should also get a `width`, same as the `TextArea`.
@cooper-joe: would you mind creating an issue for this? (not sure in Jira or Github)

varl:
@HendrikThePendric I would suggest:

- `Field` => do nothing
- `TextAreaField` (and all other `*-Fields`): `props.inputWidth` & `props.width`
- `TextArea`/`Input`/`Select`/`Radio`/etc: `props.width`

HendrikThePendric:
I don't fully understand your comment @varl:

> Field => do nothing

👍 

> TextAreaField (and all other *-Fields): props.inputWidth & props.width

Not sure if I get this one... Add `inputWidth` to `TextAreaField` and use it as value for the `width` prop of the `TextArea`?

> TextArea/Input/Select/Radio/etc: props.width

👍 

Also, you have referenced this in #460, but I don't immediately see the connection.... If the `width` only applies to the input itself, then it doesn't have affect the label, right?

varl:
> I don't fully understand your comment @varl:
> 
> > TextAreaField (and all other *-Fields): props.inputWidth & props.width
> 
> Not sure if I get this one... Add `inputWidth` to `TextAreaField` and use it as value for the `width` prop of the `TextArea`?

Yes. The reason behind the name `inputWidth` is that it's generic enough to be used for all the form controls (they are all "user inputs").

> Also, you have referenced this in #460, but I don't immediately see the connection.... If the `width` only applies to the input itself, then it doesn't have affect the label, right?

Yes, but if we have an inline option for the label, then we will also need a `labelWidth` property, so the full set will be:

- `Field`: none
- `Input`: `props.width`
- `Label`: `props.width`
- `InputField`:
  - `props.width`: the entire width
  - `props.labelWidth`: width of the label (pass through to `Label.propTypes.width`)
  - `props.inputWidth`: width of the input component (pass through to `Input.propTypes.width`)

You can replace `Input` and `InputField` with `Radio`, `Select`, etc. and the exact same props can be used.

Does that make my previous comment a bit clearer?

HendrikThePendric:
I've updated this PR as follows:
- All comments related to parts of the code have been addressed
- The UI problems with the status icons have not been addressed directly, instead I've used the a new UI which was an outcome of discussions https://github.com/dhis2/ui-core/pull/465#issuecomment-536917144 and https://github.com/dhis2/ui-core/issues/464
- The TextArea component has not been aligned with the Input. Instead, the Input should be aligned with the TextArea (also outcomes of https://github.com/dhis2/ui-core/pull/465#issuecomment-536917144 and https://github.com/dhis2/ui-core/issues/464)

So, I think this is ready for another round of reviews
@varl @Mohammer5 @ismay @cooper-joe

src/TextArea/styles.js
```diff
@@ -0,0 +1,64 @@
+import css from 'styled-jsx/css'
+import { colors, theme, spacers } from '../theme.js'
+
+export const styles = css`
+    .textarea {
```
Mohammer5: Can be moved into the next `textarea` rule.

HendrikThePendric: I don't think so, the container is a `div.textarea` and it has a `textarea` element as a child.

Issue #457, opened 2019-09-27
Should the Radio Group be usable without a Label?


amcgee:
Right now, when using `RadioGroup` without a `label` prop, the component creates an empty `<span>` which messes up the alignment of the radio buttons.  Below is a screenshot for context.  I was able to work around this by setting `display: none` on the span in CSS, but maybe the `label` prop shouldn't be required and the `<span>` should only render if it's specified?  Or maybe I'm breaking some design rules I shouldn't be here @cooper-joe?

I think in this case if we want to require a label, we should be able to render the label to the left of the radio group, rather than above it.

<img width="1232" alt="Screenshot 2019-09-27 10 56 40" src="https://user-images.githubusercontent.com/947888/65757408-56e73080-e117-11e9-9400-30f8bbb83529.png">

And this is what it looks like without the CSS hack (with no label specified):

<img width="565" alt="Screenshot 2019-09-27 11 11 37" src="https://user-images.githubusercontent.com/947888/65757527-9e6dbc80-e117-11e9-835f-2bfd0010479a.png">

cooper-joe:
Interesting use case, thanks for raising this @amcgee.

I can see here how a label-less radioGroup is useful. Normally my advice would be: "include a label, otherwise how will the user know what the radio buttons apply to?". However, in this use case it is somewhat self-evident that the radio buttons apply to the type of execution. For beginner users it may still be useful to have a label of 'Type' or something similar. 

I do see how `labelPosition` with options `top` and `left` could be useful, especially in these compact situations. I would imagine this would be relatively painless to implement, so I think it may be worth including to boost the flexibility of these field components. What do you think @dhis2/team-apps?

In conclusion:
- I am in favour of adding a labelPosition property
- I do agree that if a label is not included the `span` should be omitted, although labels are highly recommended.

Mohammer5:
We're in the process of changing `ui-core` right now (see the `next` branch).
We're extracting input components from field components (field = [label +] input [+ help] [+error]),
so the component you need could be composed easily with the atoms `ui-core` will provide.

It shouldn't be too much of a problem then to create a `RadioGroupField` that can handle multiple label positions, which will require custom css for the molecule (and the current rule is: no css for anything but atoms).

@varl here I am again, proposing to get rid of this rule because molecules will need "glue" styles until we have figured out a proper way (maybe layout components, maybe not).

If we stick to that rule, it will still be fairly easy to create this component in apps/libs because all relevant atoms will be exposed.

[...]

PR #458, opened 2019-09-28
refactor(radio): refactor the Radio, RadioGroup and RadioGroupField incl. new SVGs

varl:
Relates to #457.

This refactor will include the `Radio` (atom) component and some form of `*-Field` (molecule) component that can be used to construct radio groups.

ToDos:
- [x] Implement `Radio`
    - [x] Allow prop-type `node` in `label`
    - [x] Add `dense` prop and style
- [x] Implement `RadioGroup`
- [x] Implement `RadioGroupField`
- [x] Tweak styling of components within `RadioGroupField`
- [x] Rework `Checkbox` and `Radio` SVG (see https://github.com/dhis2/ui-core/issues/515)


HendrikThePendric:
When I implemented the `RadioGroupField` (see https://github.com/dhis2/ui-core/pull/458/commits/1ce48e54d7c50393cf31570c7679b09ae37c0343) I noticed some imperfections reg. the styles. @cooper-joe could you review the following points:
- Spacing with label: The distance between the groups seems to be equal to the spacing between the legend/label and the checkboxes. I don't think this looks right.
- Dense mode: The radios themselves support dense mode. However, the legend/label and the help/error texts don't. Is that OK? Or should we at least add a dense mode to the Legend?
- Disabled mode: The Help texts change color for error/warning/valid state, but not for disabled. Is that OK?

The deploy preview link is up-to-date so you can use that to review the points above.

HendrikThePendric:
Something else I noticed just now, but haven't been able to find an elegant solution for yet:
- The width of the `label` that wraps the radio-input is 100%. You could argue this is a feature, but quite often this is confusing.
- The most logical way to control this, in theory, would be to apply the `display: flex;` property to the `FieldSet` component....
- But as @Mohammer5 pointed out at some point, `fieldset`s can't be flex containers. So we'd have to add a div inside or around the fieldset to get this behavior to work...
- Adding `display: flex;` to the `Field` component fixes things too, but that doesn't make as much sense conceptually, and also breaks some other Fields. I suppose we could add it as a prop  to the `Field` component?

Any other suggestions on how to fix this? And/or opinions on whether this is a bug or a feature?
@ismay @cooper-joe @Mohammer5 @varl

HendrikThePendric:
**UPDATE:** probably just applying `display: inline-block` to the `FieldSet` as I did in https://github.com/dhis2/ui-core/pull/458/commits/2e815b572d61153a94ab6afea0ff54f3ebd9b2bc is the most appropriate solution.

cooper-joe:
> When I implemented the `RadioGroupField` (see [1ce48e5](https://github.com/dhis2/ui-core/commit/1ce48e54d7c50393cf31570c7679b09ae37c0343)) I noticed some imperfections reg. the styles. @cooper-joe could you review the following points:
> 
>     * Spacing with label: The distance between the groups seems to be equal to the spacing between the legend/label and the checkboxes. I don't think this looks right.

Spacing for groups added to [updated spec](https://sketch.cloud/s/DwkDk/a/Ak9J0M)

>     * Dense mode: The radios themselves support dense mode. However, the legend/label and the help/error texts don't. Is that OK? Or should we at least add a dense mode to the Legend?

Dense mode labels/help text are the same as regular mode. Dense mode controls are actually intended more for usage in UI components where help/label/states won't be used, so this isn't a issue (right now).

>     * Disabled mode: The Help texts change color for error/warning/valid state, but not for disabled. Is that OK?

Yes, that's fine.

Mohammer5:
> The width of the label that wraps the radio-input is 100%. You could argue this is a feature, but quite often this is confusing.

I think this is fine, they're behaving like block elements so they're not inline.
The fieldset with it's initial display value (which is `block` if I'm not mistaken) adjusts to it's parent's `width`, which is fine as well. If the UI requires different behavior, it's up to the app.
So I think there's no need for this:

> UPDATE: probably just applying display: inline-block to the FieldSet as I did in 2e815b5 is the most appropriate solution.

src/FieldSet.js
```diff
@@ -16,6 +16,7 @@ const FieldSet = ({ className, children }) => (
                 border: none;
                 margin: 0;
                 padding: 0;
+                display: inline-block;
```
varl: A fieldset should be a block element.

src/RadioGroupField.js
```diff
@@ -0,0 +1,86 @@
...
+
+            {helpText && <Help>{helpText}</Help>}
+
+            {validationText && (
+                <Help error={error} warning={warning} valid={valid}>
+                    {validationText}
+                </Help>
+            )}
+        </FieldSet>
+    </Field>
```
varl: Nice!

src/FieldSet.js
varl: c7e7afa

varl:
@HendrikThePendric I don't see this point addressed in this PR:

>I don't think we need to merge branch #514 that @ismay worked on. Instead I'll just cherry-pick e98f744

ismay:
@varl I assume that that's something that will be addressed in #516. Or is that not what you meant?

varl:
> 
> 
> @varl I assume that that's something that will be addressed in #516. Or is that not what you meant?

Yeah, I mixed them up.

varl:
This looks good!

Please **squash locally**[1] and add a conventional commit message before merging.

Remember that **if this is a breaking change**[2] you need to add a footer to the commit message with the words `BREAKING CHANGE:` and a list of the changes that are breaking.

[1] There are many articles on `git rebase -i`, here's one: https://stackabuse.com/git-squash-multiple-commits-in-to-one-commit/
[2] https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer

HendrikThePendric:
@varl I've tried following your instruction reg. squashing commits etc. With @ismay's help I think I've achieved the correct result. I believe it was mentioned somewhere that it would be good to keep the original commit messages in the squashed commit message, so I've done that too.

Could you let me know if this is OK, and if so, approve this PR?

PR #459, opened 2019-09-29
fix: don't use dynamic re-exports

amcgee:
Using dynamic re-exports prevents static bundlers (like rollup) from supporting CJS named imports - if we include the named exports explicitly instead, static named imports work in CJS modules.  

See [this app-platform issue](https://github.com/dhis2/app-platform/issues/69), [the app-platform rollup config](https://github.com/dhis2/app-platform/blob/master/cli/config/rollup.config.js#L77), and [the underlying rollup explanation](https://github.com/rollup/rollup-plugin-commonjs/issues/211#issuecomment-337897124) (which explains why this affects rollup but not webpack).

We should figure out how to support ES imports in rollup (maybe stop using rollup at all?) but for now this will make it a little less confusing when trying to do CJS named imports from `ui-core`

### DIFF
--- a/src/Table.js
+++ b/src/Table.js
@@ -1,8 +1,0 @@
-export { TableBody } from './Table/TableBody.js'
-export { TableCell } from './Table/TableCell.js'
-export { TableCellHead } from './Table/TableCellHead.js'
-export { Table } from './Table/Table.js'
-export { TableFoot } from './Table/TableFoot.js'
-export { TableHead } from './Table/TableHead.js'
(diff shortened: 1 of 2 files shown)


varl:
> Webpack gets around the need for `namedExports` by keeping everything as CommonJS modules and implementing Node's `require` system in the browser

I've referenced that in the past after I ran into `ui-core` being "un-bundleable" with Rollup. I prefer staying close to the ESM standard (unless it's a pure node environment), so this fix is inline with that.

Issue #464, opened 2019-09-30
New status icon position


HendrikThePendric:
### Context
Both @ismay and I encountered some problems when implementing Select and TextArea. These problems stem from the fact that in the current design we have visually tried to show the status icon inside of the input. In the actual implementation we have wrapped both the input and the status-icon in a div on which we've defined the border-color etc. Specifically this is causing the following problems:
- For these Input elements we have to keep track of the input elements focus-state in as a component state property. This is adding a little unnecessary overhead to the component.
- In the TextArea we are encountering some problems with the position of the scrollbar, see [this fiddle for illustration](https://jsfiddle.net/hendrikdegraaf/3njh2rpg/8/). This can only be solved by additional logic in the component, making this less lightweight.

### Solution
So, @ismay's suggestion to solve this was to re-implement the design and come up with a solution that doesn't try to make the container div look like an input, but just style the _actual input_ instead.

We first discussed things with @varl and he was on board with trying to solve things this way. We then discussed things with @cooper-joe and presented him with a revised design. Of course he came up with a much better solution, which looks like this:
![new_error_icon styles](https://user-images.githubusercontent.com/353236/65881546-df720500-e393-11e9-8aaf-f31d05481475.png)
This solution doesn't just help the current implementation of the Input/TextArea/Select, but we will also be able to use this for components such as `RadioGroupField` etc.

**UPDATE**
We decided to go with the icon-next-to-input after all (https://github.com/dhis2/ui-core/pull/465#issuecomment-536912958) https://github.com/dhis2/ui-core/pull/465#issuecomment-536917144
![image](https://user-images.githubusercontent.com/353236/65952439-2c141980-e442-11e9-8d1e-357728316ece.png)

### Implementation
To complete this transition the following tasks need to be completed:
- [ ] ~~Add an `icon` prop to the `Help` component~~
- [ ] Update the implementation of `Input` and `InputField`
- [ ] Update the WIP implementation of `Select` and `SelectField`
- [ ] Update the WIP implementation of `TextArea` and `TextAreaField`

varl:
Important information in:

- https://github.com/dhis2/ui-core/pull/465#issuecomment-536912958 (@HendrikThePendric)
- https://github.com/dhis2/ui-core/pull/465#issuecomment-536917144 (@varl)

varl:
@cooper-joe What do you think of these PRs?

- https://github.com/dhis2/ui-core/pull/472
[...]

PR #465, opened 2019-09-30
Feat/icon in help

HendrikThePendric:
I managed to get this to work with the icons being 4px greater than the line-height, and the lines not getting an increased height. If you have any suggestion to improve the CSS and achieve the same goal, I'd be in favour of that.

See: https://github.com/dhis2/ui-core/issues/464 for some more context of the current PR

### DIFF
--- a/src/Help.js
+++ b/src/Help.js
@@ -4,6 +4,7 @@ import React from 'react'
 
 import { statusPropType } from './common-prop-types.js'
 import { spacers, theme } from './theme.js'
+import { StatusIconNoDefault } from './icons/Status'
 
 /**
  * @module
@@ -12,7 +13,7 @@ import { spacers, theme } from './theme.js'
  * @example import { Help } from @dhis2/ui-core
  * @see Live demo: {@link /demo/?path=/story/help--default|Storybook}
  */
-const Help = ({ children, valid, error, warning, icon, className }) => (
+const Help = ({ children, valid, error, warning, loading, className }) => (
     <p
         className={cx(className, {
             valid,
@@ -20,7 +21,16 @@ const Help = ({ children, valid, error, warning, icon, className }) => (
             warning,
         })}
     >
-        {icon && <span className="help-icon">{icon}</span>}
+        {(valid || error || warning || loading) && (
+            <span className="help-icon">
+                <StatusIconNoDefault
+                    error={error}
+                    valid={valid}
+                    warning={warning}
+                    loading={loading}
+                />
+            </span>
+        )}
 
         {children}
 
@@ -35,6 +45,7 @@ const Help = ({ children, valid, error, warning, icon, className }) => (
                 color: ${theme.default};
                 cursor: help;
                 display: flex;
+                align-items: center;
             }
 
             .valid {
@@ -57,6 +68,7 @@ const Help = ({ children, valid, error, warning, icon, className }) => (
                 fill: inherit;
                 pointer-events: none;
             }
+
             .help-icon :global(svg) {
                 width: 18px;
                 height: 18px;
@@ -71,10 +83,10 @@ const Help = ({ children, valid, error, warning, icon, className }) => (
  * @static
  * @prop {string} children
  * @prop {string} [className]
- * @prop {boolean} [valid] - `valid`, `warning`, and `error`, are mutually exclusive
+ * @prop {boolean} [valid] - `valid`, `warning`, `error`, and `loading` are mutually exclusive
  * @prop {boolean} [warning]
  * @prop {boolean} [error]
- * @prop {Element} [icon]
+ * @prop {boolean} [loading]
  */
 Help.propTypes = {
     className: propTypes.string,
@@ -82,7 +94,7 @@ Help.propTypes = {
     error: statusPropType,
     valid: statusPropType,
     warning: statusPropType,
-    icon: propTypes.element,
+    loading: statusPropType,
 }
 
 export { Help }
--- a/src/Help.js
+++ b/src/Help.js
@@ -12,7 +12,7 @@ import { spacers, theme } from './theme.js'
  * @example import { Help } from @dhis2/ui-core
  * @see Live demo: {@link /demo/?path=/story/help--default|Storybook}
  */
-const Help = ({ children, valid, error, warning, className }) => (
+const Help = ({ children, valid, error, warning, icon, className }) => (
     <p
         className={cx(className, {
             valid,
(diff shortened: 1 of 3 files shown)


HendrikThePendric:
As an after-thought: perhaps when reviewing this PR we should focus one the component API only and not pay too much attention to the implementation of the styles....

Currently we don't have any detailed design specs so trying to update the CSS might be a bit pointless. If @cooper-joe provides some design specs after his holiday, we can fix the styling in another PR.

On the other hand, if the component API changes are approved, we can merge this into next, and we can move-ahead with the Select/Input/TextArea changes that depend on the current PR.

varl:
While we might need the `props.icon` for some things, I do not think it should be used for the status icons.

This should be enough to get the "Warning help text":

```jsx
<Help warning>I warn thee</Help
```

HendrikThePendric:
> While we might need the props.icon for some things, I do not think it should be used for the status icons.

@varl this PR is triggered by https://github.com/dhis2/ui-core/issues/464. Could you review that and have another look at this PR afterwards. BTW: my apologies, I should have referenced that issue in the PR description. Will update....

varl:
> @varl this PR is triggered by #464. Could you review that and have another look at this PR afterwards. BTW: my apologies, I should have referenced that issue in the PR description. Will update....

I have read #464 and I stand by the comments I made about the API.

```jsx
<Help warning>I have an Warning icon and a warning status</Help>
```

should result in this render:

image

varl:
If we drop the `props.icon`, and the `Help` component imports the status icons, then the change in `Help` could be done without an API change at all.

varl:
@HendrikThePendric and I discussed a bit on Slack and agreed that the status icon is always displayed based on the status property, this avoids changing the API.

We can always add the `props.icon` when we need it in the future as a minor bump.

HendrikThePendric:
I've updated this PR according to @varl's recommendations. @ismay also did some good suggestions to simplify the CSS somewhat ([see this fiddle](https://jsfiddle.net/hmb23our/)), but I decided to leave that for now. We'll optimise that in a separate PR once the design specs are available.

cooper-joe:
> One drawback is that when something is validating async using the loading status, then all of the elements underneath are going to jump down while the loading status is displayed, and jump back up when the async validation is done.
> 
> Generally I am against elements "jumping around" for no reason as it is jarring and disturbing. In this case there **is** a reason for the elements to jump; it is to display important information about the status of the element.
> 
> I'm not sure if it is going to be annoying in practice, but I am OK with this in theory.

I'm also wary of this, but I think the compromise will be worth it for the simplicity (and I agree with the theory too!). I haven't fully formed my thoughts here, but we could avoid this by recommending a single loader for an entire form, rather than giving each element its own loading state. This has the downside of obfuscating information from the user: which elements are actually loading? And do we wait for the full response to display the resulting validations, or one by one? 

The alternative I've seen used is to bake enough padding-bottom into a `inputField` that the a form with multiple validating elements will not jump.

Let's definitely revisit this when we have been able to test out the status displaying underneath.

HendrikThePendric:
> One drawback is that when something is validating async using the loading status, then all of the elements underneath are going to jump down while the loading status is displayed, and jump back up when the async validation is done.

Yeah, I had the same concern but decided to add the loading prop anyway.... First reason is consistency, second reason is the fact that it might sometimes add a bit of value to indicate what is being loaded (i.e. "checking if username is available"). But I agree this is not ideal.

> we could avoid this by recommending a single loader for an entire form, rather than giving each element its own loading state.

I think we can distinguish two scenarios here:
- Fields that are loading, as in fetching options etc. For this scenario I agree with the above.
- Fields that are async validating. For this scenario, I think having a loading state on the actual field is required, and a single loader is not appropriate. But, on the other hand, I also think the "jumping fields" behaviour won't actually be a problem in these cases, the user has interacted with the field, then sees a change in the UI, it makes sense.

> The alternative I've seen used is to bake enough padding-bottom into a inputField that the a form with multiple validating elements will not jump.

I think this could possibly be a valid approach to reduce jumping fields in most cases. It'll probably be impossible to prevent this for Fields with multiple Help elements though (i.e. a help text + validation text).

After thinking about this approach a little bit more, **I have started to get some concerns. Because what we have done here, is that we have implicitly made the `validationText` a required property when you want to show a Field's status.** If there is no validationText, the input will only show a coloured border but no icon, so to show a "proper field status" we now _need to show_ a message....

For the `error` and `warning` message, I think this always makes sense. However, for `loading` and `valid`, I can think of quite a lot of exceptions:
- An aSync Select fetching options: I don't think messages like "Fetching data elements" add any value (but as Joe said, this scenario could possibly be covered by having a single loader)...
- In most cases where the status is `valid`, it doesn't really help to add a message, i.e. "Username available" doesn't really add much, just showing the `valid` status would be enough...

And I am afraid that the above might trigger devs to introduce more status texts to apps, and that will mean more strings to translate...

After considering all of the above, I have started to think that perhaps putting the icon in the the Help wasn't a good decision. Perhaps the approach @varl suggested was more appropriate:
image

Sorry to open up this discussion again. Yesterday it all seemed like a good idea but I have started getting seconds thoughts.

varl:
> Sorry to open up this discussion again. Yesterday it all seemed like a good idea but I have started getting seconds thoughts.

Agreed.

> After thinking about this approach a little bit more, **I have started to get some concerns. Because what we have done here, is that we have implicitly made the `validationText` a required property when you want to show a Field's status.** If there is no validationText, the input will only show a coloured border but no icon, so to show a "proper field status" we now _need to show_ a message....

Ouch. That's not good at all.

We get into similar problems if we try to put the status icon next to the label as the label is also optional.

The best would be to have the status icon to the right of the field and maybe a `min-width` to prevent this example @cooper-joe sent me:

image

HendrikThePendric:
> The best would be to have the status icon to the right of the field

👍 Agree with that 

> maybe a min-width to prevent this example @cooper-joe sent me

I don't think we necessarily need a min-width. I would say that in the components that accept `width`, we just set that width directly on the input, and not we just render the icon behind it, so the input width won't change.

In the default scenario, where the input takes up the available space, the input width will reduce in size when the icon is rendered. But I disagree with the fact that this is a problem that is introduced by this new UI: it doesn't matter if we need to make place "inside or outside" of the input, we need to free up space to show the icon regardless and this will make reduce the amount of character we can read inside of the input.

HendrikThePendric:
I think this PR should be closed in favour of showing the icons **after** the input.

The only downside of this decision is that now, we will have to come up with a good way of displaying status on input fields such as Radio/Checkbox/Switch-Group and OrgUnitTree, etc.

Issue #437, opened 2019-09-18
Highlight text selection in disabled input fields


cooper-joe:
Disabled input fields should allow text selection. This is technically possible today, but the text does not appear highlighted. 

Fix: highlighted text inside a disabled input field should be visually highlighted.

PR #472, opened 2019-10-01
refactor(input): update implementation of Input and InputField

HendrikThePendric:
The following changes have been implemented:
- Showing the status icon next to the input
- Applying status styling directly on the input element
- Remove the `focus` state property and use the `:focus` selector instead
- Simplify CSS and remove redundant rules
- Add `readOnly` prop, styles, and story
- Add `width` prop, styles and story
- Add docs to Input

Some things I intentionally left out:
- Adjust border-color for disabled style: we need confirmation about https://github.com/dhis2/design-system/issues/15 first
- Apply `labelWidth` or `width` props as mentioned in https://github.com/dhis2/ui-core/pull/456#issuecomment-536522531. I think we should keep these changes out-of-scope for now and introduce these props once we need inline labels.

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -54,15 +54,15 @@ const styles = css`
 
     input.read-only {
         background-color: ${colors.grey100};
-        cursor: text;
         border-color: ${colors.grey500};
+        cursor: text;
     }
 
     input.disabled {
-        cursor: not-allowed;
-        border-color: ${theme.disabled};
-        color: ${theme.disabled};
         background-color: ${colors.grey100};
+        border-color: ${colors.grey500};
+        color: ${theme.disabled};
+        cursor: not-allowed;
     }
 
     .status-icon {
--- a/src/Input.js
+++ b/src/Input.js
@@ -9,97 +9,81 @@ import { StatusIconNoDefault } from './icons/Status.js'
 
 const styles = css`
     .input {
-        align-items: center;
-        box-sizing: border-box;
         display: flex;
-        height: 42px;
-        left: 1px;
-        position: relative;
+        align-items: center;
+    }
+
+    input {
+        box-sizing: border-box;
+
+        font-size: 14px;
+        line-height: 16px;
+        user-select: text;
+
+        color: ${colors.grey900};
+        background-color: transparent;
+
+        padding: 12px 11px 10px;
+
+        outline: 0;
         border: 1px solid ${colors.grey500};
         border-radius: 3px;
-        font-size: 14px;
         box-shadow: inset 0 0 0 1px rgba(102, 113, 123, 0.15),
             inset 0 1px 2px 0 rgba(102, 113, 123, 0.1);
     }
 
-    .input.dense {
-        height: 34px;
+    input.dense {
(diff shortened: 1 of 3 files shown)


Mohammer5:
(I've changed the todos to a checkbox list in the description box)

ismay:
It looks good to me already, but I was wondering, where does the `readOnly` state come from? I don't see it in the design system repo. I see that there's a difference with `disabled` in the cursor for example, but practically speaking what is the intended use-case for it vs `disabled`?

HendrikThePendric:
@Mohammer5 
> (I've changed the todos to a checkbox list in the description box)

Not sure why. These aren't todos....

> Apply labelWidth or width props as mentioned in #456 (comment). I think we should keep these changes out-of-scope for now and introduce these props once we need inline labels.

This hasn't been done. I've only implemented `inputWidth` for now, because that is not necessarily related to the displaying labels inline discussion, see @cooper-joe's  comment here: https://github.com/dhis2/ui-core/pull/456#issuecomment-536486734. This prop serves its own purpose.

The other props are out of scope IMO because the discussion about if and how we should display stuff inline hasn't settled yet.

_I'll remove these checkboxes now_

HendrikThePendric:
@ismay 
> It looks good to me already, but I was wondering, where does the readOnly state come from? 

https://sketch.cloud/s/DwkDk/a/vbvjnM (Input)
https://sketch.cloud/s/DwkDk/a/wYxqvw (TextArea)
I've also read a discussion about these read only fields at some point, but no idea where.

When @cooper-joe gets back we'll find out if this style is indeed needed...

varl:
> I've also read a discussion about these read only fields at some point, but no idea where.
> 
> When @cooper-joe gets back we'll find out if this style is indeed needed...

@cooper-joe mentioned on Slack that he's working on specs for a readOnly field in the #ui channel.

ismay:
Actually, I didn't know this, never used it, but apparently readonly is a valid attribute on native inputs and textareas (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#readonly):

> The difference between disabled and readonly is that read-only controls can still function, whereas disabled controls generally do not function as controls until they are enabled.

 So I guess it's just to stay compatible with native. TIL

HendrikThePendric:
> Actually, I didn't know this, never used it, but apparently readonly is a valid attribute on native inputs

Good find! I meant to look into this but completely forgot. This impacts the implementation I think. Will work on a commit.

varl:
> readOnly

**Note:** found it in https://github.com/dhis2/ui-core/issues/437 too.

cooper-joe:
This looks good! Some info that was requested from me:
- I've responded to the open issue regarding borders for disabled inputs (https://github.com/dhis2/design-system/issues/15)
- `read only` is required, yes. From what I see here, it's correctly implemented too 👍 thanks!

HendrikThePendric:
I've corrected the `border-color` for the disabled state, and I think no other changes are required. Could someone approve this PR? (or request changes)
@Mohammer5 @varl @ismay @cooper-joe

PR #476, opened 2019-10-04
test: add cypress setup & AlertBar tests

Mohammer5:
I've added cypress so we can set up both functional and visual tests in conjunction with storybook.
To demo this, I've also added both functional and visual tests for the AlertBar.

I tried adding cucumber as well, but it failed due to some dependency version mismatches.
While debugging it, I realized that cucumber might actually be the wrong choice here as visual tests are not behavior tests.

For visual tests I've used [`cypress-match-screenshot`](https://www.npmjs.com/package/cypress-match-screenshot), but there are [several other options](https://docs.cypress.io/plugins/#visual-testing) available which we could use instead, I just used the first google result.

Furthermore I've added the possibility to add testing stories, so we can have more sophisticated stories which won't end up on the docs pages (see README for more information)

<hr />

We can use this to properly test the Pop component and other components that will have internal state and functionality (such as the new select components)


varl:
How do I run this to test it?

Mohammer5:
@varl Oh, I forgot to add scripts & explanation, will add

varl:
Is it possible to make `cypress:open` just run the server so I can manually attach a browser? Otherwise this setup is completely broken in certain environments like WSL.

Mohammer5:
@varl I've added scripts and a testing section to the README

> Is it possible to make cypress:open just run the server so I can manually attach a browser? Otherwise this setup is completely broken in certain environments like WSL.

I have no idea what you're trying to say. What server, `cy.server()`?

Mohammer5:
@varl Any idea why the build breaks? `yarn build` exits with 0, then the travis script exits with 1.
I've cleared the cache for this branch of travis already, didn't help

varl:
There are lint violations on CI: https://travis-ci.com/dhis2/ui-core/builds/130400695#L208-L218

Mohammer5:
I've fixed the build as well

ismay:
Interesting. I tried it out, and everything works as expected, except for the last 3 screenshots diffs, which fail:

Screenshot 2019-10-07 at 10 18 44

I imagine that those failed because I'm generating the diffs on a different platform than you @Mohammer5? I couldn't find a side-by-side screenshot diff option, so I'm not sure what the difference is exactly.

Mohammer5:
> I imagine that those failed because I'm generating the diffs on a different platform than you @Mohammer5? I couldn't find a side-by-side screenshot diff option, so I'm not sure what the difference is exactly.

Interesting.. Of course this shouldn't be the case. I'll try a few different plugins and will ask you to test them on your system

varl:
The functional tests look good!

Screenshot testing is generally only reliable when run on the exact same browser/platform combo on e.g. Browserstack. I don't think it's a great addition to this setup if it is going to be flaky since we don't have anywhere to run those tests as CI tests for now.

cypress/integration/AlertBar.spec.js
```diff
@@ -0,0 +1,59 @@
+/* global cy */
+
+import { AlertBar } from '../../src'
+
+describe('AlertBar - Functional', () => {
```
HendrikThePendric: We are not always going to be adding the word "Functional" here right?

Mohammer5:
> Screenshot testing is generally only reliable when run on the exact same browser/platform combo on e.g. Browserstack. I don't think it's a great addition to this setup if it is going to be flaky since we don't have anywhere to run those tests as CI tests for now.

That would s*** a lot.. But actually I'm wondering why it's not working, because the screenshot that's been taken doesn't contain anything browser specific:


@ismay can you delete the screenshots, record them again and share them with me?

stories/AlertBar.stories.js
```diff
@@ -22,27 +23,35 @@ storiesOf('AlertBar', module)
     .add('Default', () => <AlertBar>Default - I will autohide</AlertBar>)
     .add('States', () => (
         <React.Fragment>
-            <AlertBar permanent>Default (info)</AlertBar>
-            <AlertBar permanent success>
+            <AlertBar className="first" permanent>
```
HendrikThePendric: I vaguely recollect that it would be preferred to work with `data-test-id` attributes for selecting elements in functional tests. Not sure why that would be better tbh...

Did we decide that classnames are just as good?

cypress/integration/AlertBar.spec.js
Mohammer5: I just didn't know a better word for describing the difference between functional and visual tests

HendrikThePendric: Aha OK makes sense

stories/AlertBar.stories.js
Mohammer5: I think we've decided on `data-test-id`, but we haven't decided on how to add them yet (although manually adding them seems the only do-able option to me).
But as there was no api for this yet on this component (or any other in `ui-core`), I think using the className is just as fine for now.

I'm not sure we should add that api either, in the import export app I've wrapped the ui-core components with an element that has the id, so it's purely the apps responsibility to take care of these IDs

ismay:
> > Screenshot testing is generally only reliable when run on the exact same browser/platform combo on e.g. Browserstack. I don't think it's a great addition to this setup if it is going to be flaky since we don't have anywhere to run those tests as CI tests for now.
> 
> That would s*** a lot.. But actually I'm wondering why it's not working, because the screenshot that's been taken doesn't contain anything browser specific:
> 
> 
> @ismay can you delete the screenshots, record them again and share them with me?

So these are the new screenshots from my test run:

AlertBar - Visual -- should have different states -- AlertBar states
AlertBar - Visual -- should suports icons -- AlertBar icons
AlertBar - Visual -- should wrap text -- AlertBar text wrap

And these are the diffs:

AlertBar - Visual -- should have different states -- AlertBar states
AlertBar - Visual -- should suports icons -- AlertBar icons
AlertBar - Visual -- should wrap text -- AlertBar text wrap

ismay:
Maybe it's not really a browser issue, but a cross platform cypress issue?

varl:
Flip the base to `master`, clean up eventual conflicts, then we can merge this.

Mohammer5:
@varl done

PR #479, opened 2019-10-06
feat(radio-group): add disabled prop

teug91:


Mohammer5:
@HendrikThePendric You were working on this component, right? I think right now there's not RadioGroup in the `next` branch.

PR #482, opened 2019-10-07
chore(prop-types): upgrade prop-types package to prevent false positives

HendrikThePendric:


Mohammer5:
Duplicate of #486 
We should not update deps in `next`, this can be done on master (as dependabot is doing that for us) and can then be merged into `next`

PR #501, opened 2019-10-14
refactor(styled-jsx): use resolve instead of global

HendrikThePendric:
### Context
I started work on this PR because it was recently discussed in Slack that we should be using `resolve` instead of `:global()` wherever possible.

### Concerns about style override mechanism
Working on this PR, I encountered a generic problem with our style-overriding mechanism, which I'll explain using the `FileInput` as an example:

Note that below I am using not only using the `svg` CSS selector, but I have added the class-name `file-input` to the selector too:
https://github.com/dhis2/ui-core/blob/f0b4ddb3e67220c5850e0755903dcd62d61551d0/src/FileInput.js#L12-L18
And the component with the dynamic+static class-name is being initialised here:
https://github.com/dhis2/ui-core/blob/f0b4ddb3e67220c5850e0755903dcd62d61551d0/src/FileInput.js#L80-L85

The reason I have done so has to do with CSS selector precedence:
- Without the extra static class-name, both the style rules within the component and the style override rules have the same specificity, i.e. `svg.jsx-123456789` and `svg.jsx-987654321`
- But since the styles for the `StatusIconNoDefault` is injected **after** the styles for `FileInput`, the `StatusIconNoDefault` get prioritised.
- So to combat this, I opted for adding an extra level of specificity, which causes the style-override style rule to get priority: i.e.  `svg.jsx-987654321.file-input` takes precedence over `svg.jsx-123456789`. The only alternative I could think of was adding `!important`, but that's even worse.

I wasn't aware of the above and find it quite problematic. I'd like to find out if there is a better way to do this, and/or if you guys were aware of this problem...

Just to clarify: what I mainly find problematic is this:
- We are now adding static classNames to components
- The presence of these will increase the chances of styles bleeding into our components.
- This will especially be a problem for style properties that are not present in our own style definitions.

### Specific issues
I also encountered some component-specific issues. I'll place some comments in my code to highlight them.

### Places where I left `:global()`
I think the usage of `:global()` is correct in this case, since we want to limit the width and height of whatever is passed in here....
https://github.com/dhis2/ui-core/blob/c3481402a10d5860a51a698db1706cb2614dae0a/src/Chip/Icon.js#L23-L26

Since a `Tab` may receive any type of icon, we cannot assume that we can override the styles via our class-name mechanism. So I chose to keep `:global()`. The current setup is still slightly brittle, because we are clearly expecting a SVG, but I don't think there is any way to enforce this.
https://github.com/dhis2/ui-core/blob/c3481402a10d5860a51a698db1706cb2614dae0a/src/Tabs/Tab.js#L91-L93

I also left another `:global()` selector in `Tab` and because I wasn't sure how to solve this, I created an issue https://github.com/dhis2/ui-core/issues/503:
https://github.com/dhis2/ui-core/blob/c3481402a10d5860a51a698db1706cb2614dae0a/src/Tabs/Tab.js#L43-L45


src/ButtonStrip.js
```diff
@@ -14,9 +28,27 @@ import styles from './ButtonStrip/styles.js'
...
-        <style jsx>{styles}</style>
+        {Children.map(children, child =>
+            cloneElement(child, {
+                className: cx(child.props.className, buttonStyles.className, {
+                    start: !middle && !end,
+                    middle,
+                    end,
+                }),
+            })
+        )}
```
HendrikThePendric: Code complexity increased quite a bit here to switch to style-overrides via resolve. Since the risk of styles bleeding in via the generic `start`/`middle`/`end` classNames is still very real, I wonder if this is worth it.

src/icons/Status.js
```diff
@@ -183,12 +184,15 @@ export const StatusIconNoDefault = ({
...
+    ) : info ? (
+        <Info className={className} />
     ) : null
 
 StatusIconNoDefault.propTypes = {
     valid: propTypes.bool,
     error: propTypes.bool,
     warning: propTypes.bool,
     loading: propTypes.bool,
+    info: propTypes.bool,
```
HendrikThePendric: I added the `info` prop to the `StatusIconNoDefault` because:
- It makes sense, since we also have an `Info` icon-component in `Status.js`
- It simplified the code in the `AlertBar` quite a bit

src/AlertBar/Dismiss.js
```diff
@@ -1,22 +1,21 @@
...
+import cx from 'classnames'
 import propTypes from '@dhis2/prop-types'
 import { spacers } from '../theme.js'
 import { Close } from '../icons/Close.js'
+import { iconStyles } from './styles.js'
 
 const Dismiss = ({ onClick }) => (
     <div onClick={onClick}>
-        <Close />
+        <Close className={cx(iconStyles.className, 'dismiss')} />
```
varl: Aren't these in the wrong order? I expect `cx('dismiss', iconStyles.className)`.

Does that change anything?

HendrikThePendric: I suppose that would be more readable, so I'll go ahead and change that.... However it doesn't change anything.

HendrikThePendric:
Based on this PR @varl and I concluded that using `resolve` is not always the preferred approach when styling children (or styling a component based on a parent's class). In general the changes in this PR were not an improvement.

Our conclusion is that the usage of `:global()` is permitted when both of the conditions below are met:
1. Using `resolve` adds complexity to the code
2. The style rule cannot leak out, in practice this means that you need to always use a scoped selector in combination with the global one.

This was a useful experiment, but I am now closing this PR.

PS: This PR also includes some unrelated improvements to the `StatusIconNoDefault`, `SplitButton` and `AlertBar`. These changes will be implemented slightly differently in new PRs.

PR #502, opened 2019-10-14
refactor(statusicon): change statusicon to accept status instead of flags

ismay:
Proposal for a slight API change for StatusIcon as discussed on slack. This affects a couple of other areas as well. The reasoning behind this is that the current API for statusicon uses flags for a concept that is more accurately expressed as an enum.

We're now catching this with a special propType rule, but I think for the end-user and us the API contract would be more clear if we'd accept a single prop that accepts a status string. That way it's immediately clear that it's mutually exclusive. 

p.s.: not all the components accept all the statuses that the StatusIcon supports. So I haven't inherited from StatusIcon's props, but defined the accepted types separately. Something to discuss.

p.p.s.: @HendrikThePendric mentioned the mutually exclusive booleans are also used in other areas. For consistency accepting this would probably mean taking a look at those as well

Still todo if we want this:

- [ ] Update jsdoc

### DIFF
--- a/src/icons/Status.js
+++ b/src/icons/Status.js
@@ -168,27 +168,25 @@ Loading.propTypes = {
     className: propTypes.string,
 }
 
-export const StatusIconNoDefault = ({
-    error,
-    warning,
-    valid,
-    loading,
-    className,
-}) =>
-    valid ? (
-        <Valid className={className} />
-    ) : warning ? (
-        <Warning className={className} />
-    ) : error ? (
-        <Error className={className} />
-    ) : loading ? (
-        <Loading className={className} />
-    ) : null
+export const StatusIcon = ({ status, defaultTo: DefaultTo, className }) => {
+    switch (status) {
+        case 'info':
+            return <Info className={className} />
+        case 'warning':
+            return <Warning className={className} />
+        case 'loading':
+            return <Loading className={className} />
+        case 'error':
+            return <Error className={className} />
+        case 'valid':
+            return <Valid className={className} />
+        default:
+            return DefaultTo ? <DefaultTo className={className} /> : null
+    }
+}
 
-StatusIconNoDefault.propTypes = {
-    valid: propTypes.bool,
-    error: propTypes.bool,
-    warning: propTypes.bool,
-    loading: propTypes.bool,
+StatusIcon.propTypes = {
+    status: propTypes.oneOf(['info', 'warning', 'error', 'loading', 'valid']),
+    defaultTo: propTypes.elementType,
     className: propTypes.string,
 }
--- a/src/AlertBar.js
+++ b/src/AlertBar.js
@@ -84,42 +84,31 @@ class AlertBar extends PureComponent {
     }
 
     shouldAutoHide() {
-        const { permanent, warning, critical } = this.props
-        return !(permanent || warning || critical)
+        const { permanent, status } = this.props
+
+        const warning = status === 'warning'
+        const error = status === 'error'
+
+        return !(permanent || warning || error)
     }
 
     render() {
-        const {
-            className,
-            children,
-            success,
-            warning,
-            critical,
-            icon,
-            actions,
-        } = this.props
+        const { className, children, status, icon, actions } = this.props
         const { visible, hidden } = this.state
 
         if (hidden) {
             return null
         }
 
-        const info = !critical && !success && !warning
-        const iconProps = { icon, critical, success, warning }
-
         return (
             <div
-                className={cx(className, {
-                    info,
-                    success,
-                    warning,
-                    critical,
+                className={cx(className, status, {
                     visible,
                 })}
                 onMouseEnter={this.stopDisplayTimeOut}
                 onMouseLeave={this.startDisplayTimeout}
             >
-                <Icon {...iconProps} />
+                <Icon icon={icon} status={status} />
                 <Message>{children}</Message>
                 <Actions actions={actions} hide={this.hide} />
                 <Dismiss onClick={this.hide} />
@@ -129,11 +118,6 @@ class AlertBar extends PureComponent {
         )
     }
 }
-
-const alertTypePropType = propTypes.mutuallyExclusive(
-    ['success', 'warning', 'critical'],
-    propTypes.bool
-)
 
 /**
  * @typedef {Object} PropTypes
@@ -156,9 +140,7 @@ const alertTypePropType = propTypes.mutuallyExclusive(
 AlertBar.propTypes = {
     children: propTypes.string.isRequired,
     className: propTypes.string,
-    success: alertTypePropType,
-    warning: alertTypePropType,
-    critical: alertTypePropType,
+    status: propTypes.oneOfType(['valid', 'warning', 'error', 'info']),
     icon: iconPropType,
     duration: propTypes.number,
     permanent: propTypes.bool,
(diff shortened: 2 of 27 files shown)


varl:
Some thoughts:

- In the long term we do not want to have our icons in ui-core, we will end up with something like ui-icons, how does this affect the `StatusIcon`?

- In ui-core@v1 we used string enums for types (e.g. primary, secondary, destructive, etc.) and status (warning, loading, error, etc.) and in v2 (iirc https://github.com/dhis2/ui-core/commit/6364f4c16b8a8371abb61ad161b329215edb834b#diff-7b1f8a22119441d4c8b40f777b7578c8) we rewrote the entire API to consistently use flags. Are there overwhelming reasons to rewrite it again to string enums? 

- A consistent API is a guiding principle of ui-core, so this proposal cannot happen in isolation of a single component and needs to be considered for all of ui-core.

src/icons/Status.js
```diff
@@ -168,27 +168,25 @@ Loading.propTypes = {
...
-    ) : loading ? (
-        <Loading className={className} />
-    ) : null
-
-StatusIconNoDefault.propTypes = {
-    valid: propTypes.bool,
-    error: propTypes.bool,
-    warning: propTypes.bool,
-    loading: propTypes.bool,
+export const StatusIcon = ({ status, defaultTo: DefaultTo, className }) => {
```
HendrikThePendric: Regardless of the flag API discussion, I think that the addition of the `defaultTo` prop and renaming `StatusIconNoDefault` to `StatusIcon` is a good idea. This change seems to be fully unrelated to the flags.

src/icons/Status.js
```diff
@@ -168,27 +168,25 @@ Loading.propTypes = {
...
+        case 'warning':
+            return <Warning className={className} />
+        case 'loading':
+            return <Loading className={className} />
+        case 'error':
+            return <Error className={className} />
+        case 'valid':
+            return <Valid className={className} />
+        default:
+            return DefaultTo ? <DefaultTo className={className} /> : null
```
HendrikThePendric: It would be more consistent with the rest of the codebase if we use this prop like this:
```
<StatusIcon
    defaultTo={<MyDefaultIcon className="yeah-baby />}
/>
```
And also, I think we should not default to null in the switch statement, but rather use a `defaultProp`.

The default case would then look like this:
`return defaultTo`

And you also wouldn't need to alias the prop in the function signature.

FYI: we are using this same approach approach for icons in components such as `AlertBar`, `Chip`, `Tab` etc....

ismay:
> In ui-core@v1 we used string enums for types (e.g. primary, secondary, destructive, etc.) and status (warning, loading, error, etc.) and in v2 (iirc 6364f4c#diff-7b1f8a22119441d4c8b40f777b7578c8) we rewrote the entire API to consistently use flags. Are there overwhelming reasons to rewrite it again to string enums?

> @varl and I came up with this mutually exclusive boolean props approach because we think typing <MyComp error /> is easier than typing <MyComp status="error" />. 

> Also using a string/enum is a more accurate representation of what is really going on....

I would argue yes. To me ease of typing is relatively low on the list of things to consider for an API. When having to choose I'd prefer a clear API over ease of typing. Especially since the typing benefits aren't that huge here.

We're now having to enforce the extra restrictions that we have on top of the flags (the fact that they're mutually exclusive), where a different type would express that restriction in a way that's more obvious. An clear API is a benefit to usability as well I'd argue, just like an API that's not overly verbose.

It seems also like something that's more in line with native html elements. All in all it seems to me like mutually exclusive flags are quite a unique pattern, and I don't see why we'd diverge from the enum approach, which seems like the more established pattern for this kind of api to me.

> In the long term we do not want to have our icons in ui-core, we will end up with something like ui-icons, how does this affect the StatusIcon?

I can imagine that we'd drop StatusIcon, I'm not sure. If we agree with this PR though, it'd affect more than just the StatusIcon. I can imagine we'd drop the StatusIcon, but that we'd still refactor the pattern in other areas.

> A consistent API is a guiding principle of ui-core, so this proposal cannot happen in isolation of a single component and needs to be considered for all of ui-core.

I agree 👍

src/icons/Status.js
ismay: Yeah I think we could implement that separately 👍

src/icons/Status.js
ismay: Ok n.p. In a review @Mohammer5 warned me against using defaultProps too much, as we didn't use them too much in ui-core, but that might have been a special case. That's the reason I'm not using it here. I'm fine with your suggestion though, seems good to me 👍.

src/icons/Status.js
ismay: Shall I split up the PR into two? `Add defaultTo prop`, and `Use enums instead of mutually exclusive boolean?s`

HendrikThePendric: Not sure... Perhaps see where this discussion goes first?

varl:
> We're now having to enforce the extra restrictions that we have on top of the flags (the fact that they're mutually exclusive), where a different type would express that restriction in a way that's more obvious. An clear API is a benefit to usability as well I'd argue, just like an API that's not overly verbose.

We don't have to use additional checks on the props, IMO. The `mutuallyExclusive` is sugar on top to catch potential pitfalls for developers where they accidentally use multiple props that conflict. Something which can happen with string-based enum properties too, but harder to catch.

These four usages are equivalent:

```
<Button error warning valid>
<Button valid>
<Button status="error" status="warning" status="valid">
<Button status="valid">
```

The first style because of how the classes are applied and built up, and the second style because of how React consolidates multiple duplicate props. React doesn't even care to warn you that you have defined the same property twice and you will have to rely on e.g. ESLint to catch it.

Another concern that we weighted is that it is easier to reason about:

```
cx('button', {
  error,
  warning,
  valid
})
```

Than what we had before:

```
cx('button', {
  `${props.status}`: !!props.status,
  `${props.kind}`: !!props.kind,
})
```

There was a layer of indirection between the name of the class, the property used, and the value of the property. Using a flag allowed all of the bindings between class name and prop name to be very clear.

I'd recommend going back to late v1 and not just read the code, but try to do some work with it. It seems trivial at a glance but caused real problems as the library grew.

> It seems also like something that's more in line with native html elements. All in all it seems to me like mutually exclusive flags are quite a unique pattern, and I don't see why we'd diverge from the enum approach, which seems like the more established pattern for this kind of api to me.

It's not unique in terms of React components, and this is a React component library.

I agree in principle that aligning with HTML APIs is great as a rule of thumb and the starting point. This is what we did in v1, then we evaluated pros/cons and decided to go in another direction. There are some things that are simpler with each style and some things that are harder.

In HTML elements there are also flags, such as `checked`, `selected`, `disabled`, so it is not exactly uncommon there either.

We didn't consider all the trade-offs involved, but we have a pretty good idea of the practical ones after working with each style for 6 months.

I would urge you to please argue the fundamental problems, technical or otherwise, as to why flags is not going to work out for our UI components? As long as there are not any fundamental problems with the approach, there is no overwhelming reason to rewrite the API as I see it.

ismay:
Viktor and I talked a bit on slack and he filled me in a bit on the background. If I'm paraphrasing correctly, we're shelving this for now, and then we can potentially revisit this later, for another major version. We can then also document the process of discussing this api change, so that we'll have good documentation for the decision itself. That way `next` wouldn't be blocked by this, and we'd keep the api stable for now.

varl:
Since we are shelving this I'm closing the PR.

PR #505, opened 2019-10-14
refactor: alert bar and status icon

HendrikThePendric:
- Added `info` and `defaultTo` props to status-icon
- Rename `StatusIconNoDefault` to `StatusIcon`
- Let AlertBar/Icon use `StatusIcon` to simplify component code

### DIFF
--- a/src/icons/Status.js
+++ b/src/icons/Status.js
@@ -179,17 +179,21 @@ export const StatusIcon = ({
 }) => {
     if (error) {
         return <Error className={className} />
-    } else if (warning) {
-        return <Warning className={className} />
-    } else if (valid) {
-        return <Valid className={className} />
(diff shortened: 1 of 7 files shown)


src/icons/Status.js
```diff
@@ -168,12 +168,14 @@ Loading.propTypes = {
     className: propTypes.string,
 }
 
-export const StatusIconNoDefault = ({
+export const StatusIcon = ({
```
varl: @ismay This looks like it contains the changes we discussed as well, except for the "get rid of the nested ternary" comment.

ismay: Yeah exactly. Maybe we could get rid of the ternary while we're at it? Switching based on the props for example?

Issue #492, opened 2019-10-09
Input: add types


cooper-joe:
Add types to `input` component, as documented in the [updated design documentation](https://github.com/dhis2/design-system/blob/master/atoms/inputfield.md#types)

Types required:

- [ ] number
- [ ] password
- [ ] email
- [ ] url
- [ ] telephone

HendrikThePendric:
Closing this issue because this was implemented in #511

PR #511, opened 2019-10-15
fix(input): allow types specified in design-system

HendrikThePendric:
See https://github.com/dhis2/ui-core/issues/492 for allowed types

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -193,8 +193,14 @@ Input.defaultProps = {
  */
 Input.propTypes = {
     name: propTypes.string.isRequired,
-    type: propTypes.oneOf('text', 'number', 'password', 'email', 'url', 'tel')
-        .isRequired,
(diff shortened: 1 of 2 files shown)


src/Input.js
```diff
@@ -193,7 +193,8 @@ Input.defaultProps = {
  */
 Input.propTypes = {
     name: propTypes.string.isRequired,
-    type: propTypes.string.isRequired,
+    type: propTypes.oneOf('text', 'number', 'password', 'email', 'url', 'tel')
```
Mohammer5: According to the specs, this needs to be an array, Does it work this way?
https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes

Regardless of whether it works or not, I'd stick to the official way

HendrikThePendric: Sorry, that was just plain sloppy. Corrected it now

PR #513, opened 2019-10-16
chore: enable cucumber, rewrite existing tests

Mohammer5:
This PR covers several topics

## Cucumber

I've added the `cypress-cucumber-preprocessor` npm module.
This allows us to write `.feature` files in the `cypress/integration` folder.

I've set `"nonGlobalStepDefinitions": true` in the `package.json` to enable us to place the step definitions in folders with the same name as the `.feature` file

## Rewrite existing tests

I've rewritten the previously existing tests for the AlertBar.
There are now two feature files in `cypress/integration`:
* `AlertBar/Autohiding.feature`
Step definitions in ``cypress/integration/AlertBar/Autohiding/`
* `AlertBar/ClearOnAction.feature`
Step definitions in ``cypress/integration/AlertBar/ClearOnAction/`

## Add scripts

I've added 5 scripts to the package.json of which 2 are of importance to us:
* `cypress:open`
* `cypress:run`

These will both run the storybook and then cyrpress.
What to check here:
(When running `yarn cypress:run`, make sure your BROWSER env var contains the path to either chrome or chromium: `BROWSER=$(which chrome) yarn cypress:run`)

1. Correct behavior
   * `yarn cypress:open` should simply open the cypress interface
   * `yarn cypress:run` should run the tests inside the terminal
   I've tested it on a different tty to ensure it's working properly because in ttys with a window manager, it still opens a real browser
   * When either of the two scripts are terminated or  `cypress:open` process is terminated manually:
   There should be no running process of the storybook or cypress
   * `cypress:run` should exit with code 0 if all tests pass
1. Error management
   * the storybook port is already in use
   The script should exit with code 1
   * the tests did not pass successfully
   The `cypress:run` script should exit with code 1


Mohammer5:
Lacks docs and proper PR description.
Will add them for better understanding

Mohammer5:
> Some file have some complex looking logic in them, i.e. cypress/assets/unfetch.umd.js and possibly this is something we'd need to copy into every repo that uses Gherkin. Perhaps we can use some form of package to take care of this? Or perhaps keep all these scripts in a central location (@dhis2/cli or sth)?

Unfetch is only temporarily needed until cypress supports hooking into fetch.
@amcgee and I will create a wrapper library that will take care of installing and setting up all of this in app platform apps (I've added a PR for setup script in `@dhis2/cli` but we both prefer an isolated library that will also mange the cypress and cucumber version)

amcgee:
@Mohammer5 why do we need `unfetch` here?  I don't think `ui-core` components should be doing fetches?

In cypress tests for *applications* we should be able to simply `delete window.fetch` (I did this in [maps app](https://github.com/dhis2/maps-app/blob/master/cypress/support/network-fixtures.js#L13-L17))and then rely on the app's polyfill to stub it out, I think....  am I missing something here?

Mohammer5:
> @Mohammer5 why do we need unfetch here? I don't think ui-core components should be doing fetches?
> 
> In cypress tests for applications we should be able to simply delete window.fetch (I did this in maps app)and then rely on the app's polyfill to stub it out, I think.... am I missing something here?

You're right, we don't need it in `ui-core` at all.. 
`ui-widgets` does need it though as there's no polyfill in the storybook (that's where I copied it from).

I'll remove it from this PR

amcgee:
Nice!

I think it also shouldn't be necessary in `ui-widgets` though right?  Because the `ui-widgets` storybook uses a [CustomDataProvider](https://github.com/dhis2/ui-widgets/blob/master/stories/HeaderBar.stories.js#L114) which intercepts all data requests, so they shouldn't turn into fetches either.

.eslintrc.js
```diff
@@ -3,4 +3,9 @@ const { config } = require('@dhis2/cli-style')
 module.exports = {
     parser: 'babel-eslint',
     extends: ['eslint:recommended', 'plugin:react/recommended', config.eslint],
+    env: { es6: true },
+    globals: {
+        Cypress: 'readonly',
```
amcgee: These should only apply to the `cypress` folder, if possible.  Can we put a subsidiary `.eslintrc` in `cypress/`?  That might not be supported by cli-style...  @varl?

cypress.json
```diff
@@ -1,6 +1,7 @@
 {
-  "screenshots": {
-    "screenshotOnRunFailure": false
-  },
-  "projectId": "65oh1u"
+    "testFiles": "**/*.feature",
+    "screenshots": {
+        "screenshotOnRunFailure": false
```
amcgee: Why disable this?

Mohammer5:
Nice!

I think it also shouldn't be necessary in ui-widgets though right? Because the ui-widgets storybook uses a CustomDataProvider which intercepts all data requests, so they shouldn't turn into fetches either.

I'm using that to test different scenarios, like checking what happens when there are messages and when there are not messages. 
Alternatively we could add another scenario to the storybook instead if that's what should be preferred. I kinda liked the way how I was able to override certain responses though:

[Setting the defaults](https://github.com/dhis2/ui-widgets/blob/master/cypress/integration/HeaderBar/common/index.js)
[Overwriting the defaults](https://github.com/dhis2/ui-widgets/blob/master/cypress/integration/HeaderBar/Messages/index.js)

It works quite well with gherkin as well, as there'd be nothing to be done by the `Given` step definitions if it was done by the story already

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
amcgee: Probably should remove this example fixture if we're not using it

cypress.json
Mohammer5: I think this should be explicitly turned on for CI. During development this won't be needed as we're probably debugging in the browser anyways

.eslintrc.js
Mohammer5: Oh yeah! That'd be awesome if supported

Mohammer5:
>  what is the reasoning behind everything in scripts/cypress

I've tried to use concurrently and wait-on. Unfortunately wait-on doesn't support waiting for a 200 response, so it'll continue earlier than when storybook is ready. (See ["Goals"](https://github.com/jeffbski/wait-on#goals))

The scripts also make sure that the script fails if the port used by both cypress and storybook is identical, it exits with 1 if the port is already in use. Storybook just starts with an incremented port if the specified one is already in use and there's no way to know about that except for analyzing the output.

> I'm also not sure what cypress:ci is for, why isn't that just cypress:run?

`cypress:ci` is used internally by the script that's called from `cypress:run`:
* `cypress:run` starts the `ci` process
* `cypress:ci` is the process that expects the storybook to be available

I guess the naming is a bit odd here, I'd appreciate alternative names that are more descriptive!

cypress/fixtures/example.json
Mohammer5: Removed

amcgee:
> I've tried to use concurrently and wait-on. Unfortunately wait-on doesn't support waiting for a 200 response, so it'll continue earlier than when storybook is ready. (See "Goals")

Do we need to wait for a `200` response or just a `GET` response?  With webpack-dev-server this was solved by using `wait-on http-get://....`, but maybe storybook's server responds with non-200 to GET when it isn't ready?  That would surprise me, but possible.

Mohammer5:
> Do we need to wait for a 200 response or just a GET response? With webpack-dev-server this was solved by using wait-on http-get://...., but maybe storybook's server responds with non-200 to GET when it isn't ready? That would surprise me, but possible.

I haven't tried `http-get` protocol. Will do and see if that works

amcgee:
> The scripts also make sure that the script fails if the port used by both cypress and storybook is identical, it exits with 1 if the port is already in use. Storybook just starts with an incremented port if the specified one is already in use and there's no way to know about that except for analyzing the output.

Just throwing out possibly-simpler solutions here, since I'm worried that adding a bunch of custom development scripts to a single repo (though I know we'll extract them into a wrapper lib eventually):

Could we use [detect-port](https://www.npmjs.com/package/detect-port)'s CLI to either fail if the port isn't available or pass a random available port to Storybook and on to Cypress?

amcgee:
> I haven't tried http-get protocol. Will do and see if that works

Found that after a lot of frustration and debugging with this same issue in maps, it's handy!

cypress.json
varl: Why remove the `projectId` for the Cypress Dashboard?

Mohammer5:
Thanks to @amcgee I was able to get rid of the scripts folder and do everything with concurrently and wait-on :+1:.
Please review again!

.eslintrc.js
varl: @Mohammer5 Any success with this?

Mohammer5: Yes! It's working!

cypress.json
Mohammer5: It's just at the bottom of the file

Mohammer5:
@varl @amcgee I think this is ready for another review

varl:
LGTM.

As I mentioned in the other Cypress/Cucumber PR: I don't have experience setting it up so basing my review on how it looks and works. If there are any improvements to be made they can be done incrementally.

varl:
Would be great if you could take another look @amcgee.

cypress/fixtures/example.json
amcgee: It looks like this file is still here?

cypress.json
```diff
@@ -1,6 +1,5 @@
 {
-  "screenshots": {
-    "screenshotOnRunFailure": false
-  },
-  "projectId": "65oh1u"
+    "testFiles": "**/*.feature",
+    "video": false,
```
amcgee: Why are we setting `video: false` here?  Seems like we want videos when doing recorded CI runs...

cypress/fixtures/example.json
Mohammer5: No idea what happened there.. It's gone now

cypress.json
Mohammer5: Same reason as for the screenshots: I think this should be explicitly turned on for CI. During development this won't be needed as we're probably debugging in the browser anyways.

These files will be git-ignored, so on the one hand devs won't necessarily notice these files have been created and on the other hand is the amount of disk space used by videos quite a bit.

amcgee: Ah you're right, totally fair!  I'd used the `--video false` flag for that, but putting it in the config file is definitely better if we can override it explicitly when running in CI.  Nice.

Mohammer5: Ok.. seems like right now that's not an option: kimmobrunfeldt/concurrently#33
So I'm inclined to remove these options from cypress.json until this 3.5 year old issue has been solved..... Do you agree?

amcgee: https://docs.cypress.io/guides/guides/environment-variables.html#Setting

Looks like an env bar would work too?

amcgee: (Untested)

Mohammer5: It's a bit more complicated. We'd have to set the config with code, depending on an env var, and I wasn't successful in doing so (I didn't try extensively though).

I think for now it's ok to remove it from the config and do this on a new/another branch/PR

amcgee: This works (with `"video": false` in `cypress.json`)

```
CYPRESS_VIDEO=true cypress run
```

I think you're right, it's better not to create screenshots or videos on a dev machine so let's default those to **false** in `cypress.json` (as you had done) and override them with the env vars in `.travis.yml`

amcgee: (no configuration with code required)

Mohammer5: I've added this functionality for taking screenshots manually as that didn't work.
Works for videos out of the box.

PR #514, opened 2019-10-16
refactor(checkbox): change outline to be in document flow

ismay:
I ran into something where the checkbox icon was rendered over the filter input in the `Select` because of the relative positioning on the checkbox icon wrapper:

Screenshot 2019-10-16 at 17 07 38

I could fix this from the Select, by setting a higher z-index, but this would fix it as well _and_ mean other components also won't have to deal with this problem.

Appearance of the checkbox before the fix:

Screenshot 2019-10-16 at 18 32 16

Appearance of the checkbox after the fix:

Screenshot 2019-10-16 at 18 32 07

You can see that after the fix, there is more space between the border and the checkbox. That's because the svg it's wrapping contains some whitespace around it:

Screenshot 2019-10-16 at 18 38 14

And that's what the border is wrapping around. To fix that, we should crop the svg to just the checkbox. That way there would be no need for relative/absolute positioning, and everything would still look the same.

### DIFF
--- a/src/Checkbox/Icon.js
+++ b/src/Checkbox/Icon.js
@@ -71,17 +71,13 @@ export const Icon = ({
             {icons.styles}
             <style jsx>{`
                 div {
-                    position: relative;
+                    border: 2px solid transparent;
+                    border-radius: 4px;
                     margin: 0 6px 0 0;
                 }
 
-                .focus:before {
-                    content: '';
-                    position: absolute;
-                    border: 2px solid ${colors.blue600};
-                    border-radius: 4px;
-                    width: 100%;
-                    height: 100%;
+                div.focus {
+                    border-color: ${colors.blue600};
                 }
             `}</style>
         </div>
--- a/src/Checkbox/Icon.js
+++ b/src/Checkbox/Icon.js
@@ -71,13 +71,12 @@ export const Icon = ({
             {icons.styles}
             <style jsx>{`
                 div {
-                    border: 2px solid transparent;
                     border-radius: 4px;
                     margin: 0 6px 0 0;
                 }
 
                 div.focus {
-                    border-color: ${colors.blue600};
+                    box-shadow: inset 0 0 0 2px ${colors.blue600};
                 }
             `}</style>
         </div>


Mohammer5:
This is something that's bothering me for quite a while now. All our svgs contain some space around them, which makes it very hard to set the desired dimensions of the actual icon.

I don't remember where we got this icon from, is there an online sketch file that we can "take apart" or does @cooper-joe have to give us the assets manually (if that's the case, I think we need to improve this workflow somehow for future components)

Mohammer5:
Is there a branch I can use to easily reproduce this?
The bug you've encountered seems a bit weird to me. Normally adding position relative to an element inside a container (here: the checkbox container) should not cause the container to overlap with elements next to it.

varl:
I pulled the SVG icons from: https://github.com/google/material-design-icons

cooper-joe:
The Material Design icons from Google include the built-in padding around them. The reasoning is basically: 'the padding protects the icon from being placed too closed to the items around it'. We don't really need that kind of hand holding, so I'm happy to remove it. I worked a lot with @HendrikThePendric on this, so I'd like to get his input here too.

I did in fact rebuild the SVG icons just yesterday for the `radio` and `checkbox` components. The files are [included as resources in the design-system](https://github.com/dhis2/design-system/tree/master/resources). (I included the padding for the `regular` sizes, but I can remove this if we will refactor.)

The rebuilt SVGs:
- use only `fill`, never stroke, to allow for easier CSS manipulation
- have meaningful, consistent layer names, to reduce required CSS rules and their complexity
- have removed unecessary groupings and translations
- and, they _can_ have their padding removed. 

My opinion is that removing the padding would be a good idea, and that going forward I clean-up all SVG files before we use them in a component.

HendrikThePendric:
This discussion has many different elements... not sure what to focus on:

**Why is this bug occuring**
@Mohammer5 mentions this:
> The bug you've encountered seems a bit weird to me. Normally adding position relative to an element inside a container (here: the checkbox container) should not cause the container to overlap with elements next to it.
I think this is a very valid point. Position `relative` won't normally cause this behavior. And since the pseudo element with position `absolute` is positioned against its relative parent, it should not poke out above stuff either (given the way z-index stacking works).

**Why do our SVGs have padding and where do they come from**
Like @varl says: most are just being copied from Google's `material-design-icons` repo. But as @cooper-joe mentions, these `Checkbox` and `Radio` SVGs (plus the menu-arrow-icon) are now fully owned by us so, in theory, we can do whatever we like with them. However, if we decide to remove the padding from the regular (not dense) version of the `Checkbox` and `Radio` items, we sort of "lose compatibility" with the stock material-design-icons. And this has the unfortunate consequence that Joe will have to edit every icon we incorporate before it can be used. I don't think this is a good development.

**How can this bug be fixed**
Even without changing the SVG, we can get this to look like the design specs, see https://github.com/dhis2/ui-core/pull/514/commits/33d5931e55769efe401d41b66ae74cbc3a77d683. This uses an inset box-shadow, so render a bit slower than a border. But since only one element in the document can be the `activeElement`, I don't think this is a problem.

**My preferred course of action**
- Keep the (regular) SVGs the way they are to keep "compatability with stock material design icons"
- Investigate the problem a bit more, because this seems like a problem we shouldn't be seeing
- If the problem isn't caused by something else, use the inset solution

HendrikThePendric:
Please note:
- In an ideal world I would also prefer to remove the padding from all these icons. That would be the nicest way to work for us developers.
- But I just don't like the overhead this will produce

ismay:
I would prefer to remove the padding from the icons as Joe suggested. The way I see it:

- Removing the padding is a one-time operation
- Our lib is separate from material-design-icons, though we import from them, my primary concern would be if what we're doing with our icons makes sense for us internally
- The current approach is not as straightforward as just having icons without padding, dealing with that seems more time consuming to me than a one-time operation of modifying the icons

I think the time we lose modifying the icons slightly, is offset by having a simpler approach to working with our icons.

cooper-joe:
This discussion about SVG icon usage is off topic for this PR. I've opened a new issue and quoted the relevant comments: https://github.com/dhis2/ui-core/issues/515

Let's continue the discussion there.

ismay:
So my point here is that removing the relative position gets rid of the overlap. We need the sticky behaviour here, so we cannot remove that.

So either we set a z-index, or we get rid of the relative positioning. Getting rid of the relative positioning seemed like the more general solution, so that's why I proposed this. It would address this issue for similar usecases as well, plus it would simplify working with our icons as well in my opinion.

Mohammer5:
> It is happening due to the relative positioning. See this jsfiddle: https://jsfiddle.net/wkymzu9L/1. Turning off the relative positioning fixes the overlap (on scroll), and increasing the z-index of .sticky fixes it as well.

@ismay I see, I've never encountered this behavior so far.. You never stop learning :thinking:

varl:
Going back a decade relative-absolute positioning has always been considered, at best, a workaround, and at worst, a hack.

If we can avoid using this "trick" that is for the best.

HendrikThePendric:
Closing this issue in favour of #516 which includes this approach for Checkbox and Switch

PR #516, opened 2019-10-18
refactor: checkbox and switch

HendrikThePendric:
This is the reworked version of the Checkbox, incl. the fix that gets rid of the relative and absolute positioning in the icon.

After I come back from holiday, I am planning to:
- Refactor the switch like I've done for Checkbox and Radio
- Introduce the ToggleGroup (CheckboxGroup and SwitchGroup)
- And corresponding Fields

However, I could also do so in a separate branch, so if you just want to have a version of the Checkbox without these z-index problems, you could just merge this PR into next.


HendrikThePendric:
@cooper-joe could you check if these styles are correct: https://deploy-preview-516--dhis2-ui-core.netlify.com/demo/?path=/story/checkbox--default

src/Checkbox.js
```diff
@@ -1,175 +1,192 @@
...
+                        pointer-events: none;
+                        user-select: none;
+                        margin-right: 5px;
+                        border: 2px solid transparent;
+                        padding: 1px;
+                        border-radius: 5px;
+                    }
+
+                    label.dense .icon {
+                        margin-right: 3px;
```
cooper-joe: Adding `border-radius: 4px` to `label.dense .icon` renders the correct focus border for dense.

cooper-joe:
> @cooper-joe could you check if the styles have been implemented correctly: https://deploy-preview-516--dhis2-ui-core.netlify.com/demo/?path=/story/checkbox--default

I could find only one change needed: the border-radius of focus for dense. I've left a comment for it. Great work 👍

src/Checkbox.js
```diff
@@ -1,175 +1,192 @@
 import cx from 'classnames'
 import propTypes from '@dhis2/prop-types'
-import React, { Component, Fragment } from 'react'
+import React, { PureComponent, createRef } from 'react'
```
ismay: One thing that I've noticed in ui-core is the PureComponents we're using in some places. I kind of agree with this article https://cdb.reacttraining.com/react-inline-functions-and-performance-bdff784f5578, I think we should be careful about optimizing at this stage of the lib.

Especially since we're not measuring anything to see whether it actually makes a difference, and it can cause subtle bugs.

ismay:
So the intention is for this to close #514 as well right?

src/Checkbox.js
HendrikThePendric: Good point, you're correct in thinking that this is a premature optimisation which is purely based on an assumption. We have quite a lot of components with a lot of prop, of which only a few are likely to change, but as the article you link to explains, that doesn't always mean stuff will get faster.. I think your suggestion is good: let's stick to `Component` for now, and only start optimising (i.e. using `PureComponent` etc) when we notice something.

Mohammer5: @ismay that's a very nice article you got there!

cooper-joe:
Looks good! Great job @HendrikThePendric, thanks for implementing/extending the temporary switch specs.

The only issue I saw:
- Switch is getting clipped at the right edge, by 1px, just enough to show a hard edge. This is in Firefox. In Chrome it is perhaps there, but barely noticeable.

Firefox on left, Chrome on right:
image

HendrikThePendric:
We've now ended up with the following "toggling" components that can be imported from our library:
- `Checkbox`
- `CheckboxField`
- `CheckboxGroup`
- `CheckboxGroupField`
- `Radio`
- `RadioGroup`
- `RadioGroupField`
- `Switch`
- `SwitchField`
- `SwitchGroup`
- `SwitchGroupField`

NOTE: there is no `RadioField` because a Radio is meaningless by itself

You will notice that all of the `XxField` and `XxFieldGroup` components contain very little code. This is because they had a lot in common, so they are powered by the following internal components:
- `ToggleField`
- `ToggleGroup`
- `ToggleGroupField`

Also, you will see quite a lot of nearly-identical styles in `Checkbox`, `Radio` and `Switch`. We could consider extracting this to somewhere and then just declaring the parts that are actually different in the component files. We probably prefer living with some duplication though.

src/Checkbox.js
ismay: @Mohammer5 Yeah I liked it as well, he makes a good point.

varl:
@HendrikThePendric please squash before merging.

PR #522, opened 2019-10-23
 docs(specs): components that requires positioning

cooper-joe:
First pass at specs for each component that requires positioning. A generic `position-options.feature` specs the available options/locations for positioning. This follows the [position utility discussion](https://github.com/dhis2/notes/blob/master/decisions/2019/10/15-popover-position.md).

@dhis2/team-apps, please add to and correct where necessary! I tried to keep the specs generic enough to apply to all situations, but with enough detail that they can actually be tested.

---

### Notes

I have covered the components that definitely need the position utility. There are also some 'maybe' components. These can 'reuse' the positioning in the specs, so do not have spec files.

- DropdownButton: use `menu` 
- SplitButton: use `menu`
- HeaderBar App menu: use `menu`
- HeaderBar Profile menu: use `menu`

### DIFF
--- a/cypress/integration/Menu/position.feature
+++ b/cypress/integration/Menu/position.feature
@@ -1,0 +1,51 @@
+Feature: Position of a menu component
+
+    # default max width: 380px; default max height: 380px
+    Background:
+        Given the menu component has a height and width of its default maximum
+
+    Scenario: Default rendering
+        Given there is enough space between the anchor's bottom and the body's bottom to fit the default maximum
+        When the menu is opened
+        Then the menu is below the anchor
+        And the left of the menu is aligned with the left of the anchor
+
+    Scenario: Flipped vertically
+        Given there is not enough space between the anchor's bottom and the body's bottom to fit the default maximum
+        And there is not enough space between the anchor's top and the body's top to fit the default maximum
+        When the menu is opened
+        Then the menu is above the anchor
+        And the left of the menu is aligned with the left of the anchor
+
+    Scenario: Less than 368px and more than 50px available space below and above anchor
+        Given there is not enough space between the anchor's bottom and the body's bottom to fit the default maximum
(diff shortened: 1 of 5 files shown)


cypress/integration/position/menu.feature
```diff
@@ -0,0 +1,39 @@
...
+And the menu is right aligned with the anchor
+
+Scenario: Forced body overflow
+Given the anchor has less than 128px space to the left and the right, before the limit of the <body> element
+When the menu is rendered
+Then the menu is rendered below the anchor
+And the menu is left aligned with the anchor
+And the menu overflows the body element
+// ¯\_(ツ)_/¯
+// The app would need to handle this, this is beyond the scope of the positioning utility
```
cooper-joe: I think this is beyond the scope of the position utility. Apps should have a way to deal with overflowing elements.

Mohammer5:
The folder structure is flipped, it should be:

```
cypress/integration/Menu/position.feature
cypress/integration/Popover/position.feature
cypress/integration/SelectMenu/position.feature
...
````

Have a look how I named the feature files of the AlertBar. They have the same name as the feature string itself. That makes it easier to find the right file for a certain feature. It's also what PEPFAR is doing

Mohammer5:
@cooper-joe I've corrected the indentations and added the `Background` keyword.
I'd like to change some sentences, but I won't do that on this branch. I'll prepare a PR for your branch so we can talk about it (and most likely correct my English mistakes)

Mohammer5:
@cooper-joe I've created a `Positioning` branch and [PR](https://github.com/dhis2/ui-core/pull/655) so we don't put these files on master

cooper-joe:
I've rebased this branch and squashed the commits, @Mohammer5 could you rereview before I merge?

Mohammer5:
@cooper-joe Looks good to me!

PR #527, opened 2019-10-25
chore: adjust sentences to fit a programmatic approach

Mohammer5:
I've made some changes to the 4 feature files for the components.
I tried to
* simplify the scenarios
* use more generic examples
* remove any design specific values / specs

Let me know what you think
(maybe it's good to compare side by side what I've changed)

### DIFF
--- a/cypress/integration/menu/position.feature
+++ b/cypress/integration/menu/position.feature
@@ -2,7 +2,7 @@ Feature: Position of a menu component
 
     # default max width: 380px; default max height: 380px
     Background:
-        Given the menu component has a height and height of its default maximum
+        Given the menu component has a height and width of its default maximum
 
     Scenario: Default rendering
         Given there is enough space below the anchor to fit the default maximum
--- a/cypress/integration/menu/position.feature
+++ b/cypress/integration/menu/position.feature
@@ -32,9 +32,9 @@ Feature: Position of a menu component
         And the heght of the menu is not reduced below 50px
 
     Scenario: Flipped horizontally
-        Given the anchor has less than 128px space to the left
+        Given the anchor has less than 128px space to the right
         When the menu is opened
-        Then the menu is right of the anchor
+        Then the menu is right aligned with the anchor
         And the menu is below the anchor
 
     # ¯\_(ツ)_/¯
@@ -42,5 +42,5 @@ Feature: Position of a menu component
     Scenario: Forced body overflow
         Given the anchor has less than 128px space to the left and the right
         When the menu is opened
-        Then the menu is left of the anchor
+        Then the menu is left aligned with the anchor
         And the menu is rendered below the anchor
--- a/cypress/integration/menu/position.feature
+++ b/cypress/integration/menu/position.feature
@@ -1,42 +1,46 @@
 Feature: Position of a menu component
 
+    # default max width: 380px; default max height: 380px
     Background:
-        # the default maximum
-        Given the menu component has a height of 368px
-        # also the default maximum
-        And the menu component has a width of 380px
-        And the menu component is triggered by a user click on an anchor element
+        Given the menu component has a height and height of its default maximum
 
     Scenario: Default rendering
-        Given the anchor has more than 368px space below, before the end of the <body> element
-        When the menu is rendered
-        Then the menu is rendered below the anchor
+        Given there is enough space below the anchor to fit the default maximum
+        When the menu is opened
+        Then the menu is below the anchor
         And the menu is left aligned with the anchor
 
     Scenario: Flipped vertically
-        Given the anchor has less than 368px space below, before the end of the <body> element
-        When the menu is rendered
-        Then the menu is rendered above the anchor
+        Given there is not enough space below the anchor to fit the default maximum
+        When the menu is opened
+        Then the menu is above the anchor
         And the menu is left aligned with the anchor
 
-    Scenario: Less than 368px available space below and above anchor
-        Given the anchor has less than 368px space above and below before the limits of the <body> element
-        When the menu dropdown is rendered
-        Then it is rendered below the anchor
-        And the height of the menu is reduced to fit within the <body> element
-        But the heght of the menu is not reduced below 50px (and what if it is)?****
+    Scenario: Less than 368px and more than 50px available space below and above anchor
+        Given there is not enough space above and below the anchor to fit the default maximum
+        But there is more than 50px of available space
+        When the menu is opened
+        Then the menu is below the anchor
+        And the height of the menu is reduced to fit
+
+    # ¯\_(ツ)_/¯
+    # This will cause the menu always to be off screen, but that's the apps fault
+    Scenario: Less than 50px available space below and above anchor
+        Given there is not enough space above and below the anchor to fit the default maximum
+        When the menu is opened
+        Then the menu is below the anchor
+        And the heght of the menu is not reduced below 50px
 
     Scenario: Flipped horizontally
-        Given the anchor has less than 128px space to the right, before the limit of the <body> element
-        When the menu is rendered
-        Then the menu is rendered below the anchor
-        And the menu is right aligned with the anchor
+        Given the anchor has less than 128px space to the left
+        When the menu is opened
+        Then the menu is right of the anchor
+        And the menu is below the anchor
 
+    # ¯\_(ツ)_/¯
+    # This will cause the menu always to be off screen, but that's the apps fault
     Scenario: Forced body overflow
-        Given the anchor has less than 128px space to the left and the right, before the limit of the <body> element
-        When the menu is rendered
-        Then the menu is rendered below the anchor
-        And the menu is left aligned with the anchor
-        And the menu overflows the body element
-        # ¯\_(ツ)_/¯
-        # The app would need to handle this, this is beyond the scope of the positioning utility
+        Given the anchor has less than 128px space to the left and the right
+        When the menu is opened
+        Then the menu is left of the anchor
+        And the menu is rendered below the anchor
--- a/cypress/integration/menu/position.feature
+++ b/cypress/integration/menu/position.feature
@@ -8,13 +8,13 @@ Feature: Position of a menu component
         Given there is enough space below the anchor to fit the default maximum
         When the menu is opened
         Then the menu is below the anchor
-        And the menu is left aligned with the anchor
+        And the left of the menu is aligned with the left of the anchor
 
     Scenario: Flipped vertically
         Given there is not enough space below the anchor to fit the default maximum
         When the menu is opened
         Then the menu is above the anchor
-        And the menu is left aligned with the anchor
+        And the left of the menu is aligned with the left of the anchor
 
     Scenario: Less than 368px and more than 50px available space below and above anchor
         Given there is not enough space above and below the anchor to fit the default maximum
@@ -34,7 +34,7 @@ Feature: Position of a menu component
     Scenario: Flipped horizontally
         Given the anchor has less than 128px space to the right
         When the menu is opened
-        Then the menu is right aligned with the anchor
+        Then the right of the menu is aligned with the right of the anchor
         And the menu is below the anchor
 
     # ¯\_(ツ)_/¯
@@ -42,5 +42,5 @@ Feature: Position of a menu component
     Scenario: Forced body overflow
         Given the anchor has less than 128px space to the left and the right
         When the menu is opened
-        Then the menu is left aligned with the anchor
+        Then the left of the menu is aligned with the left of the anchor
         And the menu is rendered below the anchor
--- a/cypress/integration/popover/position.feature
+++ b/cypress/integration/popover/position.feature
@@ -1,26 +1,28 @@
 Feature: Popover positioning
 
     Background:
-        Given a popover with a width of 360px and height of 200px
-        And an anchor element that renders a popover when a user clicks the anchor
+        Given the popover has a width of 360px and height of 200px
+
+    Scenario: Spacing between anchor and popover
+        When the anchor is clicked
+        Then there is some space between the anchor and the popover
 
     Scenario: Default positioning
-        Given the anchor has more than 200px above, before the limit of the <body> element
-        When the popover is rendered
+        Given the space above the anchor is more than the limit
+        When the anchor is clicked
         Then the popover is rendered above the the anchor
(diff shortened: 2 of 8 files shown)


cypress/integration/menu/position.feature
```diff
@@ -1,42 +1,46 @@
 Feature: Position of a menu component
 
+    # default max width: 380px; default max height: 380px
     Background:
-        # the default maximum
-        Given the menu component has a height of 368px
-        # also the default maximum
-        And the menu component has a width of 380px
-        And the menu component is triggered by a user click on an anchor element
+        Given the menu component has a height and height of its default maximum
```
cooper-joe: 'height and height' should be 'height and width'

cypress/integration/menu/position.feature
```diff
@@ -1,42 +1,46 @@
...
 
     Scenario: Default rendering
-        Given the anchor has more than 368px space below, before the end of the <body> element
-        When the menu is rendered
-        Then the menu is rendered below the anchor
+        Given there is enough space below the anchor to fit the default maximum
```
cooper-joe: I do prefer your version, it is much simpler. However, is it too generic? 'Enough space below the anchor' seems ambiguous to me, enough space before what? The viewport, the body, another element?

cypress/integration/menu/position.feature
```diff
@@ -1,42 +1,46 @@
...
-        Given the anchor has less than 128px space to the left and the right, before the limit of the <body> element
-        When the menu is rendered
-        Then the menu is rendered below the anchor
-        And the menu is left aligned with the anchor
-        And the menu overflows the body element
-        # ¯\_(ツ)_/¯
-        # The app would need to handle this, this is beyond the scope of the positioning utility
+        Given the anchor has less than 128px space to the left and the right
+        When the menu is opened
+        Then the menu is left of the anchor
```
cooper-joe: 'the menu is left aligned...' is the phrasing used in previous scenario, we should be consistent here.

cypress/integration/menu/position.feature
```diff
@@ -1,42 +1,46 @@
...
+        When the menu is opened
+        Then the menu is below the anchor
+        And the heght of the menu is not reduced below 50px
 
     Scenario: Flipped horizontally
-        Given the anchor has less than 128px space to the right, before the limit of the <body> element
-        When the menu is rendered
-        Then the menu is rendered below the anchor
-        And the menu is right aligned with the anchor
+        Given the anchor has less than 128px space to the left
```
cooper-joe: I think this should be 'to the right'. The default menu rendering would look like this:
image
so we should be checking available space to the right, measured from the left of the anchor I think?

cypress/integration/menu/position.feature
```diff
@@ -1,42 +1,46 @@
...
+        When the menu is opened
+        Then the menu is right of the anchor
```
cooper-joe: See comment above. I think this would result in 'Then the menu is right aligned with the anchor'.

cypress/integration/popover/position.feature
```diff
@@ -1,26 +1,28 @@
...
+        Given the popover has a width of 360px and height of 200px
+
+    Scenario: Spacing between anchor and popover
+        When the anchor is clicked
+        Then there is some space between the anchor and the popover
 
     Scenario: Default positioning
-        Given the anchor has more than 200px above, before the limit of the <body> element
-        When the popover is rendered
+        Given the space above the anchor is more than the limit
```
cooper-joe: Can we define 'the limit', perhaps in the background? It seems ambiguous to me.

cypress/integration/popover/position.feature
```diff
@@ -1,26 +1,28 @@
...
 
     Scenario: Adjusted width
-        Given the anchor has, measured from its center, less than 180px space to the right before the limit of the <body> element
-        When the popover is rendered
+        Given there is not enough horizontal space available on either side to render the popover
+        When the anchor is clicked
         Then the popover is rendered above the anchor
-        And the popover is offset by a top margin of 4px
         And the popover is centered aligned with the anchor
         And the popover width is reduced to fit in the available space
```
cooper-joe: I didn't add a scenario for 'not enough space above or below'. I think this would be up for the app to handle. Should we add a scenario for it anyway?

cypress/integration/select-menu/position.feature
```diff
@@ -1,26 +1,23 @@
...
-        # the default maximum
         Given the select menu dropdown has a height of 368px
-        # also the default maximum
         And the select menu dropdown has a width of 280px
-        And the select menu dropdown opens when the select is clicked
 
     Scenario: Default rendering
-        Given a select menu with more than 368px space below, before the end of the <body> element
-        When the select menu dropdown is rendered
+        Given the space below the select is more than its menu's height
```
cooper-joe: Again, I am unsure if 'space below' is specific enough because it doesn't define what that space is measured to.

cypress/integration/menu/position.feature
ismay: It would be nice to base ourself on the dimensions of the actual content, instead of the maximum. Which is difficult since we don't know those before rendering.

There are ways to test the dimensions, by rendering invisibly and then measuring, but that would be slower (and maybe slightly convoluted). I do think it's good to discuss this though.

Just basing ourselves on the maximum means a component with barely any content would still base itself on the maximum and might render in a different direction unnecessarily. You can't correct it afterwards, because then it'd move around, confusing the user.

cypress/integration/menu/position.feature
```diff
@@ -1,42 +1,46 @@
...
+        Given the menu component has a height and height of its default maximum
 
     Scenario: Default rendering
-        Given the anchor has more than 368px space below, before the end of the <body> element
-        When the menu is rendered
-        Then the menu is rendered below the anchor
+        Given there is enough space below the anchor to fit the default maximum
+        When the menu is opened
+        Then the menu is below the anchor
         And the menu is left aligned with the anchor
```
ismay: Should we be more precise here? `The left of the menu is aligned with the left side of the anchor`. Seems more clear to me.

ismay: Should we be more precise here? `The left of the menu is aligned with the left side of the anchor`. Seems more clear to me.

cypress/integration/menu/position.feature
```diff
@@ -1,42 +1,46 @@
...
+        Then the menu is above the anchor
         And the menu is left aligned with the anchor
 
-    Scenario: Less than 368px available space below and above anchor
-        Given the anchor has less than 368px space above and below before the limits of the <body> element
-        When the menu dropdown is rendered
-        Then it is rendered below the anchor
-        And the height of the menu is reduced to fit within the <body> element
-        But the heght of the menu is not reduced below 50px (and what if it is)?****
+    Scenario: Less than 368px and more than 50px available space below and above anchor
```
ismay: I wonder how useful this would be. We're making sure that the element doesn't render off screen by making it scrollable. But, since there will have to be scrolling anyway to reveal the content, why not render it partially offscreen and let the user scroll the entire window to reveal the rest? Gets rid of an edge case for us.

If that might cause it to render off body, then it'd probably be best to render in the other direction. And otherwise just render off body. I'd prefer that over making a the component itself scrollable.

varl:
>But it seems to me like even this is already enough work to warrant outsourcing it to an existing solution.

Even if we outsource this to an external engine, we can (and should) code the implementation of said engine to fulfill the requirements that are specified here. Then, if the engine behaviour changes through a dep update, then these tests against the specs will start failing and we have automated protection against regressions in a very important dependency.

ismay:
> > But it seems to me like even this is already enough work to warrant outsourcing it to an existing solution.
> 
> Even if we outsource this to an external engine, we can (and should) code the implementation of said engine to fulfill the requirements that are specified here. Then, if the engine behaviour changes through a dep update, then these tests against the specs will start failing and we have automated protection against regressions in a very important dependency.

Yeah that seems nice.

So but would it make sense to split the presentational parts (tooltip, popovers, etc.) from the actual positioning logic? Where we'd create the presentational components to just accept the position they should render at. And handle the actual positioning logic separately?

cypress/integration/menu/position.feature
Mohammer5: @cooper-joe Feature files shouldn't really contain design specifics, they should define behavior, which should be always the same, regardless of the design. It's not always possible to separate these two, but using pixel values is an anti pattern in my eyes. These values will be defined on the component & could change easily. The tests themselves will use these values (i. e. by reading them from the component's default props). So we're testing with real values, we're just keeping them out of the functional specs.

> The viewport, the body, another element?

This is a bit different. We can include it, I'm indecisive with this one. Let's discuss that in the meeting today?

Mohammer5: > EDIT: Ah, so I assume that here we're stating that the contents are already at least at the maximum (the menu component has a height and height of its default maximum)? Still, how would we know that before rendering?

@ismay Yes. So far the popover component I've used just uses the dimensions after initial rendering and then moved the component, no glitches so far. But if we find out there are glitches, we can just use set `visibility: hidden` and add a class name `positioned` to remove that style after its position has been determined

cypress/integration/popover/position.feature
Mohammer5: See my explanation above. We can add comments, but I wouldn't add it to the spec definitions.

cypress/integration/popover/position.feature
Mohammer5: If there's no specified behavior for it, I wouldn't add specs for it. Then it's automatically up to the app to find a solution. We could(/should?) add it to the component documentation once we're that far

cypress/integration/menu/position.feature
Mohammer5: @cooper-joe I like this, what do you think about it?

cypress/integration/menu/position.feature
Mohammer5: That's what we were doing initially. @cooper-joe preferred to keep it inside the viewport.
I also think that a bit of scrolling won't be bothering the user, while a small scrollable box is not so nice on the other hand.

ismay: Plus you have to scroll both, so there's not really a difference in the action required from the user.

cypress/integration/menu/position.feature
Mohammer5: @cooper-joe @ismay ping

cypress/integration/menu/position.feature
Mohammer5: @cooper-joe ping

cypress/integration/menu/position.feature
Mohammer5: @cooper-joe ping

Mohammer5: An addition:
When reviewing on GitHub, the window to enter the review message and type of approval (comment, approved, request changes) is more or less in a menu as well and it just stays within the body. I much prefer that over it being re-positioned, I don't mind the scrolling at all:

image

cypress/integration/menu/position.feature
cooper-joe: I'm happy with your suggestions here. The decision needed, I think is: 
> The viewport, the body, another element?
I think `<body>` is the best to use, as I am happy for a menu/other component to be positioned out of view where it can be scrolled into view, but not out of the `<body>` where it is inaccessible.

cypress/integration/menu/position.feature
cooper-joe: Yeah, that works 👍

cypress/integration/menu/position.feature
cooper-joe: Yes, I think we can leave out any kind of `viewport` measurements and just make sure these components fit within the `body`. As you said, it's easier to scroll the `body` than a small menu.

cypress/integration/menu/position.feature
Mohammer5: Done

cypress/integration/menu/position.feature
Mohammer5: So we've agreed that we want to use the body as the frame we want to position the components inside. Is it still ok to you (@cooper-joe) to keep the steps as vague as they are in this PR?

Mohammer5: ping

cypress/integration/popover/position.feature
Mohammer5: ping

Mohammer5:
@cooper-joe ping

https://github.com/dhis2/ui-core/pull/527#discussion_r353617505
https://github.com/dhis2/ui-core/pull/527#discussion_r353617646

cypress/integration/menu/position.feature
cooper-joe: I'm happy to defer that decision to someone with more experience writing effective tests but to me it is required that we state 'the body' here, or somewhere, to avoid ambiguity. If we don't state it in the test steps, how will we always remember it's to 'body' and not 'viewport'?

Mohammer5: Good point @cooper-joe. After having thought about this for a bit, I think we should include "body" or "viewport" as that's part of the behavior, and then it'll be documented as well. I'll add that to the spec files now

cypress/integration/popover/position.feature
Mohammer5: This one I def. would like to keep unclear as that's not part of the behavior.

Mohammer5:
@joe @ismay I've reworded the `Given` steps to be more precise and include the "body" as boundary for available space.

The steps' texts have gotten quite long but given the complex nature of the positioning, I think that's totally ok and better to be precise rather than optimize step text length.

cypress/integration/menu/position.feature
Mohammer5: Done, see https://github.com/dhis2/ui-core/pull/527#issuecomment-562426939

cypress/integration/popover/position.feature
Mohammer5: Done, see https://github.com/dhis2/ui-core/pull/527#issuecomment-562426939

cypress/integration/select-menu/position.feature
Mohammer5: Done, see https://github.com/dhis2/ui-core/pull/527#issuecomment-562426939

cooper-joe:
Great @Mohammer5, I think using the `body` works well 👍

PR #531, opened 2019-11-04
fix(toggles): remove position styles so element is in the document flow

HendrikThePendric:


ismay:
How about using the method from this article: https://zellwk.com/blog/hide-content-accessibly/

varl:
> 
> 
> I see that the hidden input for the checkbox with these changes now does take up a little space (and I assume for the others as well). I checked this by toggling `display: none` on and off. Is this an issue?

Yes, that's not nice.

HendrikThePendric:
Thanks to @ismay we identified this FireFox issue. It's fixed by switching to position absolute. I've squashed my commits and will proceed to merge this in.

PR #534, opened 2019-11-05
fix(tab-bar): enforce that children are instances of the Tab component

HendrikThePendric:


src/Tabs/TabBar.js
```diff
@@ -35,12 +37,15 @@ const TabBar = ({ fixed, children, className }) => (
 /**
  * @typedef {Object} PropTypes
  * @static
- * @prop {Node} children
+ * @prop {<Tab>|Array.<Tab>} children
```
varl: JSDoc complains that this is not valid?

PR #544, opened 2019-11-07
refactor(select): add conditional is required proptype check

ismay:


src/SingleSelect/Input.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import propTypes from '@dhis2/prop-types'
 import cx from 'classnames'
+import requiredIf from 'react-required-if'
```
varl: Missed one here.

ismay: Yeah I was incrementally updating and force pushing, should be all good now.

varl: Looks good to me!

PR #562, opened 2019-11-11
refactor(proptypes): loosen instanceOf checks

varl:


varl:
@Mohammer5 @HendrikThePendric  There maybe reasons to use it in higher-level libraries, and in applications.

PR #566, opened 2019-11-11
fix(select): align select with callback style changes

ismay:
This aligns the Select components with the changes in callback style. Things that aren't done yet:

#### ~~1. Property name for cb object~~ ✅ resolved (`selected`)

Currently onChange will be called with the new value for `selected`, and the event. But I understand that the intention is to use an object as the first argument. I wasn't sure what name we'd want to use there. Maybe just `selected`? Suggestions are welcome.

#### ~~2. Select option cb style~~ ✅ resolved (our style of cb)

We have to choose what to support by default for the callback style we supply to the options. We supply a callback that expects just an event, so it could be passed to an element like so: `onClick={onClick}`. I've gone with keeping that style, and am handling the exception in our MultiSelect [here](https://github.com/dhis2/ui-core/pull/566/files#diff-30592b1d577759c4462fd708b2a517cbR30). The MultiSelect option is an exception because it uses our checkbox, and thus uses the `obj, event` style.

That seems simplest to me for the user, as they can now still attach it to say, a link, and it'll work. We could also change this though, and choose to support the other style by default. Which does mean that it'd break if attached to a native element without any wrapping. What would everyone prefer here?

(Or we could change style altogether, and supply the user with our onChange cb. But that would really burden the user with logic that I think we should be handling in the lib.)

### DIFF
--- a/src/MultiSelectOption.js
+++ b/src/MultiSelectOption.js
@@ -27,7 +27,10 @@ const MultiSelectOption = ({ label, active, disabled, onClick, className }) => (
             className={checkboxClassname}
             checked={active}
             label={label}
-            onChange={onClick}
+            onChange={(value, e) => {
+                // Discarding value here because onClick just expects the event
+                onClick(e)
+            }}
             disabled={disabled}
             dense
         />
--- a/src/MultiSelectOption.js
+++ b/src/MultiSelectOption.js
@@ -27,8 +27,8 @@ const MultiSelectOption = ({ label, active, disabled, onClick, className }) => (
             className={checkboxClassname}
             checked={active}
             label={label}
-            onChange={(value, e) => {
-                // Discarding value here because onClick just expects the event
+            onChange={(data, e) => {
+                // Discarding data here because onClick just expects the event
                 onClick(e)
             }}
             disabled={disabled}
--- a/src/MultiSelectOption.js
+++ b/src/MultiSelectOption.js
@@ -27,10 +27,7 @@ const MultiSelectOption = ({ label, active, disabled, onClick, className }) => (
             className={checkboxClassname}
             checked={active}
             label={label}
-            onChange={(data, e) => {
-                // Discarding data here because onClick just expects the event
-                onClick(e)
-            }}
+            onChange={onClick}
             disabled={disabled}
             dense
         />
--- a/src/MultiSelect/Input.js
+++ b/src/MultiSelect/Input.js
@@ -21,8 +21,10 @@ const Input = ({
 }) => {
     const hasSelection = selected.length > 0
     const onClear = e => {
+        const data = { selected: [] }
+
         e.stopPropagation()
-        onChange([], e)
+        onChange(data, e)
     }
 
     return (
--- a/src/MultiSelect/Input.js
+++ b/src/MultiSelect/Input.js
@@ -22,7 +22,7 @@ const Input = ({
     const hasSelection = selected.length > 0
     const onClear = e => {
         e.stopPropagation()
-        onChange([])
+        onChange([], e)
     }
 
     return (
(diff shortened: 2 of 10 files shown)


varl:
> Maybe just `selected`? Suggestions are welcome.

I think ` { selected }` is good, or `selection` if we want to go with a noun counterpart.

Mohammer5:
I think `selected` is fine. Just like `checked` for checkboxes

HendrikThePendric:
> Currently onChange will be called with the new value for selected, and the event. But I understand that the intention is to use an object as the first argument.

I'm fine with either `selected` or `selection`. One idea I had, and this only applies to the MultiSelect, is to perhaps add one additional property: `lastSelected` / `currentSelectedItem` (naming is hard ;-)). In some contexts it might be useful to not only know the full selection but also know about the option that was clicked on last....

> The MultiSelect option **is an exception** because it uses our checkbox, and thus uses the obj, event style.

You could also argue that the ui-core components expect to work together with other ui-core components and as such the `(obj, event)` signature should be the rule, not the exception. 
I am in two minds about this because:
1. Conceptually it makes most sense to me
2. I am worried there might be a knock-on effect of us having to produce even more low-level atoms. You already mentioned using a built-in anchor as an example... I worry that we might be tempted to make counterparts of a lot of the built-in components, so they also adhere to "our" callback style.....
So yeah... not quite sure how to deal with this... pros and cons...

src/MultiSelectOption.js
```diff
@@ -27,7 +27,10 @@ const MultiSelectOption = ({ label, active, disabled, onClick, className }) => (
             className={checkboxClassname}
             checked={active}
             label={label}
-            onChange={onClick}
+            onChange={(value, e) => {
```
varl: Wouldn't the un-used `value` trigger a ESLint violation?

HendrikThePendric: I think that under the hood the ESLint rule is configured like this:
`no-unused-vars: ['error', { args: 'after-used' }
So this means that you can have unused arguments **before** the ones you are using, but not **after** the last one you've used.
https://eslint.org/docs/rules/no-unused-vars#args

ismay: Yeah as long as the last argument isn't unused it's fine.

ismay:
I went with `selected`, since that's also what's used on the select itself to accept the state.

ismay:
> You could also argue that the ui-core components expect to work together with other ui-core components and as such the (obj, event) signature should be the rule, not the exception.
I am in two minds about this because:

Yeah the thing is though, that this api isn't only expected to interact with our own (ui-core) components. It's just a possible that someone creates a custom option with a native element. So we basically have to choose, do we want the option callback to play nice with native style cb's, or our style cb's? With the current approach we have to pick one or the other.

HendrikThePendric:
> With the current approach we have to pick one or the other.

Yeah..... I think my comment mustn't have been very clear... I realise we are forced to choose one approach. My preferred option would be to treat "our" `(obj, event)` as the default, not the exception. But because I also realise that this decision can have some pretty negative consequences for combining ui-core components with built-in elements, I wasn't sure about it. Perhaps @Mohammer5 and @varl can share their opinion?

ismay:
> Yeah..... I think my comment mustn't have been very clear... I realise we are forced to choose one approach. My preferred option would be to treat "our" (obj, event) as the default, not the exception.

Ah ok I see 👍

> But because I also realise that this decision can have some pretty negative consequences for combining ui-core components with built-in elements, I wasn't sure about it. Perhaps @Mohammer5 and @varl can share their opinion?

It's not insurmountable, it's just a bit more inconvenient (plus might frustrate devs who don't know what the cb expects). If we go with the default you're suggesting, custom options that use native elements would do:

```
<a
  onClick={e => onClick({ /* empty object */}, e)}
>
  // custom option
</a>
```

ismay:
Based on @varl's suggestions in https://github.com/dhis2/ui-core/pull/430 I've refactored the callbacks that options receive to our style of callbacks. I think that that addresses all the open issues for this PR so far. Ready for review and merge!

PR #572, opened 2019-11-13
fix: introduce z-index-context to control overlapping elements

HendrikThePendric:
Before I proceed to implement this in other components, could at least one of you give me feedback on the current implementation?

I found the 16.3 context api a little quirky. This works and sort of makes sense, but perhaps there would have also been another/better way.

NB: this is a draft pull request and Edoardo's story is in there to demonstrate that this indeed fixes the problem. It should be removed before we merge this into next.

### DIFF
--- a/src/ScreenCover.js
+++ b/src/ScreenCover.js
@@ -2,7 +2,7 @@ import React from 'react'
 import propTypes from '@dhis2/prop-types'
 
 import { layers } from './theme.js'
-import { ZIndexContext, createZIndexContext } from './ZIndexContext.js'
+import { ZIndexProvider } from './ZIndexContext.js'
 
 const Backdrop = ({ onClick }) => (
     <div className="backdrop" onClick={onClick}>
@@ -51,7 +51,7 @@ Content.propTypes = {
  * @see Live demo: {@link /demo/?path=/story/screencover--default|Storybook}
  */
 const ScreenCover = ({ children, onClick, className }) => (
-    <ZIndexContext.Provider value={createZIndexContext(layers.blocking)}>
+    <ZIndexProvider value={layers.blocking}>
         <div className={className}>
             <Backdrop onClick={onClick} />
             <Content>{children}</Content>
@@ -69,7 +69,7 @@ const ScreenCover = ({ children, onClick, className }) => (
                 }
             `}</style>
         </div>
-    </ZIndexContext.Provider>
+    </ZIndexProvider>
 )
 
 /**
--- a/src/ScreenCover.js
+++ b/src/ScreenCover.js
@@ -2,7 +2,7 @@ import React from 'react'
 import propTypes from '@dhis2/prop-types'
 
 import { layers } from './theme.js'
-import { ZIndexProvider } from './ZIndexContext.js'
+import { ZIndexProvider, ZIndexConsumer } from './ZIndexContext.js'
 
 const Backdrop = ({ onClick }) => (
     <div className="backdrop" onClick={onClick}>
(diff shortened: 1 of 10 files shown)


HendrikThePendric:
I was considering the following:
- What if a modal opens another modal?
- I think in the current setup certain things could potentially break
- But what if the modal first consumes a ZIndexContext to receive its own base z-index, and then provides its own ZIndexContext based on that base z-index?
- I'm not 100% sure this is possible yet, but if so, this would cover all cases I can think of...
Worth a try?

HendrikThePendric:
OK @varl @Mohammer5, I think have finally wrapped my head around the way to set this up a bit more conveniently. Lemme know what you think of the current approach. Shall I implement that across the code-base?

And also, please let me know whether you think it would make sense for the Modal to be a Consumer as well as a Provider, as I suggested in the previous comment. I can't think of any other components in the current set of components that both "are something that pops-up" and "contain something that pops up".

src/ScreenCover.js
```diff
@@ -50,23 +51,25 @@ Content.propTypes = {
...
-                left: 0;
-                top: 0;
+                    left: 0;
+                    top: 0;
 
-                z-index: ${layers.blocking};
-            }
-        `}</style>
-    </div>
+                    z-index: ${layers.blocking};
```
varl: This should no longer be necessary.Untested comment.

varl: It is.

src/ZIndexContext.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import propTypes from '@dhis2/prop-types'
+
+const ZIndexContext = React.createContext(0)
```
varl: Wondering if we should rename this so it matches the problem domain instead of the technical domain, e.g. `LayerContext` and exporting: `LayerProvider` and `LayerConsumer`.

The prop can still be called `zIndex` in **MenuWrapper**.

src/Select/MenuWrapper.js
```diff
@@ -60,21 +61,28 @@ class MenuWrapper extends Component {
...
-                        top: ${top};
-                        left: ${left};
-                        width: ${width};
-                    }
-                `}</style>
-            </div>,
+            <ZIndexConsumer
+                propZIndex={zIndex}
+                defaultZIndex={layers.applicationTop}
+            >
```
varl: I am slightly confused by this usage. I've never seen `props` passed to the consumer this way.

Usually I would expect something like:

```
class ZIndexProvider extends React.Component {
  state = {
    zIndex: 0,
    setIndex = zIndex => this.setState({ zIndex })
  }

  render() {
    <ZIndexContext.Provider value={this.state}>
        {children}
    </ZIndexContext.Provider>
  }
}
```

```
<ZIndexConsumer>
  {({computedZIndex, setZIndex}) => {
     if (props.zIndex !== computedZIndex) {
        setZIndex(props.zIndex)
      }

    return (
      <div>...</div>
    )
  })
</ZIndexConsumer>
```

src/ZIndexContext.js
HendrikThePendric: Yes, love it.

varl:
Feel free to rebase on next and use hooks.

varl:
@HendrikThePendric I recommend taking a look at the implementation here: https://github.com/dhis2/app-runtime/tree/master/services/data/src/react

src/Select/MenuWrapper.js
HendrikThePendric: Even though I will be transitioning to hooks now, I would still like to explain how I ended up with this, perhaps unconventional, but very practical approach:
- In essence, all that the context needs to provide is the actual z-index that a component needs to build upon, so it made sense to let the context value be a simple number.
- The calculation of the z-index value for the "next layer" is based on the current z-index context, but does not actually affect the value of that context. As such it made more sense to extract that logic into the consumer rather than the provider.

I did study the docs and saw examples such as the one you provided, but I think the implementation I chose in the end is a better conceptual fit with the problem we are trying to solve. But yeah, I just made it up as I went along, I'm not following a pattern of any sorts...

HendrikThePendric: Perhaps to make what I was trying to explain a bit more tangible. In your example, there's this line:
```
setIndex = zIndex => this.setState({ zIndex }
```
But actually we don't want to mutate the z-index from that layer. We want only want built upon that. The z-index of that layer should be constant. This way we can use this in a hierarchical tree-structure.

HendrikThePendric:
Closing this one in favour of #582

PR #575, opened 2019-11-14
chore: upgrade react to support hooks

varl:
Related: #572 

This would make it easier to implement the z-index context provider in #572.

UI core overview seems to indicate that everywhere we use ui-core (atm) has hooks enabled: https://dhis2.vardevs.se/layoftheland.html

As such, this seems like a reasonable change for the `next` branch.

BREAKING CHANGE: Now requires React >= 16.8.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -6,10 +6,10 @@ update_configs:
     update_schedule: "live"
     version_requirement_updates: "increase_versions"
     ignored_updates:
-      # ui-core supports 16.3 of react, so ignore all updates
+      # ui-core supports 16.8 of react, so ignore all updates
       - match:
           dependency_name: "react"
-          version_requirement: ">=16.4.x"
+          version_requirement: ">=16.9.x"
       - match:
           dependency_name: "react-dom"
-          version_requirement: ">=16.4.x"
+          version_requirement: ">=16.9.x"
--- a/package.json
+++ b/package.json
@@ -49,18 +49,18 @@
     "cypress": "^3.6.0",
     "cypress-cucumber-preprocessor": "^1.16.2",
     "eslint-plugin-react": "^7.16.0",
-    "react": "16.3",
+    "react": "16.8",
     "react-dev-utils": "^9.1.0",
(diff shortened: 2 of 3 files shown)


.dependabot/config.yml
```diff
@@ -6,10 +6,10 @@ update_configs:
...
-      # ui-core supports 16.3 of react, so ignore all updates
+      # ui-core supports 16.8 of react, so ignore all updates
       - match:
           dependency_name: "react"
-          version_requirement: ">=16.4.x"
+          version_requirement: ">=16.9.x"
       - match:
           dependency_name: "react-dom"
-          version_requirement: ">=16.4.x"
+          version_requirement: ">=16.9.x"
```
Mohammer5: @varl why version_requirement: ">=16.9.x" version 16.9? and not 16.8?

HendrikThePendric: I know this has been discussed on Slack, but I am still confused:
1. If this is a valid configuration key, should't we flip the comparison operator to `<=16.9.x`
2. Going through the dependabot docs page it seems to me that this key is not valid. We are using `version_requirement` but in [the docs](https://dependabot.com/docs/config-file/) I can only see a reference to `version_requirement_updates`

Are we sure about this?

varl: The top level key is `ignored_updates`: https://dependabot.com/docs/config-file/#ignored_updates

Their example:

```yaml
version: 1
update_configs:
  - package_manager: "javascript"
    directory: "/"
    update_schedule: "daily"
    ignored_updates:
      - match:
          dependency_name: "express"
          # The version_requirement specifies the versions to ignore.
          # The range format is specific to the package manager
          # (e.g., ^1.0.0 for JS, or ~> 2.0 for Ruby).
          version_requirement: "4.x"
      - match:
          # Wildcards match unlimited arbitrary characters (or none)
          dependency_name: "aws*"
```

The logic is that since we want 16.8, we want to **ignore** 16.9.x, but still allow updates for any patch releases for 16.8.x.

HendrikThePendric: aha gotcha

PR #580, opened 2019-11-14
feat(select): add input max height prop

ismay:
Sets a default max-height on the input, to keep it from getting too big when there are a lot of selected options, or when a single select option contains a lot of text.

Can be overridden with `inputMaxHeight` prop.

### DIFF
--- a/src/MultiSelect/Input.js
+++ b/src/MultiSelect/Input.js
@@ -18,6 +18,7 @@ const Input = ({
     options,
     className,
     disabled,
+    inputMaxHeight,
 }) => {
     const hasSelection = selected.length > 0
     const onClear = e => {
@@ -34,7 +35,7 @@ const Input = ({
                 <InputPlaceholder placeholder={placeholder} />
             )}
             {hasSelection && (
-                <div>
+                <div className="root-input">
                     {/* the wrapper div above is necessary to enforce wrapping on overflow */}
                     <SelectionList
                         selected={selected}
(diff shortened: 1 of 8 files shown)


src/MultiSelect/Input.js
```diff
@@ -59,6 +60,12 @@ const Input = ({
                     line-height: 16px;
                 }
 
+                .root-input {
+                    max-height: ${inputMaxHeight};
+                    overflow-y: auto;
+                    flex: 1;
+                }
+
```
varl: This needs to be pulled out into a separate `<style jsx>` tag because it is dynamic. The `colors.grey900` is not because it can be resolved as a constant.

src/SingleSelect/Input.js
```diff
@@ -54,6 +55,12 @@ const Input = ({
                     line-height: 16px;
                 }
 
+                .root-input {
+                    max-height: ${inputMaxHeight};
+                    overflow-y: auto;
+                    flex: 1;
+                }
```
varl: Same as previous comment. Needs to be extracted to a separate styled-jsx tag.

src/MultiSelect/Input.js
ismay: All right, made the requested changes. Let me know if the changes aren't what you meant.

src/SingleSelect/Input.js
ismay: See https://github.com/dhis2/ui-core/pull/580#discussion_r346335238

varl:
Thanks, looks good!

PR #582, opened 2019-11-14
fix: introduce layer-context to deal with overlapping elements + story

HendrikThePendric:
Something went wrong whilst rebasing the original branch and it was much easier to start a new branch than to try and correct what had gone wrong in the other. The discussion may still be relevant and can be found here #572

### DIFF
--- a/src/AlertStack.js
+++ b/src/AlertStack.js
@@ -4,7 +4,6 @@ import propTypes from '@dhis2/prop-types'
 import cx from 'classnames'
 
 import { layers } from './theme.js'
-import { useLayerBasedZindex } from './LayerContext.js'
 
 /**
  * @module
@@ -13,10 +12,8 @@ import { useLayerBasedZindex } from './LayerContext.js'
  * @example import { AlertStack } from '@dhis2/ui-core'
  * @see Live demo: {@link /demo/?path=/story/alertstack--default|Storybook}
  */
-const AlertStack = ({ className, children, zIndex }) => {
-    const computedZIndex = useLayerBasedZindex(zIndex, layers.alert)
-
-    return createPortal(
+const AlertStack = ({ className, children }) =>
+    createPortal(
         <div className={cx(className)}>
             {children}
             <style jsx>{`
@@ -28,7 +25,7 @@ const AlertStack = ({ className, children, zIndex }) => {
                     left: 50%;
                     transform: translateX(-50%);
 
-                    z-index: ${computedZIndex};
+                    z-index: ${layers.alert};
 
                     display: flex;
                     flex-direction: column-reverse;
@@ -39,7 +36,6 @@ const AlertStack = ({ className, children, zIndex }) => {
         </div>,
         document.body
     )
-}
 
 /**
  * @typedef {Object} PropTypes
@@ -50,7 +46,6 @@ const AlertStack = ({ className, children, zIndex }) => {
 AlertStack.propTypes = {
     className: propTypes.string,
     children: propTypes.arrayOf(propTypes.element),
-    zIndex: propTypes.number,
 }
 
 export { AlertStack }
--- a/src/AlertStack.js
+++ b/src/AlertStack.js
@@ -4,6 +4,7 @@ import propTypes from '@dhis2/prop-types'
 import cx from 'classnames'
 
 import { layers } from './theme.js'
+import { useLayerBasedZindex } from './LayerContext.js'
 
 /**
  * @module
@@ -12,8 +13,10 @@ import { layers } from './theme.js'
  * @example import { AlertStack } from '@dhis2/ui-core'
  * @see Live demo: {@link /demo/?path=/story/alertstack--default|Storybook}
  */
-const AlertStack = ({ className, children }) =>
-    createPortal(
+const AlertStack = ({ className, children, zIndex }) => {
+    const computedZIndex = useLayerBasedZindex(zIndex, layers.alert)
+
+    return createPortal(
         <div className={cx(className)}>
             {children}
             <style jsx>{`
@@ -25,7 +28,7 @@ const AlertStack = ({ className, children }) =>
                     left: 50%;
                     transform: translateX(-50%);
 
-                    z-index: ${layers.alert};
+                    z-index: ${computedZIndex};
 
                     display: flex;
                     flex-direction: column-reverse;
@@ -36,6 +39,7 @@ const AlertStack = ({ className, children }) =>
         </div>,
         document.body
     )
+}
 
 /**
  * @typedef {Object} PropTypes
@@ -46,6 +50,7 @@ const AlertStack = ({ className, children }) =>
 AlertStack.propTypes = {
     className: propTypes.string,
     children: propTypes.arrayOf(propTypes.element),
+    zIndex: propTypes.number,
 }
 
 export { AlertStack }
--- a/src/LayerContext.js
+++ b/src/LayerContext.js
@@ -4,7 +4,7 @@ const LayerContext = React.createContext(0)
 const LayerProvider = LayerContext.Provider
 
 const useLayerBasedZindex = (propZIndex = 0, defaultZIndex = 0) => {
-    const contextZIndex = useContext(LayerContext) || 0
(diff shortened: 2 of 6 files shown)


src/LayerContext.js
```diff
@@ -0,0 +1,11 @@
+import React, { useContext } from 'react'
+
+const LayerContext = React.createContext(0)
+const LayerProvider = LayerContext.Provider
+
+const useLayerBasedZindex = (propZIndex = 0, defaultZIndex = 0) => {
+    const contextZIndex = useContext(LayerContext) || 0
```
varl: I don't think you need the `|| 0` as since if there is no Provider defined up the tree, the context will default to `0` as defined in `React.createContext(0)`.

HendrikThePendric:
Some thoughts about the current implementation:
- This approach is really good for working with nested layers (i.e. a Modal opening another Modal and that opens a Select)
- But could possibly cause issues when two elements with entirely different parent elements need to have a specific layering. For example:
    - When 5 layers of `ScreenCover` are open (i.e. 5 Modals on top of one-another), and the last one opens a Select, the z-index of that Select will be 5*2000+2000=12000
    - And at the same time we have an AlertStack with z-index 9999 with an AlertBar
    - Now the Select is over the Alert, and that ain't right....

One really easy way to prevent this from happening would be to edit the layers to be something like this:
```javascript
export const layers = {
    applicationTop: 5,
    blocking: 100,
    alert: 9999,
}
```
Now you'd need 20 applicationTop layers to get into trouble with blocking. And 10 blocking layers to get into trouble with the alert layer.

Alternatively we'd need a more sophisticated way to calculate the computedZIndex.

I've seen @varl and @Mohammer5 use `current + 1` types of logic in some psuedo code examples we discussed on Slack. While this would make the problem I outline above disappear, it would cause potential problems in a nested context. So I can't think of a more sophisticated way for now....

src/LayerContext.js
HendrikThePendric: Maybe it would.... In v16.3 I tried this and believe it returned `undefined` if there was no provider up the tree.... I'll give it a try.

ismay:
Relying on preventing the problem by increasing the difference between the z-indexes bothers me a bit. It does fix it in most practical cases, but on a conceptual level I'd prefer something more robust.

> While this would make the problem I outline above disappear, it would cause potential problems in a nested context. 

What would those problems be exactly?

src/LayerContext.js
HendrikThePendric: You were correct. Fixed in https://github.com/dhis2/ui-core/pull/582/commits/e3cb382a80a13951940a28027329be9ae25df869

ismay:
What I thought Jan-Gerke meant, was recursively using context to retrieve whether we're rendering at base level, or are rendering nested in a z-index provider. That way each nesting would increment the z-index. Basically going up the z-index tree like you did in your earlier suggestion, but through context.

HendrikThePendric:
> on a conceptual level I'd prefer something more robust.

From my perspective, this is very solid logic on conceptual level. In our app we deal with layers. And now we have introduced a way to layer layers on top of one-another... (Like an onion)

Examples of potential problems are hard to provide with our current component set. But I could provide a fictitious example:
1. Suppose we'd have a Modal component with z-index 100 and an Alert with z-index 200.
2. Now ModalA opens and it contains a button for opening ModalB and a button for opening AlertA
3. Alert should be on top according to the design system (see point 1)
4. But, what happens is that first AlertA opens and in the `+1` scenario this would get z-index 101. And after that the ModalB opens and gets z-index 102.
5. So now the Modal is on top of the Alert, going against the design system.

Whilst typing the above I realised how contrived the above example is....

But I actually think that on a conceptual level my `context-base-z-index + layer-z-index` makes more sense than the `+1` approach....

varl:
> What I thought Jan-Gerke meant, was recursively using context to retrieve whether we're rendering at base level, or are rendering nested in a z-index provider. That way each nesting would increment the z-index. Basically going up the z-index tree like you did in your earlier suggestion, but through context.

This is my understanding too.

HendrikThePendric:
> What I thought Jan-Gerke meant, was recursively using context to retrieve whether we're rendering at base level, or are rendering nested in a z-index provider.

Yeah... perhaps he did... I simply can't get my head around what he meant exactly.... What would I need to change?

src/AlertStack.js
```diff
@@ -12,8 +13,10 @@ import { layers } from './theme.js'
  * @example import { AlertStack } from '@dhis2/ui-core'
  * @see Live demo: {@link /demo/?path=/story/alertstack--default|Storybook}
  */
-const AlertStack = ({ className, children }) =>
-    createPortal(
+const AlertStack = ({ className, children, zIndex }) => {
+    const computedZIndex = useLayerBasedZindex(zIndex, layers.alert)
```
varl: The `alert` layer should _never_ be incremented, it is the absolute highest layer in our design system. It is the literal top.

varl: I think the alert stack warrants a hardcoded `layers.alert`.

HendrikThePendric: Yeah... I was wondering about that one whilst changing it. I've reverted these changes in https://github.com/dhis2/ui-core/pull/582/commits/3554b13bfdbd2f2b9632692941fedec9f59067ae

varl:
Ok, this would probably be good to put in a spec file but the limits are defined here: https://github.com/dhis2/design-system/blob/master/principles/spacing-alignment.md#stacking

- The alert layer (9999) is the absolute cap of our z-indexes. Nothing is ever allowed to surpass 9999.
- The application top layer range is 2000-2999 so that should be the base of the DropMenu, Select, etc. if nothing else is defined
- The blocking layer base is 3000 for Modals, etc. If the modal contains a Select or something else with a base layer of 2000, the components need to be able to get e.g. 3000+`n` per nested level.

Mohammer5:
Ok, I'll just try to lay out what I had in mind, so that it's clear:

```js
const LayerContext = createContext(0)

const useLayer = (defaultZIndex, zIndexProp) => {
  const curContext = useContext(LayerContext)

  if (zIndexProp) return zIndexProp

  // to address Viktor's concerns
  const capReached = curContext % 1000 === 999
  if (capReached) return curContext

  if (curContext) return curContext + 1

  return defaultZIndex
}

const Layer = ({ children, zIndex, defaultZIndex }) => {
  const nextZIndex = useLayer(defaultZIndex, zIndex)

  return (
    <LayerContext.Provider value={nextZIndex}>
      {children(nextZIndex)}
    </LayerContext.Provider>
  )
}

const Modal = ({ zIndex, children }) => (
  <Layer zIndex={zIndex} defaultZIndex={theme.layers.blocking}>
    {nextZIndex => (
      <aside>
        <ScreenCover style={{ zIndex: nextZIndex }}>
          {children}
        </ScreenCover>
      </aside>
    )}
  </Layer>
)

const Select = ({ zIndex, ...props }) => (
  <div>
    <InputWrapper>{props.input}</InputWrapper>
    {this.state.open && (
      <Layer zIndex={zIndex} defaultZIndex={theme.layers.application}>
        {nextZIndex => (
          <MenuWrapper zIndex={nextZIndex}>
            {menu}
          </MenuWrapper>
        )}
      </Layer>
    )}
  </div>
)
```

With this
* all components stay in their intended layer unless used in another layer.
* if there are more than one component with a layer inside a layered component, they will have the same application layer (the menus two selects in the same modal will both have a z-index of 3001)
* No component can leave the layer (there's a cap to adding 999), addressing @varl's concern
* Any component on the alert layer will still have priority over components in the blocking layer

src/LayerContext.js
```diff
@@ -1,11 +1,50 @@
...
+
+/**
+ * @typedef {Object} PropTypes
+ * @static
+ * @prop {Node} children
+ * @prop {number} zIndex
+ * @prop {number} zIndexBase
+ */
+Layer.propTypes = {
+    children: propTypes.node,
```
Mohammer5: Must be a function and needs to be required, otherwise this will cause an error

varl: Thanks. e2629ba

PR #583, opened 2019-11-14
fix(select): fix select closing on chip removal

ismay:
This needs @HendrikThePendric's PR (https://github.com/dhis2/ui-core/pull/582) in order to work properly. But with that (and thus the correct z-indexes), it works as expected. Plus it gets rid of the global click listener, which is nice.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -7,6 +7,7 @@ import {
 } from './common-prop-types.js'
 import { InputWrapper } from './Select/InputWrapper.js'
 import { MenuWrapper } from './Select/MenuWrapper.js'
+import { ScreenCover } from './ScreenCover.js'
 
 // Keycodes for the keypress event handlers
 const ESCAPE_KEY = 27
@@ -35,13 +36,10 @@ export class Select extends Component {
         }
 
         this.updateMenuPosition()
-
-        document.addEventListener('click', this.onOutsideClick)
         window.addEventListener('resize', this.updateMenuPosition)
     }
 
     componentWillUnmount() {
-        document.removeEventListener('click', this.onOutsideClick)
         window.removeEventListener('resize', this.updateMenuPosition)
 
         if (this.menuRequestId) {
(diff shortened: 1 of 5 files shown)


src/Select.js
```diff
@@ -117,24 +115,16 @@ export class Select extends Component {
...
-
-        if (!isInsideClick) {
-            if (onBlur) {
-                onBlur({ selected }, e)
-            }
-
-            this.handleClose()
+        if (onBlur) {
+            onBlur({ selected }, e)
         }
```
Mohammer5: This should be executed after `this.handleClose`, so any synchronous code called this way is not executed before the closing-logic has actually been executed.

src/Select.js
```diff
@@ -205,43 +195,47 @@ export class Select extends Component {
...
-                        maxHeight={maxHeight}
-                        selectRef={this.selectRef}
-                        menuRef={this.menuRef}
-                        menuTop={menuTop}
-                        menuLeft={menuLeft}
-                        menuWidth={menuWidth}
-                    >
-                        {menu}
-                    </MenuWrapper>
+                    <ScreenCover onClick={this.onOutsideClick} transparent />
```
Mohammer5: Why not rendering the Menu inside the ScreenCover, one less condition in the code? At least that's the intended usage of the ScreenCover.

I can see though how this works this way quite well.. so not a recommendation but rather a question

varl: If you render it inside the `ScreenCover` the `ScreenCover` tries to align it in the middle of the screen by putting the `children` in a `div` with these styles:

```
div {
    position: absolute;
    top: 50%;
    left: 50%;
    width: auto;
    height: auto;
    transform: translate(-50%, -50%);
}
```

Mohammer5: Oh, I forgot

src/Select.js
varl: I changed this in 129f14b.

PR #590, opened 2019-11-18
fix(layer-context): add support for preserving layer hierarchy in context

HendrikThePendric:
I realise that this fix only addresses a hypothetical issue given our current component set, but since we are all quite deeply into this now, it might still be a good time to tackle this....

In the current implementation, the following could hypothetically go wrong:
1. User opens modal => context z-index = 3000
2. User opens hypothetical component with `layers.blocking` from the modal => this would get z-index 3001
3. Whilst the component in (2) is open, the user opens another hypothetical component with `layers.applicationTop` from the modal => this would get z-index 3001 too

And because the components from point 2 & 3 have an equal z-index, but 3 is appended to the body later, comp3 will now be on top of comp2. And this is incorrect because comp2 has a higher layer.

After this fix comp2 would get 3002 and comp3 would get 3001 and 2 would be on top of 3 🎉 

As I see it, with this fix applied this layer context is done for the foreseeable future. We'd only have to make changes if we define more layers.

### DIFF
--- a/src/LayerContext.js
+++ b/src/LayerContext.js
@@ -1,19 +1,34 @@
 import React, { useContext } from 'react'
 import propTypes from '@dhis2/prop-types'
+import { layers } from './theme.js'
 
 const LayerContext = React.createContext(0)
+
+const getStackedLayer = (defaultLayer, context) => {
+    // Keep alert layer constant
+    if (defaultLayer === layers.alert) {
+        return layers.alert
+    }
+
+    // Differentiate between a stacked blocking and applicationTop layer
+    const layerIncrement = defaultLayer === layers.blocking ? 2 : 1
+    const layer = context + layerIncrement
+
+    // stay within stack layer boundaries defined by the design system
+    // https://github.com/dhis2/design-system/blob/master/principles/spacing-alignment.md#stacking
+    if (layer >= layers.alert) {
+        return layers.alert - 1
+    }
+
+    return layer
+}
 
 const useLayer = (defaultLayer, customLayer) => {
     const context = useContext(LayerContext)
 
     if (customLayer) return customLayer
 
-    // stay within stack layer boundaries defined by the design system
-    // https://github.com/dhis2/design-system/blob/master/principles/spacing-alignment.md#stacking
-    const capReached = context && context % 1000 === 999
-    if (capReached) return context
-
-    if (context) return context + 1
+    if (context) return getStackedLayer(defaultLayer, context)
 
     return defaultLayer
 }


Mohammer5:
I disagree with your hypothetical issue:

* If a Modal is opened within a Modal, there's no way to interact with the first Modal due to the `ScreenCover`. 
* If the app logic itself opens another application-layer component, I'd consider that a bug

HendrikThePendric:
> I disagree with your hypothetical issue:

That's because you've just made a concrete issue out of my hypothetical one ;-)

> If a Modal is opened within a Modal, there's no way to interact with the first Modal due to the ScreenCover.

I wasn't suggesting a second modal was opening, I said:
> User opens **hypothetical component** with layers.blocking from the modal

> If the app logic itself opens another application-layer component, I'd consider that a bug

That one I don't get?

Mohammer5:
> User opens hypothetical component with layers.blocking from the modal

To me this means: If the component in `ui-core` doesn't use a `ScreenCover`, it's a bug (it should be `blocking` in a literal sense)

> If the app logic itself opens another application-layer component, I'd consider that a bug

If we assume that all `blocking` components actually HAVE to use a `ScreenCover`, then the only way to open an application-layer component in the higher/first `blocking` component is through an automated trigger in the app's logic, it can't be done through user interaction alone.

ismay:
I might be running into this exactly problem with the Select (or I'm misunderstanding). When nesting a Select in a Modal, when the Selectmenu is opened, it gets z-index 3001. But there's a backdrop on top of everything that gets z-index 3002, thus preventing click access to the chips in the Selectinput.

Is that similar to the issue you're trying to solve here @HendrikThePendric?

Mohammer5:
> But there's a backdrop on top of everything that gets z-index 3002

I think the `Select` should not be using a `ScreenCover` or `Backdrop`. It's meant to prevent interaction with the rest of the page, which is `blocking-layer` behavior, not `application-layer`

ismay:
> I think the Select should not be using a ScreenCover or Backdrop. It's meant to prevent interaction with the rest of the page, which is blocking-layer behavior, not application-layer

I was just talking to Viktor about this. So the reason that the Select is using a Backdrop, is to guarantee that the menu can close on clicks outside without an even listener at the document root. And I think we want to do the same for other dropdown components, and components that need to listen to clicks outside of it.

So it seems like a valid usecase. (By the way, my assumption about chips not responding to clicks when the menu is open was incorrect. It's not a bug, that's intentional. But that's an aside)

varl:
I'm going to merge this as it works and the resulting `z-index` is more consistent.

varl:
I am also moving the `Backdrop` discussion to #430.

PR #591, opened 2019-11-18
chore: story for select with width in flex container

edoardo:
When adding `inputWidth` to the `SingleSelect`, the select overlaps the input.


ismay:
Should be fixed with the changes in #592. Once that's been published let us know if that doesn't fix it!

PR #592, opened 2019-11-18
fix(select): fix overlap on specifying input width for flex children

ismay:
Closes #591

Max-width can be used to limit flex-basis, which seems like what we want here (see: https://gedd.ski/post/the-difference-between-width-and-flex-basis). Fixes the issue demonstrated in #591.


varl:
Perfect, thanks.

PR #593, opened 2019-11-18
docs(layer context): add feature file for layer context behaviour

varl:


Mohammer5:
If there is a feature file already, I think we should add tests as well. Not only do we have tests then, it'll also guarantee that what we've been theorizing about is actually working in practice.

varl:
>If I want to have a very specific z-index, I can't set it anymore, it'll be incremented by 1 (or 2 if I choose 3000; which would be even more confusing to me as the increment is purely internal and currently not transparent).

1) It only increments by one or two if you nest layers. You can set whatever if you are not nested.

2) The `zIndex` prop only exists on the Backdrop, LayerContext, DropMenu, and MenuWrapper. None of these are exposed externally. As a consumer of the `ui-core` library, you cannot override any z-indexes through props.

3) Do we need the functionality right now or can we add it later?

varl:
I put in effort in ensuring that the z-index manipulation is entirely internal in large part because of the controversy surrounding it so that we can:

**a)** get a 4.0 version out with something that works well enough

**b)** refactor and figure out a solution we are happy with _without being stressed by getting 4.0 out_.

varl:
These are the current results:

```
Select Backdrop: 2000
    Select Menu: auto

Custom Select Backdrop: 1000
    Select Menu: auto

Modal 1: 3000
    Modal 1 Select Backdrop : 3001
        Modal 1 Select Menu: auto

Modal 2: 3002
    Modal 2 Select Backdrop : 3003
        Modal 2 Select Menu: auto

Modal 3: 3004
    Modal 3 Select Backdrop : 3005
        Modal 3 Select Menu: auto
```

varl:
>If there is a feature file already, I think we should add tests as well.

The LayerContext is best tested with Jest (unit testing).

With Cypress we should test the functional part that a Select Menu is rendered properly when in a Modal and when on the base layer. Using Cypress to test for z-indexes is the wrong tool for the job, as that would couple the end user with the implementation.

If we use Cypress to test that it behaves as we want, we can change the internal implementation without updating the tests.

PR #598, opened 2019-11-19
fix(select): fix positioning issues

ismay:
Our use of the backdrop as a parent for the menu means that the first positioned parent of the menu is now the backdrop. Since the backdrop is positioned fixed to the viewport, we need to change the measurement strategy, as now the menu will effectively be positioned relative to the viewport as well.

Since we're now positioning relative to the viewport we need to do a little more manual work in positioning the menu (previously the browser would do this for us, as we'd just supply the position relative to the document).

So I've adapted the measurement to be relative to the viewport, and am remeasuring on scroll. Eventually it would be nice to delegate this all to the browser again, as the current solution has a lot more potential for janky rendering because it's done in js.

I've also removed the Select wrapper's relative positioning, as that's no longer necessary (the fragment could also be removed because of that).

(This also means by the way that the Select menu is now very dependent on the Backdrop's implementation (its positioning). So we need to be careful to not break that, would be good to take steps to ensure that that won't happen)

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -33,17 +33,10 @@ export class Select extends Component {
         if (this.props.initialFocus) {
             this.inputRef.current.focus()
         }
-
-        this.updateMenuPosition()
-        window.addEventListener('resize', this.updateMenuPosition)
     }
 
     componentWillUnmount() {
-        window.removeEventListener('resize', this.updateMenuPosition)
-
-        if (this.menuRequestId) {
-            window.cancelAnimationFrame(this.menuRequestId)
-        }
+        this.handleMeasurementStop()
     }
 
     handleFocusInput = () => {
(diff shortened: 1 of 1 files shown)


src/Select.js
```diff
@@ -87,12 +80,30 @@ export class Select extends Component {
         })
     }
 
+    handleMeasurementStart = () => {
+        this.updateMenuPosition()
+
+        window.addEventListener('resize', this.updateMenuPosition)
+        window.addEventListener('scroll', this.updateMenuPosition)
```
varl: I don't think we need the scroll event listener.

The Backdrop stops the user from scrolling, so it will never fire.

modal-scroll

varl: (When I wiggle the mouse cursor up and down I'm using the scroll wheel.)

varl: Maybe it's different on Mac?

ismay: Strange. At first I thought maybe I was working on an outdated version of next. But I've rebased this branch on next and can still scroll with the Select open, both of FF and Chrome. I can scroll with the touchpad and with the arrow keys with the menu open.

ismay: We could either leave the scroll listeners for the Select, or if we consider it necessary for the Backdrop to stop scrolling, make sure that it stops scrolling on mac as well.

HendrikThePendric: I can also scroll, same as @ismay. What I've been reading so far about preventing scrolling effectively, is to put some blocking styles on the document body. I'd be in favour of just keeping the scroll and resize listeners for 4.0.0

varl: Interesting! I need to find someone with a Mac and see this in action.

PR #599, opened 2019-11-20
refactor(select): change input width to input max width

ismay:


ismay:
> The `SingleSelectField` and `MultiSelectField` still use `inputWidth`

Ah yeah of course. Updated.

ismay:
Also removed the stories relating to inputMaxWidth from the *Field variations of the Select, as those stories only demonstrate usage that's unique to the *Field ocmponents.

ismay:
> > The SingleSelectField and MultiSelectField still use inputWidth
> 
> Good catch, these need to be done for sure. I'm not sure what has led to this PR, so don't know which components need to have this change. But just wanted to point out that `InputField` and `TextAreaField` also still have the `inputWidth` prop.

I'll leave my slack reply here as well:

> Ah that could be, I didn't check that. My motivation for the name change are internal consistency for the Select's prop naming, and the fact that this setting functions like a max width, and not a width. (edited) 

So I think this makes sense for the Select internally, and seems like a more accurate name for the prop. I wouldn't necessarily also change this for other components, unless they also function like a max-width.

PR #600, opened 2019-11-20
feat(modal): do not render modal if open is falsy

Mohammer5:
I'm wondering if we should remove the aside as well. Although technically correct, it causes the `ScreenCover` to be rendered inside it. I think so far it hasn't caused any problems, so we can also just leave it in there until it causes problem, but it'd be a breaking change as we're removing stuff someone potentially relies upon

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -32,8 +32,8 @@ import { ModalCard } from './Modal/ModalCard.js'
  * @see Specification: {@link https://github.com/dhis2/design-system/blob/master/molecules/modal.md|Design system}
  * @see Live demo: {@link /demo/?path=/story/modal--small-title-content-action|Storybook}
  */
-export const Modal = ({ children, onClose, small, large, className }) => {
-    return createPortal(
+export const Modal = ({ children, onClose, small, large, className }) =>
+    createPortal(
         <aside className={className}>
             <ScreenCover onClick={onClose}>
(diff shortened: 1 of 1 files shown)


ismay:
> These changes make sense, given the fact that an `open` prop exists.
> 
> However, why do we need that prop in the first place? I would argue it's up to the app to render the modal, and when they do it's open.
> @ismay @varl Keep or remove `open`

Actually yeah, that's a good point. I'd be in favor of removing the prop altogether 👍 Unless there's a valid usecase for it I'm not seeing.

src/Modal.js
```diff
@@ -33,24 +32,14 @@ import { ModalCard } from './Modal/ModalCard.js'
  * @see Specification: {@link https://github.com/dhis2/design-system/blob/master/molecules/modal.md|Design system}
  * @see Live demo: {@link /demo/?path=/story/modal--small-title-content-action|Storybook}
  */
-export const Modal = ({ children, onClose, small, large, open, className }) => {
+export const Modal = ({ children, onClose, small, large, className }) => {
```
ismay: Since we're returning directly you could omit the return keyword and use parentheses.

Mohammer5:
I've changed the return-value style to the short form, removed the open prop completely & removed the `aside` wrapper, which was totally unnecessary

Mohammer5:
@varl done

PR #601, opened 2019-11-21
feat: extract centered content from screen cover

Mohammer5:
I think it's quite useful to expose the functionality of the internal `Content` component, so I've moved it to it's own component on the root level. Imagine we want to cover the content of an app but not the headerbar. Currently the `ScreenCover` covers everything.

This will make the `ScreenCover` a true molecule as it'll be a composition of `Backdrop` and `CoverCenter` instead of using its own styles.

I don't like the name I came up with too much though: `CoverCenter`
Maybe you guys have a better name?
Maybe `CoverContentCentered` (a bit too long probably...)


varl:
Do you have some place you want to use this right now or can we wait until after `next` is released? I would like to lock the scope for `next` and focus on bug fixes until we get it out the door.

Mohammer5:
I would like to use it in the new maintenance app, but there's no need to rush this at all. I don't mind waiting with this at all until `next` has been merged

Mohammer5:
Why did you close this one?

varl:
Must have been a mistake, I don't remember closing this.

On November 26, 2019 12:48:54 AM GMT+01:00, Jan-Gerke Salomon <notifications@github.com> wrote:
>Why did you close this one?
>
>-- 
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/dhis2/ui-core/pull/601#issuecomment-558393349

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

varl:
@Mohammer5 Ah, I remember now. I closed it because it has a merge base of `next` which doesn't exist anymore. So you need to open a new one to `master`.

Mohammer5:
I could've rebased to master and changed the merge base in this PR

PR #602, opened 2019-11-21
test(select): add select integration tests

ismay:
_note: this is based on next, but not necessarily to be merged in next. We can also merge it afterwards._

I was looking at implementing cypress integration tests for the Select and I found the existing naming strategy rather verbose. Personally I'd prefer a simpler, more terse naming strategy. And preferably with dashes instead of underscores (as dashes are more conventional in js codebases in my experience).

So more like this (this still uses underscores btw.):

Screenshot 2019-11-21 at 10 08 35

Instead of the current:

Screenshot 2019-11-21 at 10 08 53

With the proposed changes, when running it'll look like this:

Screenshot 2019-11-21 at 10 14 53

and

Screenshot 2019-11-21 at 10 15 12

That would be an improvement in my perspective. What does everyone else think?

### DIFF
--- a/cypress/integration/SingleSelect/opening_and_closing.feature
+++ b/cypress/integration/SingleSelect/opening_and_closing.feature
@@ -1,0 +1,35 @@
+Feature: Opening and closing the SingleSelect
+
+    Scenario: The user clicks the SingleSelect input to display the options
+        Given a closed SingleSelect with options is rendered
+        When the SingleSelect input is clicked
+        Then the options are displayed
+
+    Scenario: The user presses the down arrowkey to display the options
+        Given a closed SingleSelect with options is rendered
+        And the SingleSelect is focused
+        When the down arrowkey is pressed on the focused element
+        Then the options are displayed
+
+    Scenario: The user presses the up arrowkey to display the options
+        Given a closed SingleSelect with options is rendered
+        And the SingleSelect is focused
+        When the up arrowkey is pressed on the focused element
+        Then the options are displayed
+
+    Scenario: The user presses the spacebar to display the options
+        Given a closed SingleSelect with options is rendered
+        And the SingleSelect is focused
+        When the spacebar is pressed on the focused element
+        Then the options are displayed
+
+    Scenario: The user clicks the backdrop to hide the options
+        Given an open SingleSelect with options is rendered
+        When the user clicks the backdrop
+        Then the options are not displayed
+
+    Scenario: The user presses the escape key to hide the options
+        Given an open SingleSelect with options is rendered
+        And the SingleSelect is focused
+        When the escape key is pressed on the focused element
+        Then the options are not displayed
--- a/cypress/integration/SingleSelect/opening_and_closing/opening_and_closing.js
+++ b/cypress/integration/SingleSelect/opening_and_closing/opening_and_closing.js
@@ -1,0 +1,54 @@
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a closed SingleSelect with options is rendered', () => {
+    cy.visitStory('SingleSelect', 'With options')
+})
+
+Given('an open SingleSelect with options is rendered', () => {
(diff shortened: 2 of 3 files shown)


amcgee:
👍👍💪

Mohammer5:
Actually I think it's fine. It's really difficult to write all these long names.
My personal preference is underscores as it makes it easier to read for me (looks a bit like spaces.. kind of..)

Mohammer5:
> I thought that there would be some tool/command that would automatically do the following:
> * Create folder The_Backdrop_has_an_onClick_api (Based on Feature)

It's actually the file name that's important for the `cypress-cucumber-preprocessor` plugin. So we really don't have to name the folder after the feature text.

As for the name of the step definition files.. I'm not sure what we should do. To me it seems nice to have a file for each scenario and one for common step definitions. But how to link the file names to the scenarios when not using the names?

varl:
:heart_decoration::heart_decoration::heart_decoration::heart_decoration::heart_decoration:

Definitely on-board with shorter names, if I read two of the long ones after each other i don't remember what the first one said. :|

ismay:
> One possible consideration we could have to keep these longer names that match the strings in the .feature file 1:1, is that it offers a possibility for automating stuff in the future....

Yes. Though I think that since there's no idea if this will ever happen, making it nice to use for us now has precedence for me over the possibility of automating it in that specific way. Who knows, it might work completely different than how we imagine it now, so I'd be in favor of just doing what makes sense to us on what we know now.

> As for the name of the step definition files.. I'm not sure what we should do. To me it seems nice to have a file for each scenario and one for common step definitions. But how to link the file names to the scenarios when not using the names?

In reading the Gherkin docs (https://docs.behat.org/en/v2.5/guides/1.gherkin.html), they mention how a feature will usually have multiple scenarios. So a file per scenario seems like we'd end up with a lot of files. To me a file per feature (with the scenarios for that feature) makes sense. So like in this PR basically. Could we then still abstract common step definitions that way with our current setup @Mohammer5? Because if we could, then that'd be my preferred setup.

Mohammer5:
> To me [one] file per feature (with the scenarios for that feature) makes sense
> Could we then still abstract common step definitions that way with our current setup

Absolutely yes. Let me explain how sharing code in the step definitions can be done in our context:

1. The `cypress-cucumber-preprocessor` plugin also allows us to create `cypress/integration/common/*.js` files, which will be used for **all** tests (so we should keep these as small as possible).
1. If we need shared step definitions for just a couple of features (let's say we need shared step definitions for all `SingleSelect` features, we can create a file with these definition and simply need to import the file in the feature file's step definition file.
So if we have two features: `cypress/integration/SingleSelect/feature1.feature` and `cypress/integration/SingleSelect/feature2.feature` which share some steps, we end up with 3 folders:
   * `cypress/integration/SingleSelect/common/index.js`
   * `cypress/integration/SingleSelect/feature1/index.js`
   * `cypress/integration/SingleSelect/feature2/index.js`

   The `cypress/integration/SingleSelect/feature1/index.js` file would need the following line at the top of the file:
   ```js
   import '../common' // or import '../common/index'
   ```
1. If we decide there should be multiple files with step definition implementations for one feature file (maybe there are very large ones) and we have a `cypress/integration/Foo.feature` file, we'll need a folder `cypress/integration/Foo`. All `cypress/integration/Foo/*.js` files will be used to gather step definitions, regardless of naming. So if we decide to have multiple files for one feature, we can have a `common.js` file that includes common steps

<hr />

This is what I've tried with the very first features I added to dhis2 repos and it felt that - if the feature files has a lot of scenarios (which can easily happen) - one file for all step definitions was very hard to organize to keep it easy to have an overview. But I wouldn't mind going back there so you guys get your own experience with it (and maybe figure out a better way of formatting these files than I did)

ismay:
Nice, seems like a nice way to abstract common steps 👍

varl:
@ismay Needs a rebase on `next`.

ismay:
@varl Done, also ran it locally and tests for the single select all succeed.

PR #604, opened 2019-11-21
refactor: use a constrictor component to constrain components width

### LINKED ISSUE #605: Add min-width to input-type components
As discussed here: https://github.com/dhis2/notes/pull/83, status icons reduce the width of the input. A min-width is applied to inputs to prevent them from become unusuble when a status is shown

`input` min-width: 72px (with status icon the input results in width of 44px)
`textarea` min width: 220px (with status icon the input results in width of 192px)
`select` (single and multi-) min-width: 100px (with status icon the input results in width of 72px)

varl:
This is one of two options discussed in the meeting: https://github.com/dhis2/notes/pull/83

Resolves #605

### DIFF
--- a/src/Constrictor.js
+++ b/src/Constrictor.js
@@ -1,0 +1,41 @@
+import propTypes from '@dhis2/prop-types'
+import React from 'react'
+
+/**
+ * @module
+ *
+ * @param {Constrictor.PropTypes} props
+ * @returns {React.Component}
+ *
+ * @desc con·stric·tor | \kən-ˈstrik-tər\
+ * 1: a muscle that contracts a cavity or orifice or compresses an organ
+ * 2: a snake (such as a boa constrictor) that coils around and compresses prey
+ * 3: one that constricts
+ */
+export const Constrictor = ({ width, minWidth, maxWidth, children }) => (
+    <div>
+        {children}
+        <style jsx>{`
+            div {
+                ${width ? `width: ${width};` : ''}
+                ${minWidth ? `min-width: ${minWidth};` : ''}
+                ${maxWidth ? `max-width: ${maxWidth};` : ''}
+            }
+        `}</style>
+    </div>
+)
+
+/**
+ * @typedef {Object} PropTypes
+ * @static
+ *
+ * @prop {string} [width]
+ * @prop {string} [minWidth]
+ * @prop {string} [maxWidth]
+ */
+Constrictor.propTypes = {
+    width: propTypes.string,
+    minWidth: propTypes.string,
+    maxWidth: propTypes.string,
+    children: propTypes.node,
+}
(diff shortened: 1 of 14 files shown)


varl:
@ismay @cooper-joe I'm not going to finish the second option of having a custom input being passed in as it is becoming too complex and opening up for situations where another input type (e.g. select) can be passed into an input field, with wacky results.

ismay:
> @ismay @cooper-joe I'm not going to finish the second option of having a custom input being passed in as it is becoming too complex and opening up for situations where another input type (e.g. select) can be passed into an input field, with wacky results.

Ok yeah, seems like this is our best option then 👍

ismay:
I'll check the live storybook for this branch out (just took a look at the code) and leave a review

varl:
Maybe `Constrictor` isn't the best name for a component. 🤔

varl:
We had a meeting yesterday with @cooper-joe on how to solve this in a good way and this is the result: https://github.com/dhis2/notes/pull/83

varl:
@Mohammer5 could you take a look as well?

Mohammer5:
I'm actually wondering if we really need this component at all..

Setting a style via a prop or via css is virtually identical, as well as putting a wrapper around the components. In either case, the app will have to specify the values and add a wrapper, so I'd rather leave this to the app entirely and just add how-to guides for flex box scenarios, instead of adding complexity and components with limited flexibility to this repo and its code.

We can still add a `minWidth` prop to the control the minWidth of the containers ([as that's probably something we need to allow anyways](https://github.com/dhis2/ui-core/issues/605#issuecomment-557412874)). This will allow the app to set a higher min width on the container while a (possibly accidentally) lower min-width will have no effect (which is a desired effect)

<hr />

> @Mohammer5 could you take a look as well?

Was doing that while you wrote the comment hehe

varl:
> I'm actually wondering if we really need this component at all..

We need to allow the Label to be wider than the Input, so this will not work because it will wrap the Label and HelpText as well:

```jsx
<div style={{ width: '100px' }}>
  <InputField>
</div>
```

So we can either say to users that they have to compose their own InputField molecule with a wrapper around the Input that constricts size of the Input:

```
<Field>
  <Label>
  <div style={{ width: '100px' }}>
    <Input>
  </div>
  <HelpText />
  {{ /* and everything else /* }}
```

Which if we look at @cooper-joe's mockups for apps, we would virtually _always_ have to do because restricting the width of the Input is a tool in UX to signal intent on how to use the fields.

Since the use case is so common and an integral technique to use for design, we think it makes sense to support it as a first-class option.

This way we also allow the default behaviour of our Input, Select, TextArea, etc. to always flow to 100%, so this change only affects the *Field molecules.

PR #607, opened 2019-11-22
feat: transfer feature files

cooper-joe:
`.feature` files for the [Transfer](https://github.com/dhis2/design-system/blob/master/organisms/transfer.md) component.

I have tried to keep these feature files strictly functional. There is some repetition in some of the scenarios where two functionalities are very similar (e.g. reordering up or down) but I felt they should be tested independently. There are some comments throughout the files, please take a look and provide feedback.

### DIFF
--- a/cypress/integration/transfer/disabled-transfer-buttons.feature
+++ b/cypress/integration/transfer/disabled-transfer-buttons.feature
@@ -1,0 +1,19 @@
+Feature: Disable transfer buttons when actions are not possible
+
+  Scenario: No highlighted items in options list
+      Given the options list has items
+      When no items are highlighted
+      Then the 'move to selected list' button is disabled
+
+  Scenario: No highlighted items in selected list
+      Given the selected list has items
+      When no items are highlighted
+      Then the 'move to options list' button is disabled
+
+  Scenario: No items in options list
+      Given the options list does not have items
+      Then the 'move all to selected list' button is disabled
+
+  Scenario: No items in selected list
+      Given the selected list does not have items
+      Then the 'move all to options list' button is disabled
--- a/cypress/integration/transfer/display-order.feature
+++ b/cypress/integration/transfer/display-order.feature
@@ -1,0 +1,19 @@
+Feature: Display order of items in lists
+
+  #Trying to say: the options list does not manipulate the order in any way: is this an redundant scenario? I suppose 'not messing with the order' seems like a feature because the selected list actively changes the order.
+  Scenario: The transfer renders the options in the order they're given in
+    Given the options list has one or more items
+    Then the display order of the options matches the order the options are given in
+
+  #I can't get this one quite right. I'm trying to test for the options list always maintaining its original ordering, even when items are passed back to it.
+  Scenario: Items in options list resume order on transfer
+    Given the options list has one or more items
+    And a selected list that has at two or more items
+    When an item is transferred from the selected list to the options list
+    Then the item is rendered in the options list in the original order the options were given in
+
+  Scenario: Items in selected list ordered by last added
+    Given the options list has one or more items
+    And a selected list that has one or more item
+    When an item is transferred from the options list to the selected list
+    Then the item is rendered at the bottom of the selected list
--- a/cypress/integration/transfer/drag-and-drop-items.feature
+++ b/cypress/integration/transfer/drag-and-drop-items.feature
@@ -1,0 +1,25 @@
+Feature: Drag and drop selected items
+
+  Background:
+    Given reordering is enabled
+
+  # I have excluded any design information from this scenario: opacity, how the items look when dragging etc.
+  Scenario: Drag and drop a single selected item
+    Given the selected list has two or more items
+    When an item is clicked and dragged
+    Then the item is removed from its original position in the list
+    And the item is placed in the list at the cursor position when the click is released
+
+  Scenario: Drag and drop two or more selected items
+    Given the selected list has three or more items
+    And two or more highlighted items
+    When one of the highlighted items is clicked and dragged
+    Then the items are removed from their original positions
+    And the items are placed in the list at the cursor position when the click is released
+    And the items are ordered based on their previous ordering
+
+  Scenario: Disable reordering with a single item
+    Given the selected list has one item
+    And one highlighted item
+    When the highlighted item is clicked and dragged
+    Then the highlighted item remains in position and does not move
--- a/cypress/integration/transfer/filter-options-list.feature
+++ b/cypress/integration/transfer/filter-options-list.feature
@@ -1,0 +1,14 @@
+Feature: Filter options list
+
+  Background:
+    Given filtering is enabled
+    And the options list is being filtered
+
+  Scenario: Filtering with results
+    When the result is not empty
+    Then all the matching items are shown in the options list
+
+  Scenario: Filtering without results
+    When the result is empty
+    Then no items are shown in the options list
+    And information is displayed that no items matched the filter
--- a/cypress/integration/transfer/highlight-items.feature
+++ b/cypress/integration/transfer/highlight-items.feature
@@ -1,0 +1,43 @@
+Feature: Highlight item in lists
+
+  Scenario Outline: Highlight a single item on mouse click
+    Given the <type> list has one or more items
+    When an item is clicked
+    Then it is highlighted
+
+    Examples:
+        | type      |
+        | option    |
+        | selected |
+
+  Scenario Outline: Highlight an item with keyboard
+    Given the <type> list has one or more items
+    And an item that has keyboard focus
+    When the Enter key or Space is pressed
+    Then the focussed item is highlighted
+
+    Examples:
+        | type      |
+        | option    |
+        | selected |
+
+  Scenario Outline: Highlight multiple items using CMD/CTRL+click
+    Given the <type> list has two or more items
+    When multiple items are CMD/CTRL+clicked
+    Then all of the clicked items are highlighted
+
+    Examples:
+        | type      |
+        | option    |
+        | selected |
+
+  Scenario Outline: Highlight multiple items using SHIFT+click
+    Given the <type> list has two or more items
+    And a highlighted item
+    When any other item is SHIFT+clicked
+    Then all items between the highlighted item and the SHIFT+clicked item are highlighted, inclusive of the SHIFT+clicked item
+
+    Examples:
+        | type      |
+        | option    |
+        | selected |
--- a/cypress/integration/transfer/reorder-with-buttons.feature
+++ b/cypress/integration/transfer/reorder-with-buttons.feature
@@ -1,0 +1,35 @@
+Feature: Reorder items in the selected list
+
+  Background:
+    Given reordering of items is enabled
+
+# I created two scenarios for up and down reordering because I wanted to incorporate a check for 'is first' or 'is last' and it seemed overly ambiguous to write this in a single scenario. Or, is it enough that I have included the 'Disable move buttons' scenarios at the bottom of this file, so that doesn't need to be handled in the basic scenarios at the top?
+
+  Scenario Outline: Reorder items upwards using buttons
+      Given the selected list has two or more items
+      And the <previous> item is highlighted
+      When the 'move up' button is pressed
+      Then the highlighted item is moved to the <next> place
+
+      Examples:
+          | previous | next |
+          | 1        | 1    |
+          | 2        | 1    |
+          | 3        | 2    |
+
+  Scenario Outline: Reorder items upwards using buttons
+      Given the selected list has three items
+      And the <previous> item is highlighted
+      When the 'move down' button is pressed
+      Then the highlighted item is moved to the <next> place
+
+      Examples:
+          | previous | next |
+          | 1        | 2    |
+          | 2        | 3    |
+          | 3        | 3    |
+
+  Scenario: Disable reorder buttons when no items highlighted
+      Given the selected list has two or more items
+      When no items are highlighted in the list
+      Then the 'move up' and 'move down' reorder buttons are disabled
--- a/cypress/integration/transfer/transferring-items.feature
+++ b/cypress/integration/transfer/transferring-items.feature
@@ -1,0 +1,38 @@
+Feature: Transferring items between lists
+
+  Scenario: Transfer from options list to selected list
+    Given one or more items in the options list are highlighted
+    When the 'move to selected list' button is clicked
+    Then the highlighted items are removed from the options list
+    And the highlighted items are visible in the selected list
+
+  Scenario: Transfer from selected list to options list
+    Given one or more items in the selected list are highlighted
+    When the 'move to options list' button is clicked
+    Then the highlighted items are removed from the selected list
+    And the highlighted items are visible in the options list
+
+  Scenario: Transfer all from options list
+    Given the options list has one or more items
+    When the 'move all to selected list' button is clicked
+    Then all items are removed from the options list
+    And all items removed from options list are visible in the selected list
+
+  Scenario: Transfer all from selected list
+    Given the selected list has one or more items
+    When the 'move all to options list' button is clicked
+    Then all items are removed from the selected list
+    And all items removed from selected list are visible in the options list
+
+  Scenario: Transfer item with double click
+    When an item is double clicked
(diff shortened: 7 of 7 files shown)


cypress/integration/transfer/reorder-with-buttons.feature
```diff
@@ -0,0 +1,35 @@
...
+    Given reordering of items is enabled
+
+# I created two scenarios for up and down reordering because I wanted to incorporate a check for 'is first' or 'is last' and it seemed overly ambiguous to write this in a single scenario. Or, is it enough that I have included the 'Disable move buttons' scenarios at the bottom of this file, so that doesn't need to be handled in the basic scenarios at the top?
+
+  Scenario: Reorder items upwards using buttons
+    Given a selected list with one or more items
+    And a highlighted item in the selected list
+    And the highlighted item is not the first item in the list
+    When the 'move up' button is clicked
+    Then the highlighted item moves up one place in the selected list
```
Mohammer5: I think we can modify this one and and include what happens when the `move up` is pressed when the first item is highlighted.

```feature
Scenario Outline: Reorder items upwards using buttons
    Given a selected list with one or more items
    And the <previous> item is highlighted
    When the 'move up' button is clicked
    Then the highlighted item is moved to the <next> place

    Examples:
        | previous | next |
        | 1        | 1    |
        | 2        | 1    |
        | 3        | 2    |
```

cypress/integration/transfer/reorder-with-buttons.feature
```diff
@@ -0,0 +1,35 @@
...
+    Given a selected list with one or more items
+    And a highlighted item in the selected list
+    And the highlighted item is not the last item in the list
+    When the 'move down' button is clicked
+    Then the highlighted item moves down one place in the selected list
+
+  Scenario: Disable reorder buttons when no items highlighted
+    Given a selected list with one or more items
+    When no items are highlighted in the list
+    Then the 'move up' and 'move down' reorder buttons are disabled
```
Mohammer5: `When` should be `Given`/`And`

The buttons themselves are not disabled. The user can still press them, so I think it's better to change the `Then` to

```feature
When the `move up` is pressed
Then the order remains the same
```

And then use `Scenario Outline` to make sure both buttons are being tested

cypress/integration/transfer/reorder-with-buttons.feature
```diff
@@ -0,0 +1,35 @@
...
+
+  Scenario: Disable 'move up' button when highlighted item is first
+    Given a selected list with at least one item
+    When a highlighted item is first in the list
+    Then the 'move up' button is disabled
```
Mohammer5: `When` should be `Given` or `And`.
As above, I'd change the `Then` step here in the same way.
The button is not disabled, it's press event is simply ignored.

cypress/integration/transfer/reorder-with-buttons.feature
```diff
@@ -0,0 +1,35 @@
...
+
+  Scenario: Disable 'move down' button when highlighted item is last
+    Given a selected list with at least one item
+    When a highlighted item is last in the list
+    Then the 'move down' button is disabled
```
Mohammer5: Same here

cypress/integration/transfer/transferring-items.feature
```diff
@@ -0,0 +1,30 @@
+Feature: Transferring items between lists
+
+  Background:
```
Mohammer5: Unnecessary background

cypress/integration/transfer/transferring-items.feature
```diff
@@ -0,0 +1,30 @@
+Feature: Transferring items between lists
```
Mohammer5: What happens when items are doubled clicked while another one or multiple are selected?

Mohammer5:
> There is some repetition in some of the scenarios where two functionalities are very similar (e.g. reordering up or down) but I felt they should be tested independently. 

That's actually good!

cypress/integration/transfer/disabled-transfer-buttons.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Disable transfer buttons when actions are not possible
+
+  Scenario: Disable 'move to selected list' button
+    Given an options list with items
+    When no items are highlighted
+    Then the 'move to selected list' button is disabled
```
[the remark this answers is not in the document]
cooper-joe: This makes sense, thanks for the explanation.

I've updated all the scenarios here to reflect the context. I've also removed the references to buttons become disabled and instead recorded what should happen if those buttons are pressed (as you mentioned here: https://github.com/dhis2/ui-core/pull/607#discussion_r349565716).

cypress/integration/transfer/disabled-transfer-buttons.feature
```diff
@@ -0,0 +1,19 @@
...
+
+  Scenario: Disable 'move to options list' button
+    Given a selected list with items
+    When no items are highlighted
+    Then the 'move to options list' button is disabled
+
+  Scenario: Disable 'move all to selected list' button
+    When there are no items in an options list
```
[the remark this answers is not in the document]
cooper-joe: Fixed

cypress/integration/transfer/display-order.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Display order of items in lists
+
+  #Trying to say: the options list does not manipulate the order in any way: is this an redundant scenario? I suppose 'not messing with the order' seems like a feature because the selected list actively changes the order.
+  Scenario: Display items in options list
```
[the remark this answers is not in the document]
cooper-joe: Fixed

cypress/integration/transfer/display-order.feature
```diff
@@ -0,0 +1,19 @@
...
+
+  #I can't get this one quite right. I'm trying to test for the options list always maintaining its original ordering, even when items are passed back to it.
+  Scenario: Items in options list resume order on transfer
+    Given an options list one or more items
+    And a selected list with at least one item
+    When an item is transferred from the selected list to the options list
+    Then the item is rendered in the options list in the original order
+
+  Scenario: Items in selected list ordered by last added
+    Given an options list with at least one item
```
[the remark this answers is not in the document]
cooper-joe: Thanks for noticing, I updated these everywhere to be consistent.

cypress/integration/transfer/display-order.feature
```diff
@@ -0,0 +1,19 @@
...
+
+  #Trying to say: the options list does not manipulate the order in any way: is this an redundant scenario? I suppose 'not messing with the order' seems like a feature because the selected list actively changes the order.
+  Scenario: Display items in options list
+    Given an options list with one or more items
+    Then the display order of the options matches the order passed in
+
+  #I can't get this one quite right. I'm trying to test for the options list always maintaining its original ordering, even when items are passed back to it.
+  Scenario: Items in options list resume order on transfer
+    Given an options list one or more items
+    And a selected list with at least one item
```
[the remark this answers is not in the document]
cooper-joe: Fixed

cypress/integration/transfer/display-order.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Display order of items in lists
```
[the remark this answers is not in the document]
cooper-joe: I think that is covered in this scenario?

```
  Scenario: Items in options list resume order on transfer
    Given an options list that has one or more items
    And a selected list that has at two or more items
    When an item is transferred from the selected list to the options list
    Then the item is rendered in the options list in the original order the options were given in
```

cypress/integration/transfer/drag-and-drop-items.feature
```diff
@@ -0,0 +1,19 @@
...
+  # I have excluded any design information from this scenario: opacity, how the items look when dragging etc.
+  Scenario: Drag and drop a single selected item
+    Given a selected list with one or more item
+    When a user clicks and drags an item
+    Then the item is removed from its original position in the list
+    And the item is placed in the list where the user releases the mouse cursor
+
+  Scenario: Drag and drop multiple selected items
+    Given a selected list with one or more item
+    And one or more highlighted items
```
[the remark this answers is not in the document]
cooper-joe: fixed

cypress/integration/transfer/filter-options-list.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Filter options list
+
+  Background:
+    Given filtering is enabled
+
+  Scenario: Filtering with results
+    Given the user filters the options list
```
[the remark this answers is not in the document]
cooper-joe: I've removed references to the user throughout all the .feature files, I think it makes sense to write these from a technical perspective.

cypress/integration/transfer/highlight-items.feature
```diff
@@ -0,0 +1,23 @@
+Feature: Highlight item in lists
+
+  Scenario: Highlight an single item on mouse click
+    Given an option list or selected list with at least one item
```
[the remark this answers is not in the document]
cooper-joe: Awesome! I've added these scenario outlines rather than splitting the file.

cypress/integration/transfer/reorder-with-buttons.feature
cooper-joe: Really cool use of a Scenario Outline!

cypress/integration/transfer/reorder-with-buttons.feature
cooper-joe: I've updated this, but I just want to check we are on the same page.
> The button is not disabled, it's press event is simply ignored.
The button should be disabled (as in, use `<Button disabled>`. Are you saying that this information should not be included in a feature file?

cypress/integration/transfer/reorder-with-buttons.feature
cooper-joe: Fixed

cypress/integration/transfer/transferring-items.feature
cooper-joe: The item is transferred as normal. I didn't create a scenario for this because I thought it was covered in the other scenario by *not being mentioned*. Do you think it should be have its own scenario?

cooper-joe:
Thanks for the thorough review @Mohammer5, much appreciated. I've rewritten a lot of the scenarios now and added a couple of follow-up comments.

cypress/integration/transfer/display-order.feature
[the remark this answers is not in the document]
Mohammer5: Oops, didn't see that, my bad

cypress/integration/transfer/reorder-with-buttons.feature
Mohammer5: Oh, no. I though you were talking about the "page up" keyboard buttons. In that case you scenarios are actually correct! Sorry 'bout that

cypress/integration/transfer/transferring-items.feature
Mohammer5: I def. think we should add any behavior. Anything that's not documented will be either unclear (and therefore the dev will come up with his/her own solution) or it'll be simply buggy until a user notices it and reports a bug

cypress/integration/transfer/disabled-transfer-buttons.feature
```diff
@@ -0,0 +1,23 @@
+Feature: Disable transfer buttons when actions are not possible
+
+  Scenario: No highlighted items in options list
+    Given an options list that has items
```
Mohammer5: This (and others) still lack verbs:
```feature
Given the options list has items
```

cypress/integration/transfer/disabled-transfer-buttons.feature
```diff
@@ -0,0 +1,23 @@
...
+    And none of the items in the options list are highlighted
+    When the 'move to selected list' button is pressed
```
Mohammer5: Actually, now that I realize that the "move up" button is disabled, I'm wondering why the "move to selected list" button is not disabled. Same for the other scenarios in this feature file (which is actually called "disabled transfer buttons"

cypress/integration/transfer/drag-and-drop-items.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Drag and drop selected items
```
Mohammer5: I'm wondering if it makes sense to allow the reordering functionality if there's only one item.
We could change the first scenario to `two or more` and add an additional scenario that states that reordering is not available if there's only one item in the list. The currently second scenario would need to change to `three or more` then to make sense.

cypress/integration/transfer/reorder-with-buttons.feature
```diff
@@ -0,0 +1,41 @@
...
+      Given a selected list that has two or more items
+      And the <previous> item is highlighted
+      When the 'move down' button is pressed
+      Then the highlighted item is moved to the <next> place
+
+      Examples:
+          | previous | next |
+          | 1        | 2    |
+          | 2        | 3    |
+          | 3        | 3    |
```
Mohammer5: In this case the first `Given` needs to be specific: `Given the selected list has three items`

cypress/integration/transfer/reorder-with-buttons.feature
```diff
@@ -0,0 +1,41 @@
...
+
+  Scenario: Attempting to reorder without highlighted items
```
Mohammer5: I think we should change this scenario to just state that these buttons are disabled when no item is highlighted.

cypress/integration/transfer/transferring-items.feature
```diff
@@ -0,0 +1,28 @@
...
+
+  Scenario: Transfer from options list (left) to selected list (right)
```
Mohammer5: I think we should omit the sides. Firstly, that's part of the design, not part of the behavior. Secondly, if this is important, it should be part of the steps, not of the scenario title (which won't be tested) and thirdly this can potentially change (if we add a mobile design, and we simply display the selected list below the options list but don't change any behavior, we should not have to change this feature file)

cypress/integration/transfer/reorder-with-buttons.feature
cooper-joe: This is technically correct, as per your comment here https://github.com/dhis2/ui-core/pull/607#discussion_r349965919, so I'll revert this back to the original scenario.

cypress/integration/transfer/reorder-with-buttons.feature
cooper-joe: Oh, oops! No problem, that's good to know. I'll revert the changes I made for that.

cypress/integration/transfer/disabled-transfer-buttons.feature
cooper-joe: I prematurely removed those scenario when I thought it was wrong to reference buttons being disabled 😅 I've added these back in now.

cypress/integration/transfer/drag-and-drop-items.feature
cooper-joe: Agreed, it's best to be specific and cover all the use cases. I've modified the existing scenarios and added a new one for disabling reordering for a single item.

cypress/integration/transfer/reorder-with-buttons.feature
cooper-joe: Fixed

cypress/integration/transfer/transferring-items.feature
cooper-joe: Good point, fixed.

cypress/integration/transfer/disabled-transfer-buttons.feature
cooper-joe: Fixed.

cypress/integration/transfer/transferring-items.feature
cooper-joe: Agreed, I've added a scenario for this:
```
  Scenario: Double click an item when others are highlighted
    Given the options or selected list has two or more items
    And one or more items are highlighted
    When an item is double clicked
    Then the item is removed from its current list
    And the item is visible in the other list
    And the previously highlighted items are no longer highlighted
```

cypress/integration/transfer/reorder-with-buttons.feature
cooper-joe: Reverted to original scenario to reflect that the buttons are actually disabled.

PR #618, opened 2019-11-25
WIP: run cypress on travis

ismay:
Just to make sure the integration tests pass before merging, and to ensure they don't go out of date. Current config is just a sketch, to get it to work first.


varl:
@ismay I would like us to run this on GitHub Actions instead of Travis. Working implementation in #625.

PR #624, opened 2019-11-26
test(select): add select integration tests

ismay:
Improving the test coverage for the Select. Since it's such an intricate component it'd be good to fully test all the features. Note that some selectors used in the tests are a bit fragile. I'd love to use `data-test-*` attributes to make these more robust, but from what I can gather we've not arrived at a decision, so consider this an intermediate solution.

(I'll move the todo list below to the new PR once this has been merged)

Todo regular:

- [ ] clicking disabled options
- [ ] onBlur cb
- [ ] onFocus cb
- [ ] Placeholder
- [ ] Prefix
- [ ] Loading text
- [ ] Max-height
- [ ] Empty
- [ ] Initial focus
- [ ] Disabled
- [ ] Custom options

Todo *Field:

- [ ] Label
- [ ] Statuses
- [ ] Help text
- [ ] Required

### DIFF
--- a/cypress/integration/MultiSelect/clearing/index.js
+++ b/cypress/integration/MultiSelect/clearing/index.js
@@ -1,0 +1,17 @@
+import '../common'
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a clearable MultiSelect with a selection is rendered', () => {
+    cy.visitStory('MultiSelect', 'With clear button and selection')
+})
+
+When('the clear button is clicked', () => {
+    cy.contains('Clear').click()
+})
+
+Then('the MultiSelect is cleared', () => {
+    cy.window().then(win => {
+        expect(win.onChange).to.be.calledOnce
+        expect(win.onChange).to.be.calledWith({ selected: [] })
+    })
+})
--- a/cypress/integration/MultiSelect/filtering/index.js
+++ b/cypress/integration/MultiSelect/filtering/index.js
@@ -1,0 +1,24 @@
+import '../common'
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a filterable MultiSelect with options is rendered', () => {
+    cy.visitStory('MultiSelect', 'With filter field')
+})
+
+When('the user enters a filter string', () => {
+    cy.focused().type('one')
+})
+
+When("the user enters a filter string that doesn't match any options", () => {
+    cy.focused().type('does not exist')
+})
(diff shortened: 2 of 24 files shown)


cypress/integration/MultiSelect/clearing/index.js
```diff
@@ -0,0 +1,18 @@
+import '../common'
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a clearable MultiSelect with a selection is rendered', () => {
+    cy.visitStory('MultiSelect', 'With clear button and selection')
+    cy.window().then(win => (win.onChange = cy.stub()))
```
Mohammer5: This is already set in the `Given an onChange handler is attached` step

cypress/integration/MultiSelect/filtering/index.js
```diff
@@ -0,0 +1,24 @@
...
+When('the user enters a filter string', () => {
+    cy.focused().type('one')
+})
+
+When("the user enters a filter string that doesn't match any options", () => {
+    cy.focused().type('does not exist')
+})
+
+Then('the matching options are displayed', () => {
+    cy.contains('option one')
```
Mohammer5: To make this more clear when reading the code, you could add `.should('exist')`.
Then it'll show up as an expectation in the cypress ui as well

cypress/integration/SingleSelect/common/index.js
```diff
@@ -0,0 +1,27 @@
...
+Given('an onChange handler is attached', () => {
+    cy.window().then(win => (win.onChange = cy.stub()))
+})
+
+Given('the SingleSelect is open', () => {
+    cy.get('.select [tabIndex="0"]').click()
+
+    cy.contains('option one')
+    cy.contains('option two')
+    cy.contains('option three')
```
Mohammer5: I'd add the `.should('exist')` here as well

cypress/integration/MultiSelect/clearing/index.js
ismay: Ah yeah good catch, I missed that one

cypress/integration/MultiSelect/filtering/index.js
ismay: Ah I didn't know that. Changed 👍

cypress/integration/SingleSelect/common/index.js
ismay: Ok, done

PR #626, opened 2019-11-27
chore: upgrade d2-style, reformat, and remove travis

varl:


varl:
The Travis check is not going to pass, so I will have to force merge this after approval.

varl:
Most, if not all, of the code changes should be due to the automatic sorting of propTypes.

Mohammer5:
> The Travis check is not going to pass

@varl Can you explain this? Why should we merge a change that makes the build fail?

varl:
> 
> 
> > The Travis check is not going to pass
> 
> @varl Can you explain this? Why should we merge a change that makes the build fail?

Because I deleted the `.travis.yml` file. The Travis checks are replaced by GitHub Actions checks.

Mohammer5:
All good then, just wondering: why is travis checking this PR then?

varl:
@Mohammer5 it's configured as a required check in GitHub, so GitHub sends the webhook to Travis, but Travis cannot find the config file, so it just stalls.

PR #627, opened 2019-11-27
test(select): finalize select integration tests

ismay:
Todo regular:

- [x] clicking disabled options
- [x] Placeholder
- [x] Prefix
- [x] Disabled
- [x] Max-height
- [x] Loading text
- [x] Empty
- [x] Initial focus
- [x] Invalid options (displays them regularly, unless filtering)
- [x] Custom options
- [x] Align naming with other integration tests (there's some divergence)

Todo *Field:

- [x] Label
- [x] Help text
- [x] Required
- [x] Statuses

Notes:

- The generated content `*` for required fields is more difficult to test reliably than if it'd be regular text content
- The required field `*` does not show up if there is not a label. The label is marked as required in the jsdoc, but not in the proptypes. Should the label be required or not? And do we not want the required status to show up if there is not a label? Update tests and source to match our expectations.
- I would prefer to add the tests for the onBlur and onFocus cbs separately. I've already noticed that they'll probably need some work, and think it's better to separate that from this PR.


ismay:
> All tests are passing (I had some problems with the AlertBar, but could not reproduce them) and all feature files are looking good.
> 
> the only thing I noticed is that we are using the `should` assertion here, and in the bulk of the existing tests we are using `expect`. I'd prefer to stick to the latter.

Yeah you're right. I've been a bit inconsistent. I'll modify them to fit the existing style 👍

ismay:
So I've modified the file naming to be more of a match with the existing convention. However, I'm not sure I prefer moving the DOM node assertions to use expect instead of should. Here's the difference:

```
    // with should
    cy.get('div[role="progressbar"]').should('be.visible')

    // with expect
    cy.get('div[role="progressbar"]').then($el => {
        expect($el).to.be.visible
    })
```

I find the first version to be a lot more simple, and readable. As far as I know, there's no functional difference.

In looking through the codebase with regards to your comment @HendrikThePendric, I noticed that the expect assertions are mostly used when not asserting anything in the DOM, like visibility of a dom node (or something related to that), but when asserting whether a function has been called, etc. I'm doing that as well.

Likewise, I see that should is also used in other tests when asserting DOM related things. I think that that's the explanation for the difference in should vs expect assertions, because I'm testing a lot of DOM related things with the select.

Mohammer5:
I agree with that, using the should assertions is a lot easier than the expect assertions

ismay:
By the way @HendrikThePendric, let me know if you are ok with the way the tests are currently written.

PR #630, opened 2019-11-28
Add file input tests

Mohammer5:
* [x] Add tests for single-file `FileInput`'s onChange handler
* [x] Add tests for multi-file `FileInput`'s onChange handler

@HendrikThePendric Please check if the second scenario in the `Accepts_multiple_files.feature` is how this component is supposed to work, or if selecting files after some have been selected, should override the existing ones

### DIFF
--- a/cypress/fixtures/FileInput/file.md
+++ b/cypress/fixtures/FileInput/file.md
@@ -1,0 +1,1 @@
+# I'm a markdown file!
--- a/cypress/fixtures/FileInput/file.txt
+++ b/cypress/fixtures/FileInput/file.txt
@@ -1,0 +1,1 @@
+I'm a text file!
--- a/cypress/integration/FileInput/Accepts_multiple_files.feature
+++ b/cypress/integration/FileInput/Accepts_multiple_files.feature
@@ -1,0 +1,17 @@
+Feature: The FileInput can handle multiple files
+
+    Scenario: The user selects multiple files
+        Given a FileInput that accepts multiple files is rendered
+        And the FileInput does not have any files
+        And the FileInput is provided with an onChange handler
+        When the user selected multiple files
+        Then the onChange handler is called
+        Then the onChange handler's payload contains multiple files
+
+    Scenario: The user adds multiple files to existing files
+        Given a FileInput that accepts multiple files is rendered
+        And the FileInput is provided with an onChange handler
+        And the FileInput has some files attached
+        When the user selected multiple files
+        Then the onChange handler is called
+        Then the onChange handler's payload contains all files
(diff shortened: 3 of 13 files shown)


HendrikThePendric:
> Somehow it'd be nice if the `accept` and `disabled` prop could be tested too.

I've been thinking about the props I mention in my remark above, as well as the `multiple` prop. Do you agree with the following:
- These component props all map 1:1 to attributes on an actual file-input element
- It is safe to assume that native file-input elements work consistently across browsers
- So really, all we need to test is **if our props are producing the correct element attributes**
- Because testing the behaviour of the `multiple` attribute of a file-input is basically testing if the browser works correctly....

So I'd argue this component needs less e2e tests and more unit-tests...

Mohammer5:
>  this feature file and corresponding tests would be correct for the FileInput component in ui-forms, which does indeed append files to the current list.

Agreed. Just double checked. The native input element also always replaces the files (single and multiple).

I noticed that the component in `ui-forms` is adding the file list. I'm wondering if that's the right place or we should do that in `ui-core` already. Because the way that component behave comes from the design specs (if I remember correctly) and shouldn't just be available when using RFF.

> So I'd argue this component needs less e2e tests and more unit-tests...

Yeah we probably should just test if the callbacks are called properly

HendrikThePendric:
> I noticed that the component in ui-forms is adding the file list. I'm wondering if that's the right place or we should do that in ui-core already. Because the way that component behave comes from the design specs (if I remember correctly) and shouldn't just be available when using RFF.

I think you raise a very valid point here. I did consider what you suggested, but in the end, there were various reasons for implementing the `FileInput` the way I did in `ui-core`:
- The most important reason is the two completely different use-case that we had in mind. And because the needs of these components are quite different, I thought it'd be more suitable to expose a couple of very low level components in ui-core that _allow a developer to build a file input_:
    - The implementation in ui-forms as we have it now: basically a form of a controlled file-input that works slightly different than regular file input, because the file list grows and you can remove selected files individually.
    - Another implementation that we will be building is a `FileResourceInput`, which will be in `ui-widgets`. In this implementation the FileList won't even represent a list of actual `File` instances. Instead the component will show a list of file-records on the server.
- Another reason (but not as compelling) is that all of our input components are controlled, but a file input cannot be, because it's impossible to programatically set the value. By implementing this component as even lower level than a fully functioning input (i.e. you can only select files, not display the ones that were selected), I feel like this is less of an "exception on the other inputs" and more of a "completely different thing". I don't know if this part of my explanation makes sense.

Another way to "justify" the current implementation could be this: the way the `FileInput` works is comparable to the `Modal`. We did not produce one single component, but it is fully possible to create the UI as described in the design specs by simply combining a few lower-level ones.

> Yeah we probably should just test if the callbacks are called properly

Not sure I understand this one.... Did you perhaps mean _test if the attributes are attached properly_?

Issue #576, opened 2019-11-14
Modal positioning


cooper-joe:
During the demo of the new data visualizer options it was clear that the height of the modal changing between tabs was not a user friendly experience. I've been thinking about how this can be avoided and what the component itself can do to help minimize this issue.

Current solution:
- Set a max height on the modal. This works, but it requires setting an arbitrary value that may or may not make sense, depending on how good of a guess one can make about the content.

Possible solution:
- I do think we should allow for different positioning of the modal. Allowing a 'top-aligned' modal with a standard offset from the top of the viewport would mean that only the action buttons would move around when the height of the modal changes. This would be a much less jarring experience than a vertically centered modal changing height.

Overly complex, uncessary solution:
- I do not think we should try to do any kind of height detection between tabs. This is overly complex and doesn't belong in the ui-core component.

So, in conclusion, I think we should add the following functionality to the modal component:
- Position: centered, top
- if `position = top`, offset = x (x defaults to 72px)

---

I think this will allow for the flexibility we need. Simpler, single page/tab modal views can continue to use centered vertical alignment. Complex, tabbed modals, or modals that display user generated data that prevents a reasonable max-height being set, can use top alignment.

What do you think @dhis2/team-apps?

edoardo:
Perhaps using a viewport based height can be a solution for having a fixed size modal.
The downside is when you have little content and a big screen...
In any case it should be possible to override the `max-height` in the app.

ismay:
> I do think we should allow for different positioning of the modal. Allowing a 'top-aligned' modal with a standard offset from the top of the viewport would mean that only the action buttons would move around when the height of the modal changes. This would be a much less jarring experience than a vertically centered modal changing height.

@cooper-joe That seems like a nice solution. One thing I would miss from the vertically centered dialog is that it seems more natural to have it displayed in the middle of the viewport. Especially on a large screen. We're all in agreement that the vertically centered one wasn't as nice? The height changing never really bothered me that much, maybe it's just something to get used to?

Mohammer5:
Yeah, I'd prefer the top aligned over static/relative-to-body height as well, with 3 options in total:
1. `alignTop` (default)
1. `alignCenter`
1. `alignBottom`

PR #631, opened 2019-12-02
feat: extract centered content from screen cover

Mohammer5:
PR replacing #601


ismay:
Why don't we call it AppCover? Would that be appropriate. My rationale is that the ScreenCover covers the screen, and the AppCover the area that's unique to the app (as in, anything but the headerbar).

ismay:
Then the centering would not be that coupled to this component. It could be a feature though. But it seems that the base difference is _what_ it covers, and then secondary features might be how it renders its children (centered or not, etc.).

Mohammer5:
> My rationale is that the ScreenCover covers the screen, and the AppCover the area that's unique to the app (as in, anything but the headerbar).

My idea was that this component could be used in any context for centering stuff inside a component (as long as the component is not positioned statically in css). If that's what this component will be used for, `AppCover` wouldn't be appropriate.

> Then the centering would not be that coupled to this component (by name at least). It could be a feature though

If we incorporate that feature officially, I'd name it more generic right from the start.
If we don't incorporate it, it's not really a feature, because that could change anytime, so using this "feature" could cause a change in this component to be an invisible breaking change (don't know if that's the correct terminology).

Mohammer5:
> This component should also cover the use case in #576, so it needs additional props.

Ah, yes, good idea!

Mohammer5:
> And it needs a better name.

See the description in the original PR:

> I don't like the name I came up with too much though: CoverCenter
> Maybe you guys have a better name?

ismay:
> My idea was that this component could be used in any context for centering stuff inside a component (as long as the component is not positioned statically in css). If that's what this component will be used for, `AppCover` wouldn't be appropriate.

Ok, I thought that the intention of this component was to cover anything but the headerbar because of your comment in #601: `Imagine we want to cover the content of an app but not the headerbar. Currently the ScreenCover covers everything.`

So the intention of this component is just to center its children vertically and horizontally in the viewport?

Mohammer5:
> So the intention of this component is just to center its children vertically and horizontally in the viewport?

Not the viewport but the component that it's rendered inside in.
The component is positioned absolutely currently, so it'd work with any container that's positioned either relatively or absolutely.

ismay:
> > So the intention of this component is just to center its children vertically and horizontally in the viewport?
> 
> Not the viewport but the component that it's rendered inside in.
> The component is positioned absolutely currently, so it'd work with any container that's positioned either relatively or absolutely.

Ok. That seems very specific to me. Covering just the app seems like a useful component to me (since that's coupled to the headerbar height, which is difficult to know in an app). And a component for centering content in both directions seems useful (since that comes up from time to time). But a component for centering content in this manner seems very specific. Are there enough usecases for this that we need to export and maintain this as a separate component?

varl:
> Ok. That seems very specific to me. Covering just the app seems like a useful component to me (since that's coupled to the headerbar height, which is difficult to know in an app).

We only know the Headerbar height in `ui-widgets` though, and `ui-core` components should not be coupled to `ui-widgets`.

ismay:
> > Ok. That seems very specific to me. Covering just the app seems like a useful component to me (since that's coupled to the headerbar height, which is difficult to know in an app).
> 
> We only know the Headerbar height in `ui-widgets` though, and `ui-core` components should not be coupled to `ui-widgets`.

True.

varl:
I am also not convinced that we need to expose this component, which is why it was deferred from 4.0.

@Mohammer5, you mentioned that he has a use case for it in MARy. Could you explain that use case so we can look at if it's a common one?

varl:
This also doesn't actually **cover** anything since it doesn't have a Backdrop, so "AppCover" is not exactly right either. 🤔

Mohammer5:
I want to use it to center the loading spinner and error message in the new maintenance app.

But I don't think we should prevent the user from using the header bar (why should we stop the user from navigating to another app while (possibly slowly) loading the initial data; e. g. when the user accidentally clicked the maintenance app icon)

ismay:
> I want to use it to center the loading spinner and error message in the new maintenance app.
> 
> But I don't think we should prevent the user from using the header bar (why should we stop the user from navigating to another app while (possibly slowly) loading the initial data; e. g. when the user accidentally clicked the maintenance app icon)

I see. Maybe in that case it'd make sense to make the headerbar always be on top of the 'cover' components. Since it would be inconvenient to block the user from navigating to another app if it's stuck at a loading spinner. That would allow us to take care of this from within ui-core without coupling to ui-widgets. Without adding extra components.

Mohammer5:
I think we can close this for now (and re-open if we see the necessity). I'll create a component in the new maintenance app

PR #639, opened 2019-12-02
feat: transfer component

Mohammer5:
* [Design](https://github.com/dhis2/design-system/blob/master/organisms/transfer.md)
* [Specs](https://github.com/dhis2/ui-core/tree/Transfer/cypress/integration/Transfer)

Essentially this component is a multi-select with some advanced options (reordering, filtering, highlighting, etc).

Contrary to our existing `[Singe/Multi]Select` components, this one does not have a single and multi version. It's always working on an array structure, which can be limited in length (currently only to either `1` (kind of single) or `Infinite` (multiple). I've implemented it in such a way that we other values should work as well. The reason why I haven't enabled it is: There are no specs that describe this behavior and no design that let's the user know what the limitation is.

## What to look out for when reviewing

Ideally, if you find a bug, please propose a scenario that we could put into the feature files that will cover the bug you found. That's the easiest way for me to reproduce the bug AND cover it with tests.

Please read through the feature files to get an overview of what this component should be capable of and try to find missing scenarios / edge cases. The tests should pretty much cover the defined behavior, but there might be some things missing, so please test the component thoroughly. One example for missing behavior is: What should happen when clicking a highlighted option without a modifier key when multiple options are highlighted.

Also please try to see if everything has been fully covered by tests (both with cypress & jest) or if I have overlooked edge cases.

## Missing specs / tests

- [x] Individually disabled options
- [x] What should happen when clicking a highlighted option without a modifier key when multiple options are highlighted.
- [x] ~~Error options~~
  Not clear, will be excluded from this PR
- [x] Should options stay highlighted when the filter changes and they're not displayed anymore? 
  Will stay as is, options highlighted prior to a filter change remain highlighted
- [x] Should options stay highlighted after the filter has changed, they're not displayed anymore and the user selects the currently visible, highlighted options?
  Will stay as is, after selecting no source option is highlighted
- [x] Should the filter reset after selecting some opions?
  Will stay as is, filter remains

## Open issues

- [x] Talk about if/how we should refactor the in-relation-to-other components large `src/Transfer.js` file.
- [x] ~~The empty states have a design in the design files. But these components can be customized. Currently the `Transfer` component doesn't show them by default, a component has to be specified. Should I add these components and make them the default? If yes, do we want the developer using this component be able to disable the default component.~~
   No
- [x] The search functionality is a primitive search algorithm, it just checks if the string is inside the options' labels.
  - [x] ~~Do we want to have a more sophisticated functionality (fuzzy) and/or do we want to add a prop to define custom filter behavior?~~
    The filter function can now be overriden by apps to implement custom filtering (like a fuzzy search)
  - [x] ~~Should we also search the values of the fields? (it might be good to show them, but irritating to the user why an option is shown whose label doesn't match the search term.~~
    No
- [x] ~~What's the meaning of a 'error' option?~~
  Not clear, will be excluded from this PR

## Todo for myself

- [x] ~~Add stories to demo storybook~~
  Already there
- [x] Rename "options" and "selected" to "source" and "picked" (Quote by @cooper-joe: "we have a `source` list of `options` that can be `highlighted` and then `transferred` to the `picked` list…"
- [x] ~~Add explanation to this PR how the code works~~
  The `src/Transfer.js` now is more controller-code-ish. So just looking at that file and the feature files should help
- [x] Enable custom filtering from other components

## Other

It's the first time I've extensively used aliasing in the cypress test code. It has reduced the amount of test code quite significantly and makes it much easier to read. If applicable, I think it's a pattern we should adopt


cooper-joe:
Resources from Design are ready:

- [Pixel spec](https://sketch.cloud/s/DwkDk/a/jvobL7)
- [HTML&CSS mockup spec](https://codesandbox.io/s/transfer-mmqwu)
- [Design usage guidelines](https://github.com/dhis2/design-system/blob/master/organisms/transfer.md)
- [.feature files](https://github.com/dhis2/ui-core/tree/master/cypress/integration/transfer)

All of these resources should cover the documented use cases. If there are any edge cases underway then please ask me.

cooper-joe:
> What should happen when clicking a highlighted option without a modifier key when multiple options are highlighted.

Here is my suggestion for this scenaro:

```
Scenario: A user clicks a highlighted option without a modifier key when multiple options are highlighted
  Given multiple options are highlighted
  When a user clicks one of the highlighted options without a modifier key
  Then only the clicked option is highlighted
```

> What's the meaning of a 'error' option?

The 'error' option is not currently in use by any app that I know of. I included it in the design to ensure it _can_ be handled in the future, should we need it. So, I don't have an example for its usage yet, but my intention for it was: 
- show an error on a specific option, instead of/as well as on the Transfer in general - to help the user understand the error and debug.

Attempting to translate that into feature file, could it look like this?
```
Scenario: A single option is in an error state
  Given there is a transfer that contains one or more options
  And one of the options is in an error state
  Then the option in an error state is displayed using the option state: error.
```
I'm not sure this is really valid? It seems kind of self-explanatory (in the same way that we don't write .feature files for making a button go into hover state when the cursor hovers over them?)

>  Do we want to have a more sophisticated search functionality (fuzzy)

This could be very useful for some Transfer use cases: in analytics apps it is often used to search 10,000+ data items, many with similar/very slightly different names. A fuzzy search that helped analytics users find what they're looking for without needing to know exact data item names would be great. However, I would still classify this feature as 'nice-to-have' rather than 'need'.

Mohammer5:
@cooper-joe 

>  > What should happen when clicking a highlighted option without a modifier key when multiple options are highlighted.
> 
> Here is my suggestion for this scenaro:

:+1: cheers

<hr />

Reg. the error. How about we leave it out for now until we have a use-case for it in the real world?

<hr />

> This could be very useful for some Transfer use cases . . . However, I would still classify this feature as 'nice-to-have' rather than 'need'.

Ok, I'll implement that the naive filter function can be overwritten, then apps can do it when needed until we decide to use a more sophisticated default filter function

cooper-joe:
> Reg. the error. How about we leave it out for now until we have a use-case for it in the real world?

> Ok, I'll implement that the naive filter function can be overwritten, then apps can do it when needed until we decide to use a more sophisticated default filter function

Sounds good to me 👍

Mohammer5:
@cooper-joe I've added the possibility to override the filter function & made the search ignore case-sensitivity. I also updated this PR's description with a couple of questions/proposal regarding missing specs

src/Transfer.js
```diff
@@ -0,0 +1,463 @@
...
+        return Children.map(availableOptions, child => {
+            if (!enableFilter || filter === '') return child
+            if (!filtered.find(option => isOption(option, child.props)))
+                return null
+
+            return child
+        })
+    })()
+
+    return (
```
HendrikThePendric: There is a lot going on before we can get to return some JSX. I wonder whether all of this needs to be done on every render, and/or whether we could extract this into some dedicated helper functions....

Mohammer5:
@HendrikThePendric 

> It might be good to disable text-selection on these options. I keep getting a combination of highlighted text and highlighted options, and that's quite confusing.

Done

> In the custom list options demo, I can't select the custom options

Fixed

> In the reordering demo, I can't reorder the options

Fixed

src/Transfer/Selected.js
```diff
@@ -0,0 +1,35 @@
...
+export const Selected = ({
+    dataTest,
+    emptySelectionPlaceholder,
+    enableOrderChange,
+    selectedElements,
+    // eslint-disable-next-line no-unused-vars
+    onOrderChange,
+}) => (
+    <div data-test={`${dataTest}-selected`}>
+        {enableOrderChange && console.log('TODO onOrderChange')}
```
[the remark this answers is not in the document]
Mohammer5: I removed the comment. This was to remind me of the dnd reordering. But that won't be part of this PR

src/Transfer/Button.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+import propTypes from 'prop-types'
+
+import { borderColor, borderRadius } from './common.js'
+import { colors } from '../theme.js'
+
+export const Button = ({ children, label, dataTest, disabled, onClick }) => (
```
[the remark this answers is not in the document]
Mohammer5: Buttons have been replaced with our internal ones.
@cooper-joe please have a look at the reordering buttons. The `small` version of our button is larger than the one in the design. I think we should either leave it as is or implement a `tiny` version :rofl:

cooper-joe:
### Feature Request

I spoke to @Mohammer5 briefly on slack about this feature:
- allow the app to filter the `source` options (currently only the component can filter).

This request is to allow a user interface like this:
image

where `tabs` are used to show different `source` options and headers.

Pasting the relevant information from Slack:

> 13:21 Jan-Gerke Salomon:
technically it's possible if I change one small detail, but the app has to manage all of it:
Currently the filter value is controlled entirely by the component.
By giving the app the possibility to set a value instead, this should be possible:
The app can supply a custom header (which can be anything). So when the user clicks on a different tab, the app can set a different filter.
This would even work with the built-in filter functionality as the function returning the displayed options can be overwritten by the app as well
```

File: src/Transfer.js (hunk 1 above)
Comment (Mohammer5) [reply src/Transfer.js]: I've changed a lot in the `src/Transfer.js`. Please have a look

File: src/Transfer/LeftSide.js [hunk 4]
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import propTypes from '@dhis2/prop-types'
+
+import { borderColor, borderRadius } from './common.js'
+
+export const LeftSide = ({ children, dataTest, width }) => (
```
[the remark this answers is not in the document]
Mohammer5: I thinks that's a job for `ui/layouts` in v5

src/Transfer.js
HendrikThePendric: Nice 👍

cooper-joe:
Answering the open questions in the top comment:

> OPEN QUESTION: Should options stay highlighted when the filter changes and they're not displayed anymore? Current solution: yes (so you can search for a item after already having selected another item)
Tricky one. Technically, displaying options not included in the filter is wrong. However, I think that the user expectation is that any `highlighted` options would continue to be displayed, even when filtering (to allow for multi-select). So, I agree with the current solution 👍

> OPEN QUESTION: Should options stay highlighted after the filter has changed, they're not displayed anymore and the user selects the currently visible, highlighted options?

Waiting for clarification from @Mohammer5 about this one.

> OPEN QUESTION: Should the filter reset after selecting some options?
No, it should not.

Mohammer5:
@joe and I agreed on the proposed scenarios with one change in the second one:

```feature
Scenario: The user highlights an option, changes the filter to exclude that option & then changes the filter back
    Given an option is highlighted
    But not visible due to a set filter
    When the users changes the filter to include the hidden option
    Then the option is visible
    And the option is highlighted
```

```feature
Scenario: The user highlights an option, changes the filter to exclude that option, selects the highlighted options & then changes the filter back
    Given an option is highlighted
    But not visible due to a set filter
    When the user selects the visible, highlighted options
    And the users changes the filter to include the hidden option
    Then the option is visible
    And the option is not highlighted
```

src/Transfer/Button.js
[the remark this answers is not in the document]
cooper-joe: the `small` button is perfectly fine here ☺️

Mohammer5:
Custom filtering in the header is now possible, see the two following stories:

* Custom Filtering With Filter Input
* Custom Filtering Without Filter Input

cooper-joe:
I missed[ the conversation about double-click-to-select in the #ui channel on slack](https://dhis2.slack.com/archives/CBM8LNEQM/p1584634422040600). @Mohammer5 concluded:

>Ok, so for the existing PR I'll leave it out and then do it in the ui alpha version?

This component is intended to replace existing transfer-ish components dotted around DHIS2. All of the existing components employ a double-click-to-select interaction. I don't think the new `Transfer` will be able to replace existing components until double-click is working, it would be too much of a breaking change in interaction.

I saw [dblclick](https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event) noted as a possible solution, but it is not supported by mobile browsers. None of the existing transfer-ish using applications support mobile(DataViz, Maintenance), so I think a temporary mouse-only support would be acceptable to get the component in use. I'm trying to get some feedback to see if dblclick works on a touch-based laptop screen.

What do you think?

(Apologies if I have misunderstood and this component will not be released until the ui alpha? But as I understand it this component is ready for review and ready to use now?)

Mohammer5:
@cooper-joe I've added a simple double click mechanism

src/TransferOption.js
```diff
@@ -1,46 +1,69 @@
-import React from 'react'
+import React, { useRef } from 'react'
 import cx from 'classnames'
 import propTypes from '@dhis2/prop-types'
 
 import { colors } from './theme.js'
 
+const DOUBLE_CLICK_MAX_DELAY = 250
```
cooper-joe: Could we adjust this to `500`? I've read around and I see that the [default value in Windows is 500](https://docs.microsoft.com/en-gb/windows/win32/api/winuser/nf-winuser-setdoubleclicktime?redirectedfrom=MSDN), so I think that's a good place to start. I've tested `500` and it seems reasonable.

In an ideal world we'd use the users OS setting as double-click-speed is a common setting to adjust, but I'm not sure the browser has access to that.

cooper-joe:
Been doing some testing this morning, this component brings me joy! 

Comparing it to existing functionality, I noticed SHIFT+click works different in this component.

- In existing transfer-ish implementations, SHIFT+click selects a range of items (see this in action in Data Visualizer > Data modal)
- In this `Transfer`, SHIFT+click acts like CMD/CTRL+click and selects multiple invididual items.

I think the range selection is a nice-to-have. A scenario outline for this functionality could be: 
```
    Scenario Outline: A user SHIFT+clicks options in a list
      Given the <type> list has items "1, 2, 3, 4, 5"
      When the user clicks on <firstItem> without a modifier key and then clicks <secondItem> with SHIFT
      Then <highlightedOptions> are highlighted

      Examples:
        | type   | firstItem | secondItem | <highlightedOptions> |
        | source | 2         | 5          | 2, 3, 4, 5           |
        | source | 2         | 3          | 2, 3                 |
        | picked | 1         | 3          | 1, 2, 3              |
        | picked | 5         | 1          | 1, 2, 3, 4, 5        |
```

However, this only accomodates a limited, simple example. In theory a user could SHIFT+click multiple times to continue selecting the 'range'. @Mohammer5, two questions for you:
- can you figure out a more flexibile Scenario that specifies the ability to make multiple SHIFT+clicks?
- can this be included in this iteration of the `Transfer`, or it is a big addition that would be better tackled after the first release?

Mohammer5:
@cooper-joe The shift click behavior should work now.
@ismay @HendrikThePendric I've added all missing tests & fixed the bugs we've found during my presentation last week. Should be good for review now (will squash the commits later)

ismay:
Just FYI, I've been going through this PR today and will continue tomorrow. One thing I'm interested in, is where there's an overlap between the UX for this component and the Select. If there is, it might be nice to make sure that things work in a similar manner. So that when people have learned something on the select, they can also use it on the transfer component, and vice versa.

In a similar vein, I'm also wondering about overlaps in the implementation of either in an app. And how much we can/should align the api there. Both have search for instance, but currently have slightly different mechanisms. So that devs can also apply learned patterns from one component to the other.

And lastly, there might also be overlap in the code, so things that we could reuse in the lib, internally, or just align to work in a similar manner.

So I'll hopefully be able to answer that a bit better after reviewing the PR (besides looking at the usual PR review topics of course).

cypress/drafts/Transfer/drag-and-drop-items.feature
```diff
@@ -0,0 +1,39 @@
...
+        And the items should be ordered based on their previous ordering
+
+    Scenario: The user clicks, drags and drops a single item in a list of one
+        Given the selected list has one item
+        And the selected list has one highlighted item
+        When the user clicks, drags and drops the highlighted item
+        Then the highlighted item should remain in position and shouldn't not move
+
+
+    # an attempt an outline of how multiple items reorder on drag and drop
```
ismay, line 28: I'd change the wording slightly, `an attempt to outline how multiple items reorder on drag and drop`

cypress/drafts/Transfer/highlight-keyboard.feature
```diff
@@ -0,0 +1,33 @@
...
+    ##
+    #
+    # Keyboard interaction
+    #
+    # NOTE: For now we won't support keyboard functionality until we have a
+    # proper keyboard system in place
+    #
+    ##
+
+    Scenario Outline: The user presses enter on a focussed item that is not already highlighted
```
ismay, line 12: `focussed` -> `focused`

cypress/integration/Transfer/disabled-transfer-buttons.feature
```diff
@@ -0,0 +1,42 @@
+@component-transfer @button-states
+Feature: Disable transfer buttons when actions are not possible
+
+    Scenario: None of the available options is highlighted
```
ismay, line 4: None of the available options is highlighted -> None of the available options are highlighted

cypress/integration/Transfer/display-order.feature
```diff
@@ -0,0 +1,30 @@
...
+        Then the order of the available options should match the order of the supplied options
+
+    Scenario: Some of the supplied options have been selected
+        Given some of the supplied options have been selected
+        Then the order of the remaining available options should match the order of the supplied options
+
+    Scenario: A single selected options is deselected
+        Given some of the supplied options have been selected
+        When the user deselects one selected option
+        Then it should take its position according to the order of the supplied options
```
ismay, line 15: `it should take its position according to the order of the supplied options`, maybe you could word this as `it should be positioned according to the order of the supplied options`. Not sure about this one, but the latter reads a bit easier for me (same as with the other comments, wording occurs in other areas as well)

cypress/integration/Transfer/filter-options-list/index.js
```diff
@@ -0,0 +1,133 @@
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+import { extractOptionFromElement } from '../common'
+
+Given('filtering is enabled', () => {
```
ismay, line 5: Why are there empty Given clauses here?

cypress/integration/Transfer/highlight-range-of-options/index.js
```diff
@@ -0,0 +1,325 @@
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+import { extractOptionFromElement } from '../common'
+
+Given('the option list has one or more items', () => {
```
ismay, line 5: For the givens in this file that refer to a specific number of items/options, maybe we should assert that that's actually true. To ensure that the statement matches what's actually happening.

cypress/integration/Transfer/transferring-items.feature
```diff
@@ -0,0 +1,52 @@
...
+    #     Given some options are available
+    #     When the user double clicks an item in the options list
+    #     Then the item should be removed from its options list
+    #     And the item should be visible at the bottom of the selected list
+
+    # Scenario: The user double clicks an item in the selected list
+    #     Given some options are selected
+    #     When the user double clicks an item in the selected list
+    #     Then the item should be removed from the selected list
+    #     And the item should be visible in its original position in the options list
```
ismay, line 52: I'd prefer moving this to drafts

cypress/support/get.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Takes a space-separated list of dataTestIds and wraps & prefixes them
+ * @param {string} selectors
+ * @param {string} [prefix]
+ * @returns {Object}
+ *
+ * @example
+ * //These are equivalent:
+ * cy.get('{node-input} input')
+ * cy.get('[data-test="dhis2-uicore-node"] input')
```
ismay, line 10: Should this not be:

```
 * cy.get('{node} input')
 * cy.get('[data-test="dhis2-uicore-node"] input')
```

src/Transfer.js
```diff
@@ -0,0 +1,390 @@
...
+    labelRemoveIndivitual,
+    leftFooter,
+    leftHeader,
+    maxSelections,
+    optionsWidth,
+    rightFooter,
+    selected,
+    selectedWidth,
+    showFilterInput,
+    onFilterChange,
```
ismay, line 79: Some differences with the Select api here, that I think would be good to align:

* `emptyOptionsPlaceholder` and `emptySelectionPlaceholder`. The select accepts a component under the `empty` prop for this. So it doesn't use the word placeholder.
* `filter`, `filterLabel`, `filterOptions`, `initialFilter` and `showFilterInput`. The Select does not have a modifiable filtering algorithm. It just has a `filterable` prop for toggling the filter and `filterPlaceholder` and `noMatchText` (so different wording as well)
* `labelAddAll`, `labelAddIndividual`, `labelRemoveAll`, `labelRemoveIndivitual` the Select also accepts text for placeholders and buttons, but uses different terminology: `placeholder`, `filterPlaceholder`, `clearText`, `loadingText`, `noMatchText`
* Single component with toggleable single or multiple selections, versus the two exported components for the select.

Now I'm not saying I prefer one over the other, but merely pointing out differences in the API, where the functionality seems similar (though the components are different of course). This could be confusing to devs using both components, so I think we should make sure that we align areas that do similar things, and only diverge where we think it's justified (because for example, the component really needs different functionality). I think it would be good to discuss that.

p.s.: small typo: `labelRemoveIndivitual`

src/Transfer.js
```diff
@@ -0,0 +1,390 @@
...
+}) => {
+    /**
+     * As this component will filter and sort children
+     * in the selected area, adding keys will prevent
+     * reconciliation from suffering badly
+     */
+    const allOptions = addKeyToReactOptions(children)
```
ismay, line 86: I see that in the Transfer story the options already have keys (see [here](https://github.com/dhis2/ui-core/blob/71432d60877f0de4ebca20eacd2e49d2e1f691dc/src/__demo__/Transfer.stories.js#L31)). Given that React will also [warn for list items without keys](https://reactjs.org/docs/lists-and-keys.html#basic-list-component) and assign keys itself if they're missing (as a last resort), I think it would be more idiomatic to defer this responsibility to the dev implementing this component in an app. Unless I'm misreading sth here.

src/Transfer.js
```diff
@@ -0,0 +1,390 @@
...
+    const isFilterEnabled = showFilterInput || onFilterChange
+    const [internalFilter, setInternalFilter] = useState(initialFilter)
+    const actualFilter = onFilterChange ? filter : internalFilter
+
+    /**
+     * These are all the available option react elements.
+     * It will replace all selected options with null
+     */
+    const sourceOptions = filterOutOptions(allOptions, selected)
+    const filteredSourceOptions = getSubsetByFilter({
```
ismay, line 104: To me the wording here is slightly unclear. After reading what's going on, to me this would make sense:

```
const availableOptions = filterSelectedOptions(allOptions, selected)
const filteredAvailableOptions = getSubsetByFilter({ })
```

Technically, I'd also expect filterSelectedOptions to filter out the null values, but I'm not too bothered by that.

src/Transfer.js
```diff
@@ -0,0 +1,390 @@
...
+    } = createDoubleClickHandlers({
+        selectedPlainOptions: selected,
+        setHighlightedSourceOptions,
+        setHighlightedPickedOptions,
+        onChange,
+        maxSelections,
+    })
+
+    return (
+        <Container dataTest={dataTest} className={className} height={height}>
```
ismay, line 162: Minor detail, I see we're passing the base dataTest prop here, and then appending the suffix in the components themselves. The pattern we use in other components is the opposite way (append suffix in parent instead of child). Might be inconsequential, but it is a difference.

src/Transfer.js
```diff
@@ -0,0 +1,390 @@
...
+     * These are all the highlighted option react elements on the options side.
+     */
+    const {
+        highlightedOptions: highlightedSourceOptions,
+        setHighlightedOptions: setHighlightedSourceOptions,
+        toggleHighlightedOption: toggleHighlightedSourceOption,
+    } = useHighlightedOptions({
+        availableReactOptions: filteredSourceOptions,
+        disabled,
+        maxSelections,
```
ismay, line 134: There are a couple of option types, naming wise, in this component. We have:

* source options
* picked options
* selected options
* available options
* the filtered variants
* and the react vs plain object variants

I think that understanding those different option types is a big part to understanding this component. Naming wise, source, picked, filtered and the react vs plain all make sense to me.

* I'm not sure if there is a difference between selected and picked though?
* I see you're using available as well, I assume that that is for when a function takes options, but they're not always of the same type? I that case we could maybe just omit available?

1. I think that, after clearing up the above, a comment block at the top that clearly defines what the different option types signify would help devs a lot in understanding the component.
2. For the select we also deal with plain and react options. But there, we only call react options `reactOptions`, and the plain options just `options`. I think your distinction is clearer though, so I'm in favour of applying the same naming strategy for the select.

src/Transfer.js
```diff
@@ -0,0 +1,390 @@
...
+
+Transfer.defaultProps = {
+    dataTest: 'dhis2-uicore-transfer',
+    initialFilter: '',
+    selected: [],
+    height: '240px',
+    optionsWidth: '320px',
+    selectedWidth: '320px',
+    maxSelections: 1,
+    filterOptions: defaultFilterOptions,
```
ismay, line 325: This name confused me a bit. I thought it was for additional options you could set for the filter, but I see it's the filtering function. In that case I would have expected maybe optionFilter and defaultOptionFilter. But maybe that's just my expectations.

src/Transfer/PickedOptions.js
```diff
@@ -0,0 +1,51 @@
...
+export const PickedOptions = ({
+    children,
+    dataTest,
+    toggleHighlightedPickedOption,
+    emptySelectionPlaceholder,
+    highlightedPickedOptions,
+    deselectSingleOption,
+}) => (
+    <div data-test={`${dataTest}-pickedoptions`}>
+        {!React.Children.count(children) && emptySelectionPlaceholder}
```
ismay, line 17: For something like this (rendering a fallback), I'm more used to seeing an if at the top that returns the empty placeholder early, and then with the rendering of the children after that.

I don't see a technical downside to doing it this way (except for maybe a call to map without children if there are none, which doesn't bother me). And the benefit is that you only have to define the wrapper once.

I do think I'd personally prefer the if block, and repeating the wrapper. I'm more used to seeing that, so it's easier to parse for me. But that might just be me.

src/Transfer/common.js
```diff
@@ -0,0 +1,113 @@
...
+ * Click modes when clicking on an option with/without
+ * a modifier key (ctrl, alt, cmd, shift)
+ */
+
+// no or multiple modifier keys
+export const REPLACE_MODE = Symbol('REPLACE_MODE')
+// add/remove options from selection
+export const ADD_MODE = Symbol('ADD_MODE')
+// create selection range
+export const RANGE_MODE = Symbol('RANGE_MODE')
```
ismay, line 17: We could just use regular string here right? I could see the use of a symbol if we have sth like a object where we don't want the key to be accidentally overwritten or something, but in this case I'd just go with a string const. But let me know if I'm missing something.

src/TransferOption.js
```diff
@@ -0,0 +1,82 @@
...
+                if (doubleClickTimeout.current) {
+                    clearTimeout(doubleClickTimeout.current)
+                    doubleClickTimeout.current = null
+
+                    onDoubleClick({ option }, event)
+                } else {
+                    doubleClickTimeout.current = setTimeout(() => {
+                        clearTimeout(doubleClickTimeout.current)
+                        doubleClickTimeout.current = null
+                    }, DOUBLE_CLICK_MAX_DELAY)
```
ismay, line 36: Should we have to take unmounting after a click but before the timeout into account? Or will that not be an issue?

ismay: I guess the worst they could do is remove the current property, in which case you're passing undefined to clearTimeout and setting current to null.

src/__e2e__/Transfer/add_remove-highlighted-options.stories.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable react/prop-types */
+import React from 'react'
+
+import { Transfer } from '../../index'
+import { options, statefulDecoator /*, useTransferState*/ } from './common'
+
+export default {
+    title: 'Transfer add/remove highlighted options',
+    decorators: [statefulDecoator()],
```
ismay, line 9: Tiny typo: `statefulDecoator`

src/Transfer.js
ismay: Made a proposal here: https://dhis2.slack.com/archives/CBM8LNEQM/p1586440987039200

src/Transfer/PickedOptions.js
Mohammer5: Yeah I prefer to keep a single container instead of an if-block, but I guess it's personal preference. I prefer to see at one glance what's going to be rendered. It visually combines the why, what and where.

Also when using the wrapper several times, either there's a slight chance for misalignment due to a type or it slightly increases complexity because of keeping stuff DRY so the misalignment can't happen.

src/Transfer.js
Mohammer5: The reason I chose this name is because the general function is called `filter` and this one is a filter function for options. But I get that the word "Options" is ambiguous here, which def. will cause confusion. Maybe `applyFilterToOptions`?

src/Transfer.js
Mohammer5: "available" is not gone.

src/Transfer.js
Mohammer5: Done

src/Transfer.js
Mohammer5: The reason I did this is that it will work with static children as well.
But a `Transfer` with a lot of static options is probably a never occurring edge-case, we could get rid of this. Do you prefer this to be removed?

cypress/integration/Transfer/transferring-items.feature
Mohammer5: Feature is actually available. I uncommented the scnearios and added the test code

cypress/integration/Transfer/filter-options-list/index.js
Mohammer5: This is due to the nature of cypress and cucumber.
These are background steps so they're true for all scenarios (which makes sense from a feature-file structure perspective).

But we can't assert it's true because the step happens before the step that's loading the correct story. The story already has been prepared to match this scenario, so I kept it as a no-op

cypress/integration/Transfer/display-order.feature
Mohammer5: I don't really care about this one. @cooper-joe ?

src/Transfer.js
Mohammer5: > The select accepts a component under the empty prop for this.

`empty` sounds like a modifier, like `dense`. I think we should somehow incorporate the word "default", because it's defaulting to a specified component under a certain condition. The word "placeholder" should probably go as it has a very specific meaning for input elements. But I'm not sure how to name the props then.. `sourceDefaultComponent` & `pickedDefaultComponent`?

> `filter`, `filterLabel`, `filterOptions`, `initialFilter` and `showFilterInput`. The Select does not have a modifiable filtering algorithm. It just has a filterable prop for toggling the filter and filterPlaceholder and noMatchText (so different wording as well)

I first used `enableFilter` which is the equivalent of `filterable`, but I ran into issues once the behavior could be changed.

Let's say you want to add custom components in the header which filter the options (e. g. with a select; by year) but you don't want to show the filter input field. Providing the `filterOptions` callback should suffice to trigger the "filterable" behavior that's inside the component. If it does want to show the filter input, a `filterable` prop wouldn't make much sense, so I chose `showFilterInput`

> `labelAddAll`, `labelAddIndividual`, `labelRemoveAll`, `labelRemoveIndivitual` the Select also accepts text for placeholders and buttons, but uses different terminology: placeholder, filterPlaceholder, clearText, loadingText, noMatchText

I think the word "placeholder" is appropriate in your case. "My" props are "text"s as well. So the question is: `label` vs `text` (tbh I don't care) and prefix vs suffix. I def. prefer the term we choose to prefix the prop as that will group them in the prop types (and ideally in the component's argument list; that's not one of our conventions though).

> Single component with toggleable single or multiple selections, versus the two exported components for the select.

Two separate components is what I tried initially, it increased the complexity so I chose against it. This component does not have a `single` vs `multi` approach, instead the amount of selections can be limited. It currently only allows two values (`1` and `Infinity`), but we can increase the range of allowed selections without introducing a breaking change, we'd just have to extend the UI. It also removed the need to handle two different prop types for `selected`. Bot the range instead of single/multi and the single prop type decrease complexity.

Mohammer5: `labelRemoveIndivitual` <- fixed

cypress/integration/Transfer/display-order.feature
cooper-joe: I agree with @ismay `it should be positioned according to the order of the supplied options` is better.

cooper-joe:
One thing that catches me out when doing testing with the `Transfer` is remembering to set `maxSelections` to `infinity`. I think that's because all the `Transfer` use-cases I know of have an unlimited max selections, so I just expect it to work that way. I think having `maxSelections` prop default to `1` is not helpful and it would be better to default to `inifinity`.

src/Transfer.js
Mohammer5: Discussion over: We keep the current names & I've added a legend in the doc block of the component

Mohammer5:
> One thing that catches me out when doing testing with the Transfer is remembering to set maxSelections to infinity. I think that's because all the Transfer use-cases I know of have an unlimited max selections, so I just expect it to work that way. I think having maxSelections prop default to 1 is not helpful and it would be better to default to inifinity.

@cooper-joe done

src/Transfer/PickedOptions.js
ismay: I don't have a strong preference for either, so going with your preferred approach seems fine to me 👍

src/Transfer.js
ismay: Yeah, I'd personally remove it. It seems more idiomatic.

cypress/integration/Transfer/filter-options-list/index.js
ismay: Ok, I see. In that case it makes more sense to me to add the wording to the step that's loading the story. If the story is providing multiple steps, then I'd reflect that in the title, instead of having empty clauses that don't actually do anything.

src/Transfer.js
ismay: > But I'm not sure how to name the props then.. sourceDefaultComponent & pickedDefaultComponent?

Yeah. I think in that case I'd go with `sourceEmptyComponent` and `pickedEmptyComponent`. To me that seems the clearest from a user perspective. Default still seems like it could refer to sth else.

> empty sounds like a modifier, like dense.

Yeah true. Using emptyComponent for the select seems good, if we do the above.

> Let's say you want to add custom components in the header which filter the options (e. g. with a select; by year) but you don't want to show the filter input field. Providing the filterOptions callback should suffice to trigger the "filterable" behavior that's inside the component. If it does want to show the filter input, a filterable prop wouldn't make much sense, so I chose showFilterInput

I see. I guess we could chalk this up to the different functionality then 👍

> I think the word "placeholder" is appropriate in your case. "My" props are "text"s as well. So the question is: label vs text (tbh I don't care) and prefix vs suffix. I def. prefer the term we choose to prefix the prop as that will group them in the prop types (and ideally in the component's argument list; that's not one of our conventions though).

In this case, I'd go with what the Select has done for simplicity. Unless you really hate it, but then we should also change it for the select.

> Two separate components is what I tried initially, it increased the complexity so I chose against it. This component does not have a single vs multi approach, instead the amount of selections can be limited. It currently only allows two values (1 and Infinity), but we can increase the range of allowed selections without introducing a breaking change, we'd just have to extend the UI. It also removed the need to handle two different prop types for selected. Bot the range instead of single/multi and the single prop type decrease complexity.

Makes sense, so yeah, then we can chalk this one up to a functionality difference as well 👍

src/Transfer.js
ismay: Or, I'm not sure if this applies, but filterFn and filterOptionsFn?

src/Transfer.js
varl: Also vote to remove it.

src/Transfer.js
varl: > In this case, I'd go with what the Select has done for simplicity. Unless you really hate it, but then we should also change it for the select.

I vote for going with the established way for now, and if it needs to be changed, raise an issue and propose how we do it across all components.

Mohammer5: I chose `selectedEmptyComponent` instead of `pickedEmptyComponent` as that matches with the `selected` prop

src/Transfer.js
Mohammer5: Now it's `searchTerm` and `filterOptions`.
If we want to make clear that `filterOptions` needs to be a function, I'd prefer `filterCallback`, what do you think?

ismay: Yeah filterCallback sounds good to me as well 👍

Mohammer5: Done

Mohammer5:
@ismay @cooper-joe @varl This component is ready for another review, I think I've addressed all mentioned issues

There are two open questions in this PR's description:
1. Should the default search algorithm also search the values of the options or just the label
1. There are no specs and no design for disabled options (skip for now, or is that important? It'd be relatively straight forward to add that though)

Mohammer5:
@dhis2/team-apps @dhis2/front-end ping.
This component is almost done and ideally is released soon so it can be put into the new `ui` library as well. I'm just waiting for approval and answers to the two questions above

cooper-joe:
> There are no specs and no design for disabled options (skip for now, or is that important? It'd be relatively straight forward to add that though)

My suggestions for the disabled items @Mohammer5 :

### Specs 

```
Scenario: The user clicks a disabled option
  Given a source list that contains a disabled option
  When the user clicks a disabled option
  Then the disabled option is not highlighted

Scenario: The user double clicks a disabled option
  Given a source list that contains a disabled option
  When the user double clicks a disabled option
  Then the disabled option is not transferred to the picked list

Scenario: The user clicks the 'move all to picked list' button with disabled options in the source list
  Given a source list that contains at least one disabled option and at least one enabled option
  When the user clicks the 'move all to picked list' button
  Then the disabled options are not transferred to the picked list
  And the enabled options are transferred to the picked list

Scenario: The user CMD/CTRL+clicks a disabled item with other items highlighted
  Given a source list that contains at least one disabled option and at least one enabled highlighted option
  When the user CMD/CTRL+clicks the disabled item
  Then the disabled item is not highlighted
  And the previously highlighted item(s) are still highlighted

Scenario: The user SHIFT+clicks a disabled item with other items highlighted
  Given a source list that contains at least one disabled option and at least one enabled highlighted option
  When the user SHIFT+clicks the disabled item
  Then the disabled item is not highlighted
  And only the previously highlighted item(s) are highlighted
```

Not included:
- meta+click on a disabled option without other options highlighted. I figured this was covered and/or implied, but can be made explicit if needed.
- filtering disabled options. Disabled options are shown when filtering - so I didn't think there was any functionality to describe here.
- disabled option in the picked list. This shouldn't be happening because there would be no way to remove them. This scenario of 'what happens if a picked option becomes disabled at a latest date' belongs to app logic, not this component.

### Design

`grey600` and `cursor: not-allowed` for the default options.

ismay:
> Should the default search algorithm also search the values of the options or just the label

I would say just the labels. That's the only thing that's visible to a user. Otherwise it would be confusing.

Mohammer5:
All issues / open questions have been tackled/handles. Please approve/request changes.

src/Transfer.js
```diff
@@ -0,0 +1,403 @@
...
+    maxSelections,
+    optionsWidth,
+    rightFooter,
+    searchTerm,
+    selected,
+    selectedWidth,
+    showFilterInput,
+    onFilterChange,
+}) => {
+    /**
```
varl, line 99: Are these comments intended to show up in the JSDocs?

Leading a comment with two stars (`/**`) means JSDocs is going to pick it up. If these comments are meant for the dev reading the source code only, then we should use a single star comment (`/*`).

cooper-joe:
Gave it the final check, I have only one change request:
- when there is only a disabled option in the `source` list, the `move all to picked list` button should be disabled. Currently it is clickable but nothing happens on click.

Screenshot:
image


Here is a Scenario to test for it:
```
Scenario: Only disabled options in source list
  Given a source list that has only disabled options
  Then the 'move all to picked list' button should be disabled
```

src/Transfer.js
Mohammer5: Thanks. I've removed one asterix from all internal comments

src/Transfer.js
```diff
@@ -169,14 +169,15 @@ export const Transfer = ({
         maxSelections,
     })
 
+    console.log('filterable && !hideFilterInput', filterable, !hideFilterInput)
```
paschalidi, line 172: do we want this console?

Mohammer5: Oops, thanks!

Mohammer5:
This component now offers the following props for using/modifying the filtering behavior:

* `filterable` (needs to be used to enable filtering)
* `filterCallback` (customizes the filtering mechanism)
* `filter` (controlled filter value when using `filterCallback`)
* `hideFilterInput` (can be used to hide the filter input when using custom filtering in the header)

Mohammer5:
@cooper-joe "add all" button is now disabled when there's no enabled option in the source list

Mohammer5:
Aaaaaand it's in

PR #640, opened 2019-12-02
test: add data-test attribute proposal

ismay:
A proposal for adding data-test attributes to our ui-core components. When talking about it I thought that static ones made sense, but since a lot of components are not regular html but react components, they still require that we implement a prop for it in certain areas.

So actually what @varl was suggesting is necessary for the `data-test` attr to be set on react components. To enable that we can add:

* A dataTest prop
* Have a default for it (so the Label has `dhis2-uicore-label` for example) that can be overridden

Though that doesn't mean that we can reach into a component's internals from the outside and add dataTest attrs there (otherwise we'd need to add and pass through a ton of props). What I've done so that child components can be targeted is use the dataTest string as a base. So `${dataTest}-label`, `${dataTest}-help`, etc.

To demonstrate I've quickly sketched how something like this would work for the SingleSelectField for example. Which would result in the following html:

Screenshot 2019-12-02 at 16 06 05

(Note that this is just a sketch, so things are missing and it's not exhaustive, but just as a demonstration).

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -10,8 +10,8 @@ import { spacers } from './theme.js'
  * @example import { Field } from @dhis2/ui-core
  * @see Live demo: {@link /demo/?path=/story/field--default|Storybook}
  */
-const Field = ({ children, className }) => (
-    <div className={className}>
+const Field = ({ children, className, dataTest }) => (
+    <div className={className} data-test={dataTest}>
         {children}
         <style jsx>{`
             div {
--- a/src/Help.js
+++ b/src/Help.js
@@ -12,13 +12,14 @@ import { spacers, theme } from './theme.js'
  * @example import { Help } from @dhis2/ui-core
  * @see Live demo: {@link /demo/?path=/story/help--default|Storybook}
  */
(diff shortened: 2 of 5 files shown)


varl:
OK, I'm still onboard with that approach.

ismay:
@varl I've updated the approach a little bit. What we were saying about chaining selectors doesn't work if there are multiple components of the same type. So the two `Help` components in the `SingleSelectField`. So I'm concatenating the base `dataTest` string with `-validation` and `-help` for the two different components.

That way we don't need to pass through a prop for every dataTest attr that should be set.

ismay:
How does everyone feel about this approach? So if we go with this our components would:

1. Accept a `dataTest` prop.
2. Have a default for `dataTest`
3. If useful, concat `dataTest` with further strings for more precise targeting

Let me know if there's anything I might've overlooked!

varl:
I'm open to give this a go.

ismay:
All right, I'm working on a more complete implementation in #641 (it's also branched from master instead of the Select integration PR). I'll close this in favor of that one.

PR #641, opened 2019-12-02
test: add data-test attributes

ismay:
Adding data-test props to our components, which will set data-test attributes. I'm putting the data-test attr. on the root html element that gets rendered by our components. Feel free to comment and review, there's probably plenty that we need to decide on. See: 

* https://github.com/dhis2/ui-core/pull/640 (discussion before this PR)
* https://jira.dhis2.org/browse/TECH-261 (tracking issue)
* was blocked by: https://github.com/dhis2/ui-core/pull/652

Changes required:

~~- The TabBar is a bit of an exceptional case. If we attach the data-test attribute to the div with className `tab-bar` it'll consistently be attached to the same component. But since it wraps TabBar in the ScrollBar component if it's scrollable the data-attribute won't always be attached to the root component. And it won't be possible to target the ScrollBar through a data-test attribute. I propose we wrap the entire component, including the ScrollBar, in a div where we set the data-test attribute, to keep it consistent.~~
~~- For other components, which are returning an array of components directly, we might also have to add a wrapper to be able to add a sensible data-test attribute (CheckboxGroup, RadioGroup, SwitchGroup, ToggleGroup). I think I'd prefer doing that, as otherwise there's no way to add a correct data-test prop.~~

Notes:

- For components that use the `Layer` component, I've set the attribute on the first available child, since that component doesn't output html directly.
- I've removed spaces where I encountered them between the prop-types. They're alphabetically sorted, so the original grouping no longer applies.
- Some components we're spreading prop-types from others into their own. This prevents linting of unused props in eslint, as eslint doesn't evaluate those. It's also easier to accidentally add wrong proptypes when using spread (as happened with splitbutton). I can remove or extract this, as it's not directly related or crucial to the PR. Let me know what everyone prefers.
- I've not set a default for `dataTest` for child components that are not directly publicly exported. It won't ever be used as it'll always be set from the parent. So instead I made the `dataTest` prop required to indicate clearly that the parent should pass it.
- In adding the child data-test attributes I'm trying to add them based on functionality, not html structure. That way test will keep working if the html changes and functionality stays the same, but tests will fail if the functionality changes/was removed and a test still tries to use it. That seems the most useful to me.
- For now I've added child data-test attributes for children that: are action triggers (X's on chips, etc.), that have duplicate siblings (multiple help components, multiple buttons, etc.) or are rendered in portals. The other ones I thought we could add on-demand.

Tasks:

- [x] Finalize adding dataTest props for the components that need it
- [x] Set defaults for all dataTest props
- [x] We have a couple of components where we're spreading all props on a child. I can separate out the dataTest prop*
- [x] Extract splitbutton initialFocus fix
- [x] Add sub-dataTest attrs to children where necessary
- [x] Update jsdoc

Follow-up:

- [ ] Fix issue https://github.com/dhis2/ui-core/issues/669 (added by @HendrikThePendric)

---

_* However, would it not also be good to just be explicit about the props we're passing to the child? It's a one-time effort, but seems good to be more explicit (could take care of that in a separate PR)._

### DIFF
--- a/src/Chip.js
+++ b/src/Chip.js
@@ -26,6 +26,7 @@ const Chip = ({
     onRemove,
     onClick,
     icon,
+    dataTest,
 }) => (
     <span
         onClick={e => {
@@ -38,10 +39,11 @@ const Chip = ({
             disabled,
             dragging,
         })}
+        data-test={dataTest}
     >
         <Icon icon={icon} />
         <Content overflow={overflow}>{children}</Content>
-        <Remove onRemove={onRemove} />
+        <Remove onRemove={onRemove} dataTest={`${dataTest}-remove`} />
 
         <style jsx>{`
             span {
@@ -95,6 +97,10 @@ const Chip = ({
     </span>
 )
 
+Chip.defaultProps = {
+    dataTest: 'dhis2-uicore-chip',
+}
+
 /**
  * @typedef {Object} PropTypes
  * @static
@@ -107,15 +113,15 @@ const Chip = ({
  * @prop {boolean} [disabled]
  * @prop {boolean} [dragging]
  * @prop {boolean} [overflow]
+ * @prop {string} [dataTest]
  */
 Chip.propTypes = {
     children: propTypes.string.isRequired,
     className: propTypes.string,
+    dataTest: propTypes.string,
     disabled: propTypes.bool,
-
     dragging: propTypes.bool,
     icon: propTypes.element,
-
     overflow: propTypes.bool,
     selected: propTypes.bool,
     onClick: propTypes.func,
--- a/src/AlertBar.js
+++ b/src/AlertBar.js
@@ -102,6 +102,7 @@ class AlertBar extends Component {
             critical,
             icon,
             actions,
+            dataTest,
         } = this.props
         const { visible, hidden } = this.state
 
@@ -120,6 +121,7 @@ class AlertBar extends Component {
                     critical,
                     visible,
                 })}
+                data-test={dataTest}
                 onMouseEnter={this.stopDisplayTimeOut}
                 onMouseLeave={this.startDisplayTimeout}
             >
@@ -132,7 +134,7 @@ class AlertBar extends Component {
                 />
                 <Message>{children}</Message>
                 <Actions actions={actions} hide={this.hide} />
-                <Dismiss onClick={this.hide} />
+                <Dismiss onClick={this.hide} dataTest={`${dataTest}-dismiss`} />
 
                 <style jsx>{styles}</style>
             </div>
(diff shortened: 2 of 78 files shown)


varl:
> * However, would it not also be good to just be explicit about the props we're passing to the child? It's a one-time effort, but seems good to be more explicit (could take care of that in a separate PR).

Please be explicit and do not spread props to the child. As you say it is a one-time effort and a lot easier to reason about. In `ui-core` spreading props to children is an anti-pattern.

ismay:
> > We have a couple of components where we're spreading all props on a child. I can separate out the dataTest prop*
> 
> Where are we doing this? Must have slipped past me. :|

These are the ones I'm working on now: https://github.com/dhis2/ui-core/pull/652. I've just checked the components at src/ for starters. For the AlertBar it's debatable by the way, but it seems unnecessary, so I thought I might as well tackle it.

src/Chip.js
```diff
@@ -38,10 +39,11 @@ const Chip = ({
             disabled,
             dragging,
         })}
+        data-test={dataTest}
     >
         <Icon icon={icon} />
         <Content overflow={overflow}>{children}</Content>
-        <Remove onRemove={onRemove} />
+        <Remove onRemove={onRemove} dataTest={`${dataTest}-remove`} />
```
Mohammer5: Not sure if this is necessary. If we gave the components in the sub folders a default value as well, the `Remove` would have `dhis2-uicore-chip-remove` and could be targeted like this: `[data-test="custom-chip-id"] [data-test="dhis2-uicore-chip-remove"]`.

This would eliminate the need to create these dynamic attributes and simplify our components' code a bit by not having to add the dataTest unless we explicitly need it

ismay:
> What I miss in this list is: "Unspread" props, these changes are part of the "add data-test" commits, which I think is wrong, because you won't see these changes from the commit history. Not a biggie here, because it's "only" prop types, but I think we should pay attention to that

> would have been nice to have some of the separation that @Mohammer5 mentions. But in this case it's not a deal breaker for me.

Definitely. I also felt the same, see also the top notes. I could extract them to a separate PR?

varl:
> Definitely. I also felt the same, see also the top notes. I could extract them to a separate PR?

If it isn't too much work, then splitting them out will make it easier to understand the history of when/why we did so.

ismay:
> > Definitely. I also felt the same, see also the top notes. I could extract them to a separate PR?
> 
> If it isn't too much work, then splitting them out will make it easier to understand the history of when/why we did so.

I'll do that, I think it should be easy enough 👍

Mohammer5:
> I'm OK with landing these commits without squashing

I'm still in favor of squashing, but I don't mind @ismay deciding what he prefers.
Squashing the commits would result in a better overview in the history as it'd be shorter and we most likely won't need a commit per component to understand the history in future.

src/Chip.js
ismay: We talked a bit about this on Slack, in the end we decided to stick with the existing approach.

PR #648, opened 2019-12-04
test(transfer): add transfer feature files

ismay:
Adding the transfer feature files to the Transfer PR. These are the changes from bf26891140b9cbd6ec6854e68a5cf39bfb157f71. (note: the failing tests are intentional, this just adds the feature files for the component that we're still creating)


ismay:
> Ah, there's one thing I didn't notice before. Can you rename the `transfer` folder to `Transfer` so it's consistent

Ok, done

PR #652, opened 2019-12-04
refactor: pass props explicitly instead of spreading props

ismay:
- [x] Currently I've just looked at the jsdoc for inferring which props the component needs. I will need to verify that the jsdocs are also actually accurate (they weren't for the dropdownbutton for example)
- [x] Check nested components in src/ as well (which are public, if there are any)

### DIFF
--- a/src/AlertBar.js
+++ b/src/AlertBar.js
@@ -109,7 +109,6 @@ class AlertBar extends Component {
         }
 
         const info = !critical && !success && !warning
-        const iconProps = { icon, critical, success, warning, info }
 
         return (
             <div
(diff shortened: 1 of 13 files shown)


Mohammer5:
> nested components . . . which are public

This should not be the case at all. If we ever encounter this, we should move them to `src` or discuss what's happened (maybe it just got exposed accidentally)

ismay:
> Looks good. I do wonder why we need to do this for `Checkbox/Radio/Switch*` components though. These are literally just passing their props to a `Toggle*` component (plus adding one `component` prop).

Basically because spreading props onto a child like that can be considered an anti-pattern: https://codeburst.io/react-anti-pattern-jsx-spread-attributes-59d1dd53677f

ismay:
> > nested components . . . which are public
> 
> This should not be the case at all. If we ever encounter this, we should move them to `src` or discuss what's happened (maybe it just got exposed accidentally)

Ah ok, I wasn't sure whether that was an explicit convention. Then I'll just take a look and see whether that's the case and then we should be done.

ismay:
> > > nested components . . . which are public
> > 
> > 
> > This should not be the case at all. If we ever encounter this, we should move them to `src` or discuss what's happened (maybe it just got exposed accidentally)
> 
> Ah ok, I wasn't sure whether that was an explicit convention. Then I'll just take a look and see whether that's the case and then we should be done.

Just the Select filtermenu was spreading props. It's not public, but still an anti-pattern, so I've updated that as well.

PR #655, opened 2019-12-06
refactor: positioning

Mohammer5:
This branch will be used to collect feature files, changes to our components and possibly a positioning engine (ours or third party)

ToDos (added by @HendrikThePendric 23-01-2020):
- [x] Implement **internal** Popper component, which is a react-binding to popper.js 
- [ ] Refactor `DropMenu` to use `Popper`
- [ ] Refactor `Menu` and `SubMenu` to use `Popper`, incl. flipping logic
- [ ] Refactor `Select` to use `Popper`
- [ ] Revise/refactor `Popper` API (establish a good design when refactoring the components), and expose `Popper` component

### DIFF
--- a/cypress/drafts/Select/position.feature
+++ b/cypress/drafts/Select/position.feature
@@ -1,0 +1,25 @@
+Feature: Position of select menu dropdown
+
+    Background:
+        Given the select menu dropdown has a height of 368px
+        And the select menu dropdown has a width of 280px
+
+    Scenario: Default rendering
+        Given there is enough space between the anchor's bottom and the body's bottom to fit the Select's menu
+        When the menu is opened
+        Then it is rendered below the select
+        And the left of the select is aligned with the left of the anchor
+
+    Scenario: Flipped rendering when insufficient space below
+        Given there is not enough space between the anchor's bottom and the body's bottom to fit the Select's menu
+        And there is enough space between the anchor's top and the body's top to fit the Select's menu
+        When the menu is opened
+        Then it is rendered above the select
+        And the left of the select is aligned with the left of the anchor
+
+    Scenario: A select with less than 368px available space below and above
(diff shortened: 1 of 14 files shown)


HendrikThePendric:
@Mohammer5 @varl @ismay As agreed in Oslo, we shouldn't have a "feature branch" like this. It will take too long before it gets merged into master.

Before this could go into master, I had to move the untested/future feature files into a `drafts` folder. This way we won't get failing tests on master.

The remaining tasks will be completed via separate branches of master:
> - Refactor DropMenu to use Popper
> - Refactor Menu and SubMenu to use Popper, incl. flipping logic
> - Refactor Select to use Popper
> - Revise/refactor Popper API (establish a good design when refactoring the components), and expose Popper component

cypress/drafts/Select/position.feature
```diff
@@ -0,0 +1,25 @@
...
+        When the menu is opened
+        Then it is rendered above the select
+        And the left of the select is aligned with the left of the anchor
+
+    Scenario: A select with less than 368px available space below and above
+        Given there is not enough space between the anchor's bottom and the body's bottom to fit the Select's menu
+        And there is not enough space between the anchor's top and the body's top to fit the Select's menu
+        When the menu is opened
+        Then it is rendered below the select
+        And the height of the select dropdown menu is reduced to fit within the <body> element
```
ismay: I'm interested to see in how this will pan out. I can imagine that this leads to weird behaviour and that extending beyond would be preferable. But let's see how it works out in practice.

HendrikThePendric: Yeah, if my understanding is correct, these draft feature files are meant to serve as "starting points" for the final feature files that are going to be included with the finalised component. As such it's OK to encounter things here that we are not 100% sure about.

PR #660, opened 2019-12-09
fix(deps): set dependencies provided by app-platform to peerDeps

varl:


varl:
> I get why we are taking this move but I have a few reservations about `styled-jsx`.... Typically, I think most developers would agree with the following distinction between direct and peer dependencies:

I agree with your points.

We are driven by the fact that these libraries we rely on need to be singletons, and the safest way to ensure that is to have the runtime dependency in a single place (the app, or the app platform).

Technically, we don't need to peerDep **@dhis2/prop-types**, but consider doing so to minimize the bundle.

varl:
Removing the peerDep for our prop-types, leaving it as a runtime dep.

PR #663, opened 2019-12-11
docs(changelog): fix select callback object typo

ismay:
Tiny fix for a typo in the docs I noticed whilst preparing for the demo. I assume that these changes will persist and not be overwritten in our automated release process @varl?


varl:
I actually have no idea if they persist. I assume that they do; let's test it!

ismay:
I took a look at the changelog generated for 4.1.2, it's left the changes intact, so it just appends 🎉

PR #666, opened 2019-12-12
fix(splitbutton): allow initialfocus on primary button

ismay:


HendrikThePendric:
Oh... one thing you might want to consider, but not a required change: should we perhaps include a story for `initialFocus`?

ismay:
> Oh... one thing you might want to consider, but not a required change: should we perhaps include a story for `initialFocus`?

Yeah you're right, that would be a good thing to add. I'll do that now.

ismay:
Added the splitbutton initialfocus story

PR #668, opened 2019-12-12
refactor(tabbar): add wrapper to allow for data-test

ismay:
I've considered a couple different approaches to allow for a data-test attribute to be set on the TabBar. We've attached it to the root rendered node everywhere else, so that's my goal for the TabBar as well.

Initially I tried keeping the component as is. The existing markup is:

Screenshot 2019-12-12 at 14 12 28

### Minimal approach, semantically confusing

Keeping this as-is, means attaching the data-test attr. to either the Scrollable component, or a div wrapper (which would replace the current Fragment), depending on the value of `scrollable`. That would keep the markup reasonably similar, but I thought it made for a more confusing data-test attr. Because now, depending on the value of `scrollable`, the attribute is either attached to the scrollable component, or the root div. Which are different components, and at different points in the eventually rendered markup, kind of defeating the purpose of using a data-test attribute in my opinion.

### More elaborate, semantically more correct

The fix in this PR is more elaborate and adds a div around the entire TabBar component, including the Scrollbar. The main thing I don't like about _this_ approach is that to not break existing className rules from users, we're passing className down, and not applying it to the root node. Which kind of goes against our conventions as well I think. Still, to me this seemed like the more correct approach, and later on we could move className up for a major version.

Let me know what you all think.

### DIFF
--- a/src/TabBar.js
+++ b/src/TabBar.js
@@ -1,9 +1,8 @@
 import React from 'react'
 import propTypes from '@dhis2/prop-types'
-import cx from 'classnames'
 
-import { colors } from './theme.js'
 import { ScrollBar } from './TabBar/ScrollBar.js'
+import { Tabs } from './TabBar/Tabs.js'
 
 /**
  * @module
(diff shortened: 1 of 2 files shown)


ismay:
> One possible change would be the name of the new `Bar` component that has emerged, I would have a slight preference for `Tabs`

That sounds good, changed the name to Tabs.

PR #670, opened 2019-12-12
feat(input): enable additional types on input

varl:
E.g. date from UA:

2019-12-12-192020_275x162_scrot

This adds additional types of input controls:

- date
- datetime
- datetime-local
- month
- week
- time
- search

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -231,6 +231,13 @@ Input.propTypes = {
         'email',
         'url',
         'tel',
+        'date',
+        'datetime',
+        'datetime-local',
+        'month',
+        'week',
+        'time',
+        'search',
     ]),
     valid: statusPropType,
 


Mohammer5:
I remember I discussed these additional types @cooper-joe. If I remember correctly, he consciously chose only a limited set of possible types, but I don't remember where we has said discussion or why there's only a limited set of input types..

varl:
In the old design these types didn't work very well (since the icon overlapped the internal icon), but I don't see any reason we should restrict this especially as we do not have our own calendar widgets.

Users of this library may want to use the native datepickers _even though_ we have our own internal ones.

We also need the native datepickers in e.g. Usage Analytics right now, so either we have to recompose our component there and switch out the input, or we relax the props in Input. I find the former silly in this specific case.

tl;dr:

- I don't think that these props need to be documented in the design system.
- I don't think that these props need to be heavily documented in ui-core.
- I do think they are useful even if we have calendar widgets.
- I need them in Usage Analytics right now.

cooper-joe:
This is fine by me. The limited types I discussed before (URL, password etc.) were just the ones where I would document their recommended usage in the design system. I am not against adding all input types, undocumented 👍 Depending on what happens with the calendar input stuff I may even document these.

PR #672, opened 2019-12-13
docs(api): add workaround for jsdoc bug

varl:
JSDoc has a bug where if a named import is the last statement before the
doc comment, then it fails to pick up the doc comment.

Any valid statement between the named import and the doc comment works
around the problem, such as `;` or `''`. Our code style rewrites a `''`
statement as `;('')`, and that is the shortest non-functional workaround
I have found to work around the problem.

Refs https://github.com/jsdoc/jsdoc/issues/1718
Refs https://github.com/jsdoc/jsdoc/issues/1706

Compare the result (broken on the left, correct on the right):

2019-12-13-165002_2019x1791_scrot

It also had the result that most of the modules were dropped, and the PropTypes rendered without their owner:

2019-12-13-165110_2427x2094_scrot

### DIFF
--- a/src/AlertBar.js
+++ b/src/AlertBar.js
@@ -7,6 +7,7 @@ import { Dismiss } from './AlertBar/Dismiss.js'
 import { Icon, iconPropType } from './AlertBar/Icon.js'
 import { Message } from './AlertBar/Message.js'
 import styles, { ANIMATION_TIME } from './AlertBar/styles.js'
+;('') // TODO: https://github.com/jsdoc/jsdoc/issues/1718
 
 /**
  * @module
--- a/src/AlertStack.js
+++ b/src/AlertStack.js
@@ -4,6 +4,7 @@ import propTypes from '@dhis2/prop-types'
 import cx from 'classnames'
 
 import { layers } from './theme.js'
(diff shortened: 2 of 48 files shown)


amcgee:
Naive question - would a non-JSDoc comment work?

varl:
> Naive question - would a non-JSDoc comment work?

Sadly, no.

With a normal comment the first AST node that JSDoc identifies is:

2019-12-13-174543_1297x808_scrot

With the `;('')` the first AST node is:

2019-12-13-174500_1909x1141_scrot

Which is what we expect. :thinking:

varl:
> Naive question - would a non-JSDoc comment work?

Sadly, no.

With a normal comment the first AST node that JSDoc identifies is:

2019-12-13-174543_1297x808_scrot

With the `;('')` the first AST node is:

2019-12-13-174500_1909x1141_scrot

Which is what we expect. :thinking:

varl:
> Naive question - would a non-JSDoc comment work?

Sadly, no.

With a normal comment the first AST node that JSDoc identifies is:

2019-12-13-174543_1297x808_scrot

With the `;('')` the first AST node is:

2019-12-13-174500_1909x1141_scrot

Which is what we expect. :thinking:

amcgee:
@varl :-( OK

PR #673, opened 2019-12-13
feat(modal): allow modal without actions

varl:
To allow a Modal to be rendered with only content, we need to adjust the
margin/paddings on the components.

This also cleans up the dead prop `open` which was removed in v4.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -19,7 +19,10 @@ import { ModalCard } from './Modal/ModalCard.js'
  * Use Model with the following Components:
  * ModelTitle (optional)
  * ModelContent (required)
- * ModelActions (required)
+ * ModelActions (optional)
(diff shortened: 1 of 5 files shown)


varl:
Context @cooper-joe:

2019-12-13-183732_578x1186_scrot

varl:
Not sure how well Netlify previews work so here are some example shots:

2019-12-13-183823_634x233_scrot

2019-12-13-183833_432x459_scrot

2019-12-13-183844_431x1261_scrot

cooper-joe:
Good stuff, thanks. This does loosen the usage requirements of the modal, as you said, so I will make the necessary changes to the guidelines.

varl:
Thanks @cooper-joe.

PR #675, opened 2019-12-16
fix: input and select styling

cooper-joe:
Some minor style changes to the `input` and `select` components.

I am changing the `min-height` defined for the `select`. I remember that this was discussed before and the current solution was the simplest. Right now there is a lack of consistency when using these types of user input side by side: 
Screenshot 20191216142928

With the changes in this PR, default sized select and input have the same height:
Screenshot 20191216142903

Are there any issues with these changes? [There was a slack discussion about this](https://dhis2.slack.com/archives/CBM8LNEQM/p1572261046057200), it looks like @Mohammer5 and @ismay had a call to discuss.

Ps. Are we merging straight into `master` now? I see there is no `next`. Please confirm before I merge 😉

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -21,16 +21,14 @@ const styles = css`
         user-select: text;
 
         color: ${colors.grey900};
-        background-color: transparent;
+        background-color: white;
 
         padding: 12px 11px 10px;
 
         outline: 0;
         border: 1px solid ${colors.grey500};
         border-radius: 3px;
-        box-shadow: inset 0 0 0 1px rgba(102, 113, 123, 0.15),
-            inset 0 1px 2px 0 rgba(102, 113, 123, 0.1);
-
+        box-shadow: inset 0 1px 2px 0 rgba(48, 54, 60, 0.1);
         text-overflow: ellipsis;
     }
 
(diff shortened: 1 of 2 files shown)


cooper-joe:
Thanks for the explanation @ismay. I agree with you: adjusting the min-height is worth it for the consistency when using an input and a single select. I think any difference in the height of the multi-select should be handled by adjusting its contents: the chip. I remember one of the key outcomes of the discussions being "let's not set and force a specific height on input elements", and I still agree with that.

---

Not related: I see this PR is failing all netlify checks. I've never noticed this happening before. Is it something I've changed or missed that is causing it?

ismay:
Yeah weird, this is the error: 

Screenshot 2019-12-16 at 18 13 14

I think it's failing at the `yarn build:docs` command. I see this branch is also out of date with master, and for me on master `yarn build:docs` succeeds. I assume a rebase on master will resolve the issue.

ismay:
I've rebased it on master, hope you don't mind. Let's see if this fixes it.

ismay:
That seems to have done it. It all looks good to me. I think the only thing that you could do is scope the commits to a specific area. So instead of `fix: input background color`, you could use `fix(input): update background color` or something of the sort, to indicate which area this fixes.

Other than that I think it's good to go 👍. @Mohammer5, @HendrikThePendric what do you think?

ismay:
I propose we squash this to:

Screenshot 2019-12-19 at 12 43 57

@Mohammer5, or we could merge it as is. Let me know what you prefer.

PR #676, opened 2019-12-16
refactor(node): abstract child components

ismay:
Similar to https://github.com/dhis2/ui-core/pull/674 and #664. Noticed, whilst working on the data-test attrs, that the node has child components that reside in the same file. Whereas for our other components we've abstracted them to their own subdir. Did that for node as well.

One other thing I was wondering is why we're not conditionally rendering the Arrow based on whether the node hasLeaves or not? But thought I'd leave that question for later, this is just to quickly align the organisation with the other components.


ismay:
> LGTM 👍
> 
> I'm afraid I don't know about this, I think @Mohammer5 might.
> 
> > One other thing I was wondering is why we're not conditionally rendering the Arrow based on whether the node hasLeaves or not? But thought I'd leave that question for later, this is just to quickly align the organisation with the other components.

Yeah let me know @Mohammer5, we could address it if necessary.

Mohammer5:
> One other thing I was wondering is why we're not conditionally rendering the Arrow based on whether the node hasLeaves or not? But thought I'd leave that question for later, this is just to quickly align the organisation with the other components.

I'm using the Arrow component to fill up the space, even if there's no arrow, so all node will start on the same horizontal position. Of course this could be implemented differently (either by giving some extra padding to the content or renaming the `Arrow` component to something that is less confusing), open to suggestions

ismay:
I see, because I tried conditionally rendering the Arrow, to see if something would break if not rendering it, and all the examples in the Storybook still looked ok.

ismay:
Created a follow up with a proposal for the Node component here: https://github.com/dhis2/ui-core/pull/679

PR #679, opened 2019-12-17
refactor(node): add spacer component

ismay:
Follow-up of https://github.com/dhis2/ui-core/pull/676.

I've updated the rendering of the Arrow slightly, so that the intention of rendering either an Arrow or empty space is more obvious from the code. I've also updated classes, that seemed to be old style hooks, but seemed to be no longer used. Correct me if I'm wrong though.

The integration tests have to be updated, as they still target the classes. We could hold off on merging this until the data-test attrs. are in, rebase and then use them to fix the integration tests for this PR and merge this in.

### DIFF
--- a/cypress/integration/Node/Can_be_closed/index.js
+++ b/cypress/integration/Node/Can_be_closed/index.js
@@ -11,7 +11,9 @@ Given('the Node is provided with an onClose handler', () => {
 })
 
 When('the arrow is clicked', () => {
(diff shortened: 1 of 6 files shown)


ismay:
Just so that we can add the data-attrs all in the same PR, I've changed the Node integration tests to not use the classes we were using, since they no longer exist. This is of course not an ideal integration test at all, and merely an interim solution until the data-attrs can be used for reliable targeting.

PR #689, opened 2019-12-19
docs(specs): data-table feature specs

cooper-joe:
Feature files for the data table. 

Reference: [HTML & CSS mockup of all data table styles/functions](https://codesandbox.io/s/table-styles-bvf5r)

These .feature files cover the behaviour of the data table and it's varied functions. I have not included what I consider to be 'options', because I didn't think they were classified as behaviour. For example, the .feature files do not cover:
- displaying a header or a footer
- displaying table actions (the app defines table actions, so there would be no inherent behaviour except for 'display a button'. Perhaps that counts? `Given there are actions passed to the table, Then a button for each action is displayed in the header` for example.)
- bordered cells
- status badges
- empty states

The main functionality that I did not cover that I think *should* be covered is filtering. However, I think a proper discussion is warranted as to the level of filtering functionality that will be provided with this component (none, perhaps).

Feedback much appreciated on the .feature files and the points above.

(I also created a `feat` branch for this to be merged into. I figured all data-table related stuff can live there).

### DIFF
--- a/cypress/integration/DataTable/Clickable_header.feature
+++ b/cypress/integration/DataTable/Clickable_header.feature
@@ -1,0 +1,17 @@
+Feature: Clicking a sortable column header cell toggles sorting
+
+    Background:
+        Given a table with three or more rows is rendered
+        And all column headers have sorting enbaled
+
+    Scenario: User clicks a header cell that is not currently sorting the table
+        When a header cell is clicked
+        And the clicked header cell is not currently sorting the table
+        Then the table rows are sorted low to high by the content of that column
+        And the clicked header cell changes background color during the click
+
+    Scenario: User clicks a header cell that is currently sorting the table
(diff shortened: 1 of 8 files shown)


cooper-joe:
This PR containing feature specs for the data-table component is ready for review. https://github.com/dhis2/ui-widgets/pull/202 is waiting on these to start work.

Unfortunately the E2E tests are failing with the same node error that https://github.com/dhis2/ui-core/pull/743 was giving, @Mohammer5 any idea what could be causing this? It could be my setup, it only seems to be happening on PRs I submit?

Mohammer5:
@cooper-joe I don't think it has anything to do with your PR.. Something is weird lately.
@varl No idea why it's failing here now as well...

varl:
Re-open in dhis2/ui.

PR #691, opened 2019-12-19
chore: allow multiple types per child prop-type

HendrikThePendric:
### **IMPORTANT NOTICE**
**This isn't intended to ever get merged into ui-core. If we were to do something with this, it would have to be via `@dhis2/prop-types`.**

### Context
This is a continuation of https://github.com/dhis2/ui-core/pull/690, some things have improved:
- You can now assign arrays of propTypes to the properties of the `childWithPropTypes` payload object
- If this validator encounters them, it will check if the child propType matches any and if so, it will allow it.

So this at least fixes the actual problem I encountered and described in my previous draft PR.

But I wonder if this now makes this into a fully robust solution. Is this a bulletproof way to check if a child component has implemented the correct API or not?

I can imagine that if a developer were to implement a custom Checkbox, and that Checkbox were to have a custom prop-validator for its `checked` property, we'd still have a problem. Unless the developer were to override the `CheckboxGroup.propTypes.children` prop after importing it. (and that'd be pretty hacky). 

### Conclusions
Personally, I'm still not quite sure what to do with this.... I've got a feeling that I still haven't considered a lot of cases, like conditional rendering, mapping, etc. And I don't feel confident that this is something we will ever be able to completely "get right". We are still leaning on quite a few assumptions:
- That the child will have propTypes in the first place
- That the child uses the same (as in referential equality) prop-type functions as the parent

### Implications for past work
Suppose we cannot enforce a "contract between a parent and child" in this (or a different way)... Then we have some components with interdependencies that are not enforced in any way. And if this is true, then we have to accept that we are dependent on the developer doing the right thing...  And if we accept that, I think that doesn't justify the introduction of a wrapper `<div/>` around the `<label/>` in the `ToggleGroup`, because that only decouples the style. I see no point in doing so, if the more significant aspect of the loose coupling is still present, and completely brittle/unenforced.

So IMO that means we should have gone for this implementation https://github.com/dhis2/ui-core/pull/690/commits/e081e32131ed96cea72e66cc6f1fb5567978748b and NOT this one https://github.com/dhis2/ui-core/pull/690/commits/66ebc9d14d81e607ce20ef597e328d7212eefcf1

### DIFF
--- a/src/Checkbox.js
+++ b/src/Checkbox.js
@@ -219,4 +219,4 @@ Checkbox.propTypes = {
     onFocus: propTypes.func,
 }
 
-export { Checkbox }
+export { Checkbox, uniqueOnStatePropType }
--- a/src/CheckboxGroup.js
+++ b/src/CheckboxGroup.js
@@ -3,7 +3,7 @@ import propTypes from '@dhis2/prop-types'
 
 import { statusPropType } from './common-prop-types.js'
 import { ToggleGroup } from './ToggleGroup.js'
-import { Checkbox } from './Checkbox.js'
+import { uniqueOnStatePropType } from './Checkbox.js'
 ;('') // TODO: https://github.com/jsdoc/jsdoc/issues/1718
 
 /**
@@ -77,8 +77,11 @@ const childWithPropTypes = propDefinition => (children, index) => {
     for (const key in propDefinition) {
         const prop = propDefinition[key]
         const childProp = childPropDefinition[key]
+        const isValid = Array.isArray(prop)
+            ? prop.some(p => p === childProp)
+            : prop === childProp
 
-        if (prop !== childProp) {
+        if (!isValid) {
             return new Error(
                 `Child does not implement the required props API for ${key}`
             )
@@ -91,10 +94,9 @@ const childWithPropTypes = propDefinition => (children, index) => {
 CheckboxGroup.propTypes = {
     children: propTypes.arrayOf(
         childWithPropTypes({
-            value: Checkbox.propTypes.value,
+            value: propTypes.string,
             onChange: propTypes.func,
-            // checked: Checkbox.propTypes.checked,
-            checked: propTypes.bool,
+            checked: [uniqueOnStatePropType, propTypes.bool],
         })
     ).isRequired,
 


ismay:
This reminds me a bit of the way the `Select` depends on its children (the options) having a certain api. Since there's a kind of coupling there I've written integration tests that test whether the combination of both works as expected.

That way, if someone accidentally breaks the contract between them, the test will fail. I think that we could apply the same solution to the interaction between the ToggleGroup and the Checkbox (and Radio, etc.).

As for styling interaction between both, that's not as nice to test for in my opinion. Plus there was no need for any coupling styling wise. Which is why I preferred moving that completely to the ToggleGroup (via the Spacer, but any other solution that enforces the styling without making assumptions about the children will do in my opinion).

So adding integration tests to ensure the combination of components works as expected would be my preferred solution in this case.

Mohammer5:
Yeah, integration tests sound good to me to prevent bugs in the library.

I think @HendrikThePendric tried to make sure that we have an approach that also the app dev can create custom components and will be provided with checks that make sure their implementation works.

After having seen the complexity of this, I think we should try to do that but instead throw errors in our components if the provided components do not meet certain criteria, this would probably only happen during development, not production and we wouldn't have to analyse components at runtime, just react to wrong implementations.

ismay:
> I think @HendrikThePendric tried to make sure that we have an approach that also the app dev can create custom components and will be provided with checks that make sure their implementation works.
>
> After having seen the complexity of this, I think we should try to do that but instead throw errors in our components if the provided components do not meet certain criteria, this would probably only happen during development, not production and we wouldn't have to analyse components at runtime, just react to wrong implementations.

Ah I see. Yes I agree, clear runtime errors sounds like a nice way to deal with this. We could add this to the select as well.

HendrikThePendric:
> I think @HendrikThePendric tried to make sure that we have an approach that also the app dev can create custom components and will be provided with checks that make sure their implementation works.

Yup that pretty much sums it up. However, I would only like to add these if it wouldn't require additional complexity in the component. Otherwise I'd prefer to just rely on having the appropriate documentation in place.

> After having seen the complexity of this, I think we should try to do that but instead throw errors in our components if the provided components do not meet certain criteria, this would probably only happen during development, not production and we wouldn't have to analyse components at runtime, just react to wrong implementations.

I think I know what @Mohammer5 is saying, but there is some ambiguity there:
> I think we **should** try to do that **but instead** throw errors [..]

So I am assuming that sentence was supposed to read:
> I think we **shouldn't** try to do that but instead throw errors [..]

I agree that the prop-types approach is brittle and should be avoided. If run-time checks could be implemented in a way that avoids making assumptions, I would be in favour of doing so. However, in the case of the `CheckboxGroup` I don't think there would be a way to do this in a way that wouldn't impact the component's complexity. I guess it's worth exploring.

As it stands, my hunch is that catching incorrect usage was worthwhile to investigate, but not really feasible / worth it. Instead we should probably just focus on getting the right documentation in place.

HendrikThePendric:
Closing this PR. We have come to the conclusion that enforcing the contract between parent and children on API level is not viable either via prop-type validation nor via runtime checks.

Also, we have decided to keep the `Spacer` in the ToggleGroup

PR #696, opened 2020-01-06
feat(select): append loading spinner to children instead of replacing

ismay:
This changes the behaviour of the loading spinner. When enabled, it'll be appended to the children, instead of replacing them. See this discussion: https://dhis2.slack.com/archives/CBM8LNEQM/p1577954451004500

It looks like this:

Screenshot 2020-01-06 at 14 14 02
Screenshot 2020-01-06 at 14 14 18

Todo:

- [x] Check if there are other areas that could use the new helpers (input)
- [x] Check if Select can handle null, undefined, etc. properly (later on)
- [x] Check consequences of current changes for other areas of the select

### DIFF
--- a/cypress/integration/MultiSelect/Accepts_loading.feature
+++ b/cypress/integration/MultiSelect/Accepts_loading.feature
@@ -4,11 +4,11 @@ Feature: Loading status
         Given a MultiSelect with options and a loading flag is rendered
         When the MultiSelect input is clicked
         Then the loading spinner is displayed
-        And the options are not displayed
+        And the options are displayed
 
     Scenario: The user opens a loading MultiSelect with a loading text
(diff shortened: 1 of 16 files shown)


ismay:
Thx 👍. Yeah it's tripping over the `null` when there's no loading spinner, I'll fix it before merging.

ismay:
The logic I'm using to append the loading spinner is slightly more complex than just using `&&`:

```
                    {loading
                        ? React.Children.toArray(children).push(
                              <Loading message={loadingText} />
                          )
                        : children}
```

Reason for that is that we're using React.Children.count to check if the Select is empty. And `null` and `undefined` still get counted, even though they don't get rendered. Which breaks our empty behaviour. So the above solution properly appends the Loading spinner only when necessary, and does not add anything otherwise.

PR #697, opened 2020-01-06
fix(select): fix incorrect data-test attribute

ismay:
Tiny fix, there was an incorrect data-test attribute on the menu wrapper div.


varl:
This could be bundled with #696 to only trigger a single release.

ismay:
Ah yes, I'll rebase it onto #696 when that's done 👍

PR #713, opened 2020-01-13
test: separate testing stories from user docs

ismay:
https://jira.dhis2.org/browse/TECH-280

- [x] Separate stories used for testing from stories used as user docs
- [x] Change `Default` story name where used, to prevent clashes
- [x] Remove testing code from user docs


varl:
> I'm wondering if it wouldn't be better to completely separate the testing stories from the docs stories.
> They're entirely used by cypress, so putting them into the cypress dir would make things very clear.

This does sound nice to me.

ismay:
Yeah I like that as well.

PR #735, opened 2020-01-16
feat(component-cover): introduce transparent mode and fix styles

HendrikThePendric:
### `transparent` prop
As [discussed on slack](https://dhis2.slack.com/archives/CBM8LNEQM/p1579163079038000?thread_ts=1579105721.036000&cid=CBM8LNEQM), I've added a `transparent` prop that makes the background transparent. The prop-name is still up for debate, especially @ismay expressed a desire to change this, so hoping for suggestions here.

### Styles
I've discussed this with @varl on Slack. In a nutshell, the current solution is very brittle, and only works for parents with a fixed height/width. The solution implemented in this PR also comes with a prerequisite: the parent needs to have a non-static position. On balance, I think this is a prerequisite that is easier to fulfil than the fixed height/width one, so I would say this is an improvement.

----------------

Below is a summary of some things I tried (I've just copy-pasted this from my Slack discussion with Viktor):

The implementation of the `ComponentCover` in `master` breaks if, in the parent element in `/story/componentcover--modal` has the following changes (currently `height: '400px'`):
- `minHeight: '400px'` (breaks badly)
- `height: '50%'` (breaks a little bit)
- `display: flex` (breaks badly)

What does work is `height: '50vh'`. This is consistent with what I regard as a “fixed” height in CSS: percentages are relative to a given parent, so dynamic. `vh` is just a fraction of a fixed amount, so fixed.

Now, switching to the top/right/bottom/left=0 implementation on my branch, I find that the story in its current form is broken pretty badly. We need to fix this by adding `position: 'relative'` to the parent. I’ll list the same examples I tried for the current approach:
- fixed height (i.e. current example, 400px) works
- relative height (i.e. 50%): works
- minHeight: works
- flexbox layout (i.e. display: flex): works
**But these ALL requires `position: 'relative'` to be added to the parent** (any non-static form of positioning would work)

I think this summary covers the most important scenarios. I’ve only focussed on height related stuff, but the same would obviously apply to width related styles.

Having to add a non-static position to the parent, could possibly be problematic in some scenarios. I.e. we’ve seen some interference between sticky and relative positioning when building the Select…. So I won’t claim that this new solution is 100% perfect.

However, the current solution requires a static height/width on the parent, and that’s a prerequisite that can’t be fulfilled very often.

So, on balance the new implementation is an improvement over the current.

### DIFF
--- a/src/ComponentCover.js
+++ b/src/ComponentCover.js
@@ -10,9 +10,17 @@ import { layers } from './theme.js'
  * @example import { ComponentCover } from @dhis2/ui-core
  * @see Live demo: {@link /demo/?path=/story/componentcover--circularloader|Storybook}
  */
-const ComponentCover = ({ children, className, dataTest }) => (
+const ComponentCover = ({ children, className, dataTest, transparent }) => (
     <div className={className} data-test={dataTest}>
         {children}
+        <style jsx>{`
+            div {
+                background: ${transparent
+                    ? 'transparent'
+                    : 'rgba(33, 43, 54, 0.4)'};
+                z-index: ${layers.applicationTop - 1};
+            }
+        `}</style>
         <style jsx>{`
             div {
                 display: flex;
@@ -20,12 +28,10 @@ const ComponentCover = ({ children, className, dataTest }) => (
                 justify-content: center;
 
                 position: absolute;
-
-                height: inherit;
-                width: inherit;
-
-                z-index: ${layers.applicationTop - 1};
-                background: rgba(33, 43, 54, 0.4);
+                top: 0;
+                right: 0;
+                bottom: 0;
+                left: 0;
             }
         `}</style>
     </div>
@@ -41,11 +47,13 @@ ComponentCover.defaultProps = {
  * @prop {string} [className]
  * @prop {Node} [children]
  * @prop {string} [dataTest]
+ * @prop {boolean} [transparent]
  */
 ComponentCover.propTypes = {
     children: propTypes.node,
     className: propTypes.string,
     dataTest: propTypes.string,
+    transparent: propTypes.bool,
 }
 
 export { ComponentCover }
--- a/stories/ComponentCover.stories.js
+++ b/stories/ComponentCover.stories.js
@@ -4,7 +4,7 @@ import { CircularLoader, ComponentCover, Card } from '../src'
 
 storiesOf('ComponentCover', module)
     .add('CircularLoader', () => (
-        <div style={{ width: '400px', height: '400px' }}>
+        <div style={{ width: '400px', height: '400px', position: 'relative' }}>
             <ComponentCover>
                 <CircularLoader />
             </ComponentCover>
@@ -14,8 +14,30 @@ storiesOf('ComponentCover', module)
         </div>
     ))
 
+    .add('CircularLoader - transparent', () => (
+        <div
+            style={{
+                width: '400px',
+                height: '400px',
+                border: '1px solid grey',
+                position: 'relative',
+            }}
+        >
+            <ComponentCover transparent>
+                <CircularLoader />
+            </ComponentCover>
+        </div>
+    ))
+
     .add('Modal', () => (
-        <div style={{ width: '400px', height: '400px', background: 'purple' }}>
+        <div
+            style={{
+                width: '400px',
+                minHeight: '400px',
+                background: 'purple',
+                position: 'relative',
+            }}
+        >
             <ComponentCover>
                 <div style={{ width: '200px', height: '200px' }}>
                     <Card>Some text.</Card>


src/ComponentCover.js
```diff
@@ -10,22 +10,28 @@ import { layers } from './theme.js'
...
-const ComponentCover = ({ children, className, dataTest }) => (
+const ComponentCover = ({ children, className, dataTest, transparent }) => (
     <div className={className} data-test={dataTest}>
         {children}
+        <style jsx>{`
+            div {
+                background: ${transparent
+                    ? 'transparent'
+                    : 'rgba(33, 43, 54, 0.4)'};
+                z-index: ${layers.applicationTop - 1};
```
varl: `layers.applicationTop` isn't regarded as a dynamic property since it can be resolved through a constant, so this can live in the second style tag.

ismay:
> The prop-name is still up for debate, especially @ismay expressed a desire to change this, so hoping for suggestions here.

Yeah I was mostly trying to address the concern that @cooper-joe had that people might be using it when there's content it's blocking. I was thinking that besides documentation, we could also give it a name that makes it clear how we intend this to be used.

But I'm not sure what that name would be. It's just that transparent doesn't really convey "you should only use this if it's not blocking any content". But anyway, maybe it's difficult to convey that succinctly.

ismay:
Then again, if it really shouldn't block any content whilst being transparent, you could say that we could build that in to the design by creating an `Center` component that centers in the flow (say with flexbox). That way you can be sure that it won't block content.

So if not blocking content is a requirement, maybe this component isn't the right place to implement it, if we want to be strict about it.

varl:
### API design

I don't think that low-level components needs to convey that usage restriction, that's for the usage documentation to do.

In general I want the properties to be declarative, and `transparent` is just that, e.g. I want a transparent cover for the component.

---

### Component structure

Thinking about this some more, the `ComponentCover` should match the `ScreenCover`.

With the `ComponentCover` transparent by default, we would use a shim to provide the opacity/background:

```
<ComponentCover>
  <Shim opacity="100%">
    <CircularLoader />
  </Shim>
</ComponentCover>
```

That shim could be re-used for the `ScreenCover` too.

```
<ScreenCover>
  <Shim opacity="50%">
    <CircularLoader />
  </Shim>
</ScreenCover>
```

Of course, the shim could be completely internal and toggled on/off through a `transparent` prop on both `ScreenCover` and `ComponentCover` as well. My preference would be to compose that in another molecule though so we still have access to the atoms.

varl:
The intention is for `ScreenCover` and `ComponentCover` to be **blocking** because if you **cover** something you no longer have access to it. So the "non-blocking" discussion is out-of-scope in this discussion.

HendrikThePendric:
> The intention is for ScreenCover and ComponentCover to be blocking because if you cover something you no longer have access to it.

This actually applies to my use-case in the usage-analytics app too: the content is empty, so I definitely don't need to block anything. All I actually want is to make sure the spinner is aligned horizontally and vertically. So I guess the `ComponentCover` is a bad fit with my use case anyway, and instead of adding props to it to make it so, I should have opted for creating a new component that does fits this use case.... Perhaps something like `CenteredContent`....

So, building on the **Component structure** suggested by @varl, I would like to suggest the following:
- `ScreenCover`: blocks the entire window (does nothing else)
- `ComponentCover`: blocks the entire parent (does nothing else)
- `Shim`: Provides a semi-transparent background/overlay color (I'm not sure we even
- `CenteredContent`: Vertically and horizontally aligns its content
- `CircularLoader`: Spins around and looks 😋 

And we could compose these like so:
```javascript
// Screen blocking loader
<ScreenCover>
    <Shim>
        <CenteredContent>
            <CircularLoader/>
        </CenteredContent>
    </Shim>
</ScreenCover>

// Component blocking loader
<ComponentCover>
    <Shim>
        <CenteredContent>
            <CircularLoader/>
        </CenteredContent>
    </Shim>
</ComponentCover>

// Centered spinner
<CenteredContent>
    <CircularLoader/>
</CenteredContent>
```

With this structure, I am not even sure the `Shim` needs any props. I would argue that if you need a "semi-transparent-overlay" you add it, and otherwise you just don't include it.

What do you think? Proceed like the above?

ismay:
I like it, seems like a nice distinction between the components to me. Quite clear purposes for all of them 👍

HendrikThePendric:
I just realised that, even though I think what is discussed above makes sense, these would be breaking changes. I don't see any pressing need to implement them right now.

@varl we [recently discussed on Slack](https://dhis2.slack.com/archives/CBM8LNEQM/p1578989678003800) how we could deal with breaking changes.
> we can collect the RFC doucment in notes/proposals or something too

Would this be the time to to start creating sth like this?

varl:
The only breaking change here is removing the automatic content centering from `ScreenCover` and `ComponentCover`, right?

In that case, I'd say skip the `CenteredContent` for now and avoid breaking changes.

HendrikThePendric:
We'd also break the semi-transparent background right? Even if we'd skip the `CenteredContent`....

Currently, to create a "Screen blocking loader" a dev would need to do this:
```jsx
<ScreenCover>
    <CircularLoader />
</ScreenCover>
```

And with the changes that have been proposed:
```jsx
<ScreenCover>
    <Shim>
        <CircularLoader />
    </Shim>
</ScreenCover>
```

varl:
@HendrikThePendric The `ScreenCover` is transparent today, so wouldn't break anything there. It's actually the `Backdrop` component that has a `transparent` prop and gives the `ScreenCover` the current tint.

The `ComponentCover` has a background color, but since it is really only a style change and doesn't change the API, I don't consider it a breaking change.

If we consider style changes breaking, then any change we do on appearance will be a breaking change.

HendrikThePendric:
> The ComponentCover has a background color, but since it is really only a style change and doesn't change the API, I don't consider it a breaking change.

Makes sense, I have enough info to proceed now. Thanks

HendrikThePendric:
I discussed this with @varl and we decided to take a more holistic approach to fixing this in the next major release. See https://github.com/dhis2/ui-core/issues/760 for details. Closing this one now.

PR #743, opened 2020-01-17
fix(button): button style fixes

cooper-joe:
Fixes and adjusts styling for buttons:
- adjusts global padding for all buttons
- reduces padding for small buttons
- adjusts dropdown arrow (pending fix)
- adjusts borders and fills to use ui-core colors
- buttons that use an icon only use reduced padding

There is a pending fix that I haven't quite worked out yet, I can't seem to target the dropdown arrow svg to give it a `margin-left: {spacers.dp8}`:
image

Once I figure that out this is ready for review (tips appreciated!).


cooper-joe:
@HendrikThePendric, I think it was you that worked on the dropdown button? Could you assist me in getting the last fix here? See above, I'm having trouble adding a `margin-left` to the dropdown button arrow icon.

HendrikThePendric:
I'll have a look

HendrikThePendric:
I think using a resolved style like I did in https://github.com/dhis2/ui-core/pull/743/commits/5a70bb15301a0b582401f5b029ee58752b033bdf is the cleanest solution in this case, because:
- The arrow icon component is being re-used in several places, so we don't want to add the icon styles to the icon component directly
- The `DropdownButton` renders a `Button` (component) wrapped in a `div`, and the `Button` renders a `button` (element) with an `svg` inside of it. So we could have used a global selector to reach the `div :global(svg)`. But the `children` prop of the `DropdownButton` is arbitrary content, so this could also contain an `svg`, and if it did, our global styles would affect that `svg`, and that would be bad.

cooper-joe:
Thanks for the updates @HendrikThePendric. For reference, waiting to merge this PR until the test fixes are merged in https://github.com/dhis2/ui-core/pull/734

cooper-joe:
This PR is ready for review. There have been a few rebases to include the changes in https://github.com/dhis2/ui-core/pull/734. Unfortunately I think the E2E tests are still failing.

varl:
Tests are failing on a Node test, which is slightly strange to me.

Mohammer5:
That's very weird indeed.. It's been passing before

Mohammer5:
No idea what happened. I just ran the tests again, now they completed. We should keep an eye on this. this is not something that should be happening...

varl:
@cooper-joe You can rebase this on master, then it should be good to go.

varl:
@cooper-joe You can rebase this on master, then it should be good to go.

PR #749, opened 2020-01-20
chore: port to app-scripts

varl:
This removes the need for custom babel and build configurations, and
allows us to leverage the centralised configuration in the app platform.

There are some changes necessary to make CRA and styled-jsx to work when
moving, e.g. code and tests needs to be under `src` by default.


varl:
This opts for a structure of:

```
src/
...
├── __stories__
└── __tests__
``

File: package.json [hunk 1]
```diff
@@ -32,7 +30,9 @@
         "format": "yarn format:js && yarn format:text",
         "format:js": "d2-style js apply",
         "format:text": "d2-style text apply",
-        "test": "jest"
+        "test": "d2-app-scripts test",
+        "storybook": "start-storybook -p 9009",
```
amcgee: It seems this isn't used, and is a duplicate of `yarn start:demo`, can we remove it?

varl: Yeah, Storybook CLI re-added it when I was trying to figure out what was wrong with our setup. Removed in e6e370f.

varl:
The `styled-jsx/css` error is quite annoying. It only happens in the Cypress runner. :thinking:

It's because the tests themselves actually import un-transpiled components to e.g. grab default properties to avoid hardcoding test values. Going through and fixing them now.

amcgee:
That is annoying...

One possible fix would be to run both `cypress` and `storybook` in `.d2/shell` which contains the pre-compiled source files, but that would require a bit of coordination and might not actually be easier.

varl:
> That is annoying...
> 
> One possible fix would be to run both `cypress` and `storybook` in `.d2/shell` which contains the pre-compiled source files, but that would require a bit of coordination and might not actually be easier.

There is no shell in a library, unless it's very well hidden.

amcgee:
Shoot you’re right, it would be ./build for libs

varl:
Hm, maybe we can always import from the `package.json` entry-points instead of going to `src/` in that case. :thinking:

amcgee:
Yeah that would work if we only import things exposed at the top level (probably a fair assumption for storybook)

varl:
> Yeah that would work if we only import things exposed at the top level (probably a fair assumption for storybook)

Thinking about it some more, then I don't want the tests inside of `src` if they import the package entry-point.

varl:
Alright, reimplemented on a clean master and works better. Thanks for the hint about the test separation, @ismay.

Refer to the full comment in the commit message for why all stories live under `src/` now: https://github.com/dhis2/ui-core/pull/749/commits/cf14a73b41904cde6c4da8b5b33104fa5c09f748

amcgee:
Looks like all stories are at root of `/src` or in `/src/__e2e__`, is that intentional?  Multiple comments indicated it should be `/src/__stories__` and `/src/__testing__`...

varl:
> Looks like all stories are at root of `/src` or in `/src/__e2e__`, is that intentional? Multiple comments indicated it should be `/src/__stories__` and `/src/__testing__`...

Yeah, wanted to try it out, forgot to rewrite the amended commit. I saw it recommended in the Storybook docs and figured I'd try it on for a size.

- `src/__e2e__` is for stories for Cypress.
- `src/__tests__` is for Jest.
- A bit on the fence on where to put the demo stories, maybe `src/__stories__` or `src/__demo__`?

I also played around with `src/__tests__/{unit,e2e,demo}` but Jest insisted on running (well, attempting to run) everything under `src/__tests__`.

Not sure how I feel about it. I liked having them outside of `src`; it felt clean, was easy to exclude or include from fuzzy finding based on cwd, etc.

ismay:
Looks good!

> A bit on the fence on where to put the demo stories, maybe src/__stories__ or src/__demo__?

Don't know if this has been decided already, but I'd have a slight preference for `demo` over `stories`, since the e2e dir also contains stories. But it's not important, just a preference.

Maybe this is sth for a follow up PR, but unless I'm missing something, we're not running the jest tests on ci right? That would be nice to have. Maybe running jest and cypress tests on master as well?

varl:
> Don't know if this has been decided already, but I'd have a slight preference for `__demo__` over `__stories__`, since the e2e dir also contains stories. But it's not important, just a preference.

Makes sense to me, adjusted in 4e00c7f.

> Maybe this is sth for a follow up PR, but unless I'm missing something, we're not running the jest tests on ci right? That would be nice to have. Maybe running jest and cypress tests on master as well?

Good point. Added in e510872.

ismay:
Looks good to merge to me. Only thing that's failing for me locally is the demo storybook build. Looks like the import paths for the stories in demo need to be updated.

src/__demo__/AlertBar.stories.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { storiesOf } from '@storybook/react'
-import { AlertBar } from '../src'
-import { AttachFile } from '../src/icons/AttachFile'
+import { AlertBar } from './index.js'
```
amcgee: Pretty sure all these imports need to be `../index.js` now right?

src/__demo__/AlertBar.stories.js
```diff
@@ -1,7 +1,7 @@
...
+import { AttachFile } from './icons/AttachFile'
```
amcgee: And `../icons/AttachFile` (etcetera, for all these stories)

.storybook/.babelrc
```diff
@@ -0,0 +1,8 @@
+{
```
amcgee: Can we use `extends: '@dhis2/cli-app-scripts/config/app.babel.config.js'` here instead of duplicating the plugin list?  I'm also not sure we need anything but the `styled-jsx` plugin (the `app` babel config includes syntax support for other things, but doesn't actually transpile them since CRA already does.  If we need to do the transpiling here (we might, if CRA's webpack doesn't merge configs) then we can use `dhis2/cli-app-scripts/config/babel.config.js` instead.

varl:
Gah !

.storybook/.babelrc
varl: We can do this:

```
{
    "extends": "../node_modules/@dhis2/cli-app-scripts/config/app.babel.config.js"
}
```

Using `@dhis2/cli-app-scripts/config/app.babel.config.js` directly makes it look in the wrong place:

```
Error: Cannot find module '/home/varl/dev/dhis2/libs/ui-core/.storybook/@dhis2/cli-app-scripts/config/app.babel.config.js' from '/home/varl/dev/dhis2/libs/ui-core'
```

amcgee: Could do `.babelrc.js` and use `require.resolve`, might be a bit more resilient if we bake this as a storybook wrapper lib (`../node_modules` doesn't work in monorepos)

varl: It has to be named `.babelrc` but using it according the JS config format works.

```
info => Loading custom babel config as JS
```

PR #751, opened 2020-01-21
refactor: use inline content for required indicator

ismay:
Closes a task from: https://jira.dhis2.org/browse/TECH-280

This refactors our required indicator (an asterisk) from css generated content to an inline asterisk. That makes it easier to test for us (and others) if the indicator is present in integration tests, because now we can attach a data-test attribute to it.

I've decided to put the `Required` component under `src/Label`, since it's mostly used in labels, and not to pollute the `src` dir. It's also used in the `Legend` so maybe people prefer a different location. Let me know if you do.

Implementing this required the `Checkbox` and `Switch` to accept a `required` prop. Before this we were (mostly) targeting the `label` element in those components and generating a `*` after that. Adding the new prop could be seen as a downside, but I've tried multiple approaches and this way kept the changes the simplest and is more explicit in my opinion. Previously we assumed a label would be present in the Checkbox and Switch, whereas now we're relying on a prop. Plus, apparently it's necessary for our Checkboxes and Switches to be able to show a required status, so I felt it's fair to be explicit about it.

Let me know if anything is unclear.

Added integration tests for:
- [x] SingleSelectField
- [x] MultiSelectField
- [x] CheckboxField
- [x] CheckboxGroupField
- [x] FileInputField
- [x] InputField
- [x] Legend
- [x] RadioGroupField
- [x] SwitchField
- [x] SwitchGroupField
- [x] TextAreaField

### DIFF
--- a/src/Label/Required.js
+++ b/src/Label/Required.js
@@ -1,0 +1,18 @@
+import React from 'react'
+import propTypes from '@dhis2/prop-types'
+import { spacers } from '../theme.js'
+
+export const Required = ({ dataTest }) => (
+    <span data-test={dataTest}>
+        *
+        <style jsx>{`
+            span {
+                padding-left: ${spacers.dp4};
+            }
+        `}</style>
+    </span>
+)
+
+Required.propTypes = {
+    dataTest: propTypes.string.isRequired,
+}
--- a/src/ToggleField/AddRequired.js
+++ b/src/ToggleField/AddRequired.js
@@ -1,0 +1,16 @@
+import propTypes from '@dhis2/prop-types'
+import React from 'react'
+import { Required } from '../Label/Required.js'
+
+export const AddRequired = ({ label, required, dataTest }) => (
+    <React.Fragment>
+        {label}
+        {required && <Required dataTest={`${dataTest}-required`} />}
+    </React.Fragment>
+)
+
+AddRequired.propTypes = {
+    label: propTypes.node.isRequired,
+    dataTest: propTypes.string,
+    required: propTypes.bool,
+}
--- a/cypress/integration/CheckboxField/Can_be_required.feature
+++ b/cypress/integration/CheckboxField/Can_be_required.feature
@@ -1,0 +1,5 @@
+Feature: Required status for the CheckboxField
+
+    Scenario: Rendering a CheckboxField that is required
+        Given a CheckboxField with label and a required flag is rendered
+        Then the required indicator is visible
--- a/cypress/integration/CheckboxField/Can_be_required/index.js
+++ b/cypress/integration/CheckboxField/Can_be_required/index.js
@@ -1,0 +1,11 @@
+import { Given, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a CheckboxField with label and a required flag is rendered', () => {
+    cy.visitStory('CheckboxField', 'With label and required')
+})
+
+Then('the required indicator is visible', () => {
+    cy.get('[data-test="dhis2-uicore-checkboxfield-required"]').should(
+        'be.visible'
+    )
+})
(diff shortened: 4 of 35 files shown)


varl:
Do we want to use a `required` prop to toggle the `Required` component or should we use the same convention we do for other components? Namely:

```
{ required && <Required /> }
```

ismay:
> Do we want to use a `required` prop to toggle the `Required` component or should we use the same convention we do for other components? Namely:
> 
> ```
> { required && <Required /> }
> ```

Ah yeah, that's better 👍

src/Label/Required.js
```diff
@@ -0,0 +1,18 @@
...
+        <style jsx>{`
+            span {
+                padding-left: ${spacers.dp4};
+            }
+        `}</style>
+    </span>
+)
+
+Required.propTypes = {
+    dataTest: propTypes.string.isRequired,
```
varl: No default value for the `dataTest` prop in this case? Maybe not needed as this is an internal component?

ismay: Yeah for the child data-test attributes (like the buttons in SplitButton, or the remove icon in the Chip), I've not set a default since they're always a child of another component, and that component should set the data-test attribute.

ismay: But since it has to be set I've made it required for all of them.

varl: Cool, thanks for explaining the reason.

ismay:
> I'm not sure about this change. Imo the *Field components should accept the required and the * components should not. I don't really see why this would be difficult to implement, could you elaborate?

So I initially wanted to implement this in the ToggleField, that seemed like it would work. So this is the ToggleField:

```
<Field className={className} dataTest={dataTest}>
    <ToggleComponent
        value={value}
        label={label}
        name={name}
        tabIndex={tabIndex}
        onChange={onChange}
        onFocus={onFocus}
        onBlur={onBlur}
        checked={checked}
        disabled={disabled}
        valid={valid}
        warning={warning}
        error={error}
        dense={dense}
        initialFocus={initialFocus}
        required={required}
    />

    {helpText && <Help dataTest={`${dataTest}-help`}>{helpText}</Help>}

    {validationText && (
        <Help
            error={error}
            warning={warning}
            valid={valid}
            dataTest={`${dataTest}-validation`}
        >
            {validationText}
        </Help>
    )}
</Field>
```

So I thought I'd insert the required indicator after the ToggleComponent. But, the ToggleComponents (Switch and Checkbox) are both `display: flex` so the indicator ends up on the next line.

So the display: flex is used in the Switch and Checkbox for aligning the svg with the text, so I can't easily do something else there. I've committed a sketch where I've set them to inline-flex, which preserves the alignment inside the Checkbox and Switch and allows the asterisk to be rendered next to it, but then you're dealing with aligning the asterisk with the label text for the Checkbox and the Switch.

ismay:
Hendrik suggested an approach that I prefer over the one I chose initially. It allows us to keep the public api as is, and still ticks all the boxes. I've added it in the last commit. I'll notify the original reviewers in #ui once CI has succeeded.

src/ToggleField/AddRequired.js
```diff
@@ -0,0 +1,22 @@
...
+        return (
+            <React.Fragment>
+                {label}
+                <Required dataTest={`${dataTest}-required`} />
+            </React.Fragment>
+        )
+    }
+
+    return label
+}
```
Mohammer5: Really nitpicky, but I'd prefer

```js
export const AddRequired = ({ label, required, dataTest }) => (
    <React.Fragment>
        {label}
        {required && <Required dataTest={`${dataTest}-required`} />}
    </React.Fragment>
)
```

ismay: Yeah nice 👍

PR #752, opened 2020-01-21
feat: rich text support

cooper-joe:
PR for implementing the rich text support.

I've started with some feature files for keyboard shortcuts and clickable buttons, probably to be implemented via a toolbar. The parser is being handled in ui-widgets.

To do:
- [x] feature files
- [x] [design specs](https://sleepy-yalow-3c0c76.netlify.com/rich-text)
- [ ] implementation


cypress/integration/RichTextInput/Rich_Text_Keyboard_Shortcuts.feature
```diff
@@ -0,0 +1,34 @@
...
+      | ctrl+b   | bold   |
+      | cmd+b    | bold   |
+      | ctrl+i   | italic |
+      | cmd+i    | italic |
+
+  Scenario Outline: The user uses a keyboard shortcut without text selected
+    Given there is focus inside the rich text enabled input
+    When the user presses <shortcut>
+    Then the <format> tags are added
+    And the text cursor is in the middle of the added tags
```
Mohammer5: This one is a bit unclear. What are the tags added to? Around the word under the cursor? The whole line? (Like when selecting "Headline", "Paragraph", etc.) What happens if the cursor is on a space character?

I assume this is meant to change the word under the cursor, so I think we should specify that

```feature
And the cursor is on a word
# ...
Then the <format> tags surround the word under the cursor
```

cypress/integration/RichTextInput/Rich_Text_Toolbar.feature
```diff
@@ -0,0 +1,68 @@
+Feature: Rich text toolbar
+
+  Scenario Outline: The user clicks formatting buttons with text selected
+    Given there is some text selected
```
Mohammer5: I think this should be `Given there is some plain text selected` (either `plain` or `unformatted`).

cypress/integration/RichTextInput/Rich_Text_Keyboard_Shortcuts.feature
cooper-joe: Good point. You're right, if the cursor is inside a word then that word will get the formatting. I've adjusted that one scenario to account for this and added a new scenario for the `inside a word` example. I've also added this to the toolbar file too in: `The user clicks a formatting button with the cursor inside a word`.

cypress/integration/RichTextInput/Rich_Text_Toolbar.feature
cooper-joe: Indeed, that's important for these to work. Added to all scenario where appropriate.

cooper-joe:
[Design specs for the rich text user interface](https://sleepy-yalow-3c0c76.netlify.com/rich-text) (toolbar) are complete.

varl:
Re-open against dhis2/ui.

PR #758, opened 2020-01-22
feat(popper): introduce popper.js based Popper component and e2e tests

HendrikThePendric:
This PR introduces the Popper. It's very basic at the moment, but this is sufficient to get all the tests that were written for it to pass. I expect that once we start using it in our components we'll have to start adding more and more complexity.

We're on popper.js version 2.0.0-rc.3. I chose that over going for a v1, which would require us to rewrite stuff once v2 becomes final.

@cooper-joe: I've made some substantial changes to the feature file. Could you please check whether you agree with my changes?


HendrikThePendric:
I don't think the has-external-dependency is a criterium we use to differentiate between these two libs. If I recall correctly, something should go in ui-widgets if:
- It reaches out to the server (i.e. HeaderBar, OrgUnitTree)
- Or it contains domain/organisation specific logic (i.e. expression-manager, period-picker, formula-editor)

And even if this distinction turns out to be incorrect, I think we should make an exception here, because this will power quite a few of the components in ui-core and as far as I am aware, ui-core should definitely not depend on ui-widgets.

varl:
I feel that adding a dependency to ui-core is a defeat, and that it opens the Dependency Floodgates, and that we are use this as a precedence to adding _other_ external dependencies to ui-core...

I still think that this is an exception, and that we should allow another dependency to be added to ui-core.

I have one criteria that must be upheld: the dependency is 100% internal, meaning that if we have to, we could remove it and switch it out with something else.

HendrikThePendric:
> I have one criteria that must be upheld: the dependency is 100% internal, meaning that if we have to, we could remove it and switch it out with something else.

Technically speaking, we are meeting this criterium already. However, I have chosen to follow the popper.js API pretty closely when naming and structuring the props. It might be better to structure these differently:
- It might be better to lift (some of) the properties out of the `options` object
- The modifiers api is quite specific to popper.js. Each modifier provides a specific behaviour and can be configured. I have to give this one some thought
- I like the way popper js describes the different positions, I think we should just keep this. If we switch to some other lib, we'd have to map the popper.js position-names to ones that that lib is using

varl:
> > I have one criteria that must be upheld: the dependency is 100% internal, meaning that if we have to, we could remove it and switch it out with something else.
> 
> Technically speaking, we are meeting this criterium already. However, I have chosen to follow the popper.js API pretty closely when naming and structuring the props.

I know, I just wanted to write it out explicitly. :+1: 

I think following the Popper API is fine for now, I imagine they have thought quite a bit about the names so we don't have reinvent the wheel here without good reason.

HendrikThePendric:
> I think following the Popper API is fine for now

Currently this is also an internal component, we are not exporting it from `./src/index.js`. Ultimately, I think it'll be good to expose it though. I've added a ToDo list in the main branch https://github.com/dhis2/ui-core/pull/655#issue-349760040. If we follow the steps there, I think this current PR is ready to go into `Positioning`.

You guys agree? @Mohammer5 @ismay @varl

HendrikThePendric:
I interpreted the 👍 from @varl to my last comment https://github.com/dhis2/ui-core/pull/758#issuecomment-577585142 as his approval

PR #759, opened 2020-01-22
feat(node): add icon prop to Node to allow icon customization

Mohammer5:
With this change the arrow icon of the node can be replace (e. g. with a loading spinner; a required change from the org unit tree).

This also "fixes" the circular loader so that the svg shrinks if the size has been changed by an outside component.


varl:
Shouldn't `Leaves` be `Leaf`? It doesn't make sense to have a component in plural..

HendrikThePendric:
> Shouldn't `Leaves` be `Leaf`? It doesn't make sense to have a component in plural..

Good point... I think this component contains several leaf-nodes, hence the plural... Perhaps LeafList or sth would be better....

Mohammer5:
Hmm yeah the discussion about singular vs plural always has been a controversial one. Technically & theoretically it can have multiple leaves, but also just one or none.

I don't really care what we go for, the only domains that I know about that have an unwritten "standard" on this are databases and redux stores, which use the singular, so I wouldn't mind changing it to singular @varl @HendrikThePendric

varl:
The component is entirely internal, right? So we could rename it later if we want?

Mohammer5:
@varl yes, it's internal, but the `data-test` attribute depends on it. I guess changing that one could be viewed as breaking change (as it could break tests)

Mohammer5:
I guess the current state if kind of ok. It's minor details. If you agree, I'll rebase this PR onto master @varl

varl:
We can rename the component but leave the `data-test` name as-is and it won't be a breaking change. I don't think the `Leaves` name is a particularly descriptive one, but it's better than the generic `Content` which is overused everywhere, so let's go with it. :+1:

PR #773, opened 2020-01-31
chore: create stubs inside stories to not manipulate tested environment from the outside

Mohammer5:
# Status quo on master

So right now, in order to test whether the event handlers work, we create a stub in cypress and attach it to the `window` object of the testing story.  This approach works but manipulates the environment to be tested.

# What this PR changes

We have `window.Cypress.cy.stub` available in the stories, so we can make use of it,
so instead of creating the stub in the test, we create the stub in the story, attach it to the window object (so cypress has access to it) and then use it as the callback.
We also won't have to create anonymous functions anymore, we can pass in the stub directly because we're initializing it before the story is rendered.

### DIFF
--- a/cypress/integration/AlertBar/Accepts_hidden_cb/index.js
+++ b/cypress/integration/AlertBar/Accepts_hidden_cb/index.js
@@ -3,12 +3,6 @@ import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
 
 Given('an Alertbar with onHidden handler is rendered', () => {
     cy.visitStory('Alertbar', 'With onHidden')
-
-    cy.window().then(win => {
-        // The property has to be present to allow cy.stub
-        win.onHidden = () => {}
-        cy.stub(win, 'onHidden')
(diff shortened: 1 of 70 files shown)


varl:
@Mohammer5 Interesting. Could you write a description that outlines this technique? Why it works, how and what the benefits are? Would be nice to have in the PR.

Mohammer5:
@varl done

Mohammer5:
@varl @HendrikThePendric as discussed

Mohammer5:
@varl @HendrikThePendric as discussed in the meeting, instead of doing some weird stuff in the dom, we're using cypress' stubbing functionality just inside the story

HendrikThePendric:
Hey @Mohammer5 should this one perhaps be closed?

Mohammer5:
No, it just changed, I have to change the description & title though, thanks for the reminder.

varl:
I think the new approach is understandable and less "magic" than what we have now.

PR #774, opened 2020-01-31
feat: add alignment options to Modal and ScreenCover

### LINKED ISSUE #576: Modal positioning

Mohammer5:
Fixes #576

Question: Should I outsource "positioningPropType" to `src/common-prop-types.js`?


src/Modal/ModalCard.js
```diff
@@ -5,23 +5,33 @@ import { resolve } from 'styled-jsx/css'
...
-        height: auto;
-        max-height: 95vh;
         flex-direction: column;
+        height: auto;
         left: 50%;
-        max-width: calc(100vw - ${2 * spacers.dp24}px);
-        max-height: calc(100vh - ${2 * spacers.dp24}px);
+        top: 0;
+        max-height: 95vh;
+        max-height: calc(100vh - ${2 * spacersNum.dp64}px);
```
edoardo: Duplicate rule?

Mohammer5:
Default has changed to top, other than that, will do @HendrikThePendric

Mohammer5:
@edoardo Fixed the issue and outsourced the prop.
@HendrikThePendric renamed. Could you have a look again please

src/Modal.js
```diff
@@ -1,10 +1,11 @@
...
 import { createPortal } from 'react-dom'
+import React from 'react'
+import propTypes from '@dhis2/prop-types'
 
-import { sizePropType } from './common-prop-types.js'
-import { ScreenCover } from './ScreenCover.js'
 import { ModalCard } from './Modal/ModalCard.js'
+import { ScreenCover } from './ScreenCover.js'
+import { insideAlignment } from './common-prop-types'
+import { sizePropType } from './common-prop-types.js'
```
HendrikThePendric: on one line

HendrikThePendric: Also perhaps keep using the naming convention that's already present, i.e. use the `PropType` suffix

src/Modal/ModalCard.js
```diff
@@ -33,19 +45,17 @@ const cardBoxStyle = resolve`
...
-            {cardBoxStyle.styles}
-        </Card>
-    )
-}
+export const ModalCard = ({ children, small, large, position }) => (
+    <Card
+        className={cx(cardBoxStyle.className, 'modal-card', position, {
+            small,
+            large,
+        })}
```
HendrikThePendric: Perhaps I am overlooking sth, but I don't see how the classList will ever get a `top` class here...

Mohammer5: Default prop is missing :+1:

Mohammer5:
@HendrikThePendric Could you have a look at this one again?

PR #780, opened 2020-02-03
ci(cypress): run tests on push instead of pr

ismay:
Running cypress and jest unit tests on push instead of pr, so that we run them on master as well.


ismay:
Waiting on #781, so tests pass before we merge this.

PR #781, opened 2020-02-03
fix: use "old" Node arrow icon & rename arrow to toggle

Mohammer5:
This will fix the currently failing cypress tests


Mohammer5:
@cooper-joe approved the additional click-functionality, so merging this one

PR #787, opened 2020-02-06
feat: introduce file-input-field-with-list

HendrikThePendric:
Apart from having to squash my commits, I think the `FileInputFieldWithList` is done now. For the bigger context, please see [the issue description in JIRA](https://jira.dhis2.org/browse/TECH-288). This PR actually contains changes for the first two subtasks mentioned there:
> 1. Attach onBlur and onFocus to button for FileInput
> 2. Introduce FileInputWithList in ui-core

I should have perhaps prepared a separate PR for each task, but since I haven't I'll just try to explain it below:
- Currently our `FileInput` is the only input component which doesn't have support for `initialFocus`, `onFocus` and `onBlur`, so that needed to be added
- Since the visible UI component for the `FileInput` is our `Button`, I needed to add those props there, and then "propagate" these props upwards all the way to the `FileInputFieldWithList`.
- So I worked on that first and then introduced the `FileInputFieldWithList` component with accompanying features and tests.

Hope the above explanation helps.


src/FileInputFieldWithList.js
```diff
@@ -0,0 +1,152 @@
...
+*/
+
+class FileInputFieldWithList extends Component {
+    handleChange = ({ files: fileList }, event) => {
+        const { onChange, name } = this.props
+
+        onChange({ files: this.updateFileArray(fileList), name: name }, event)
+    }
+
+    updateFileArray(fileList) {
```
Mohammer5: I think this one could use a jsdocs comment:
```js
/**
 * @param {FileList} fileList
 * @returns {File[]}
 */
```

src/FileInputFieldWithList.js
```diff
@@ -0,0 +1,152 @@
...
+
+        return [...files, ...fileList].reduceRight(
+            (acc, file) => {
+                if (!acc.unique.has(file.name)) {
+                    acc.unique.add(file.name)
+                    acc.files.unshift(file)
+                }
+                return acc
+            },
+            { files: [], unique: new Set() }
```
Mohammer5: Is there a reason why you used `reduceRight` and `unshift` instead of `reduce` and `push`?

src/FileInputFieldWithList.js
```diff
@@ -0,0 +1,152 @@
...
+            warning,
+        } = this.props
+
+        return (
+            <FileInputField
+                accept={accept}
+                buttonLabel={buttonLabel}
+                className={className}
+                dataTest={dataTest}
+                disabled={disabled || (!multiple && files.length >= 1)}
```
Mohammer5: This means that the button is disabled once a file has been selected when multiple is not true.
When thinking about people who are not as versatile with how the web works as we are, then they might think, that they can't replace the file anymore, because the button is disabled, all they can do is removing the existing file.

What I really want to say is: This way seem correct/good from a technical perspective, but it requires the user to think, which is not so good imo.

@cooper-joe You're our UX guy, what do you think about that?

cypress/integration/FileInputFieldWithList/Deduplicates_the_file_list.feature
```diff
@@ -0,0 +1,7 @@
+Feature: The FileInputFieldWithList deduplicates the file list
+
+    Scenario: The file list is deduplicated
+        Given a FileInputFieldWithList with a file list including file.md is rendered
+        When the file file.md is selected
+        Then the onChange handler is called
+        Then the onChange handler's payload contains a single entry for file.md
```
Mohammer5: This is quite "hard" to read, I'd prefer:

```feature
Given a FileInputFieldWithList with file.md is rendered
When the file "file.md" is selected
# ...
```

Mohammer5: Also the next feature file says "without multiple", so I think this one needs to mention, that this one can hold multiple files, otherwise it's a bit ambiguous & this and the other feature files contradict each other.

cypress/integration/FileInputFieldWithList/Disables_button_when_full.feature
```diff
@@ -0,0 +1,5 @@
+Feature: The FileInputFieldWithList disables its button when full
+
+    Scenario: The button is disabled when the fileInput is full
+        Given a FileInputFieldWithList without multiple and a single file is rendered
```
Mohammer5: This could/should be split into two steps:

```feature
Given a FileInputFieldWithList without multiple is rendered
And the FileInputFieldWithList holds a file
```

src/FileInputFieldWithList.js
```diff
@@ -0,0 +1,152 @@
...
+        )
+    }
+}
+
+FileInputFieldWithList.defaultProps = {
+    dataTest: 'dhis2-uicore-fileinputfieldwithlist',
+    files: [],
+}
+
+FileInputFieldWithList.propTypes = {
```
Mohammer5: Needs jsdocs

cypress/integration/FileInputFieldWithList/Deduplicates_the_file_list.feature
Mohammer5: You could split this even more and reuse another step:

```feature
Given a FileInputFieldWithList with multiple files is rendered
And the FileInputFieldWithList holds a file "file.md"
When the file "file.md" is selected
# ...
```

src/FileInputFieldWithList.js
```diff
@@ -0,0 +1,152 @@
...
+
+import { FileInputField } from './FileInputField.js'
+import { FileListItemWithRemove } from './FileInputFieldWithList/FileListItemWithRemove.js'
+import { statusPropType, sizePropType } from './common-prop-types.js'
+
+/* 
+NOTE:
+A JavaScript FileList instance is read-only, so we cannot add files to it
+FileList also doesn't have a .map method so by destructuring the FileList
+instance into an array we can add, remove and map
```
ismay: So eventually we can remove the comment above, right?

HendrikThePendric: @ismay, I'm not sure I understand you question correctly. If you mean that this comment can be removed and replaced by JSdoc comments, then YES I agree.

src/FileInputFieldWithList.js
cooper-joe: I think the current implementation is the simplest solution. `Replacing a file` is a more advanced concept than `Add>Delete>Add again`. 

Allowing file replacement would open a more difficult problem: how do we inform to the user that only one file is allowed, even though they can "upload" many? The simpler solution is to ensure the user understands the concept by enforcing the `Add > Delete > Add again` method, thereby avoiding any confusion with one vs. multiple files being uploaded.

Disabling the button is intentionally slowing down the user, highlighting to them that they can’t upload more files without removing the one that is already there.

src/FileInputFieldWithList.js
ismay: I might have misunderstood, but I thought that instead of using an array the intention was to move to using a FileList object (appending to it by creating a new FileList)? Or does that not apply to this section? That's why I thought that eventually this comment could be removed as well.

src/FileInputFieldWithList.js
HendrikThePendric: I've looked into this some more and will past a section of our Slack conversation below:

The current setup works, but it is quite important to realise that:
1. The File instances connected to an input is a dynamic reference to a file in the file system, so whenever the file is updated, that update is “propagated to the input”. For example, the `lastModified` date etc are updated.
2. When adding the same file to our FileInput a second time, we create a **new reference** to the same file, i.e. we have two references to the same thing. The references themselves are in-equal but the underlying file is the same.

So what to conclude from that:
1. We only need to make sure we de-duplicate the list. I previously thought I needed to make sure we use the “most recently added” version of the file in case of file-edits etc. But this is not the case.
2. For spotting duplicates we can’t use object equality. But, having said that, checking both the `name` and `lastModified` property on the file should be good enough… (in a later part of the discussion we decided to add `size` too.

I've updated the implementation to reflect the above.

HendrikThePendric: To clarify: this method needed some attention anyway, and when doing so, I got rid of the `unshift` and `reduceRight`.

src/FileInputFieldWithList.js
HendrikThePendric: @ismay thanks for clarifying. You are correct in thinking that we did consider switching to a `FileList` instance. However, after looking into this in more detail, I didn't seem it would offer any benefits when handling the form data, and at the same time it would complicate the component code a bit. So I kept things the way they were.

HendrikThePendric: Below is a copy of my full motivation which I provided to @varl on Slack.
> I’ve looked into that Array-of-File-instances VS FileList-instance stuff quickly and think we should keep using the current array-based approach. I’ve looked at how you can append several files to a `FormData` instance, and all I see is examples that loop over the FileList to append stuff using the bracket notation, i.e.
>
>```javascript
>formData.append('userpic[]', myFileInput.files[0], 'chris1.jpg');
>formData.append('userpic[]', myFileInput.files[1], 'chris2.jpg');
>```

> source: https://developer.mozilla.org/en-US/docs/Web/API/FormData/append

> ```javascript
> var ins = document.getElementById('fileToUpload').files.length;
> for (var x = 0; x < ins; x++) {
>    fd.append("fileToUpload[]", document.getElementById('fileToUpload').files[x]);
> }```

> https://stackoverflow.com/a/14908250/1143502

> Since it seems we need to append every file individually to the FormData, using an array is equally convenient as using a FileList instance. And since it is more convenient from the perspective of the `FileInputWithList` component to keep using an array, I would say we stick to that.

src/FileInputFieldWithList.js
Mohammer5: Fair enough. Now I'm wondering if there's a way now we can let the user know that he has to `Add > Delete > Add`. Maybe by changing the text of the button (optionally by default, app could override this behavior) or by displaying a helper text once a file has been selected like "If you want to add a different file instead, please remove the existing one first"

src/FileInputFieldWithList.js
ismay: Ah I see. Ok yeah, makes sense, thanks for explaining!

HendrikThePendric:
@Mohammer5 this one is ready for a re-review.

cypress/integration/FileInputFieldWithList/Deduplicates_the_file_list/index.js
```diff
@@ -0,0 +1,41 @@
...
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('the list contains the file duplicate.md', () => {
+    cy.contains('duplicate.md').then($label => {
+        expect($label).to.have.lengthOf(1)
+        expect($label).to.have.class('label')
+        expect(
+            $label.closest('[data-test="dhis2-uicore-filelistitem"]')
+        ).to.have.lengthOf(1)
+    })
```
Mohammer5: I think we should use cypress' dom assertions when expecting some dom scenarios:
```js
cy.contains('duplicate.md')
  .should('have.lengthOf', 1)
  .should('have.class', 'label')
  .closest('[data-test="dhis2-uicore-filelistitem"]')
  .should('have.lengthOf', 1)
```

But this is not a biggie, so wouldn't mind if you don't change it

cypress/integration/FileInputFieldWithList/Deduplicates_the_file_list/index.js
```diff
@@ -0,0 +1,41 @@
...
+})
+
+When('the file duplicate.md is selected', () => {
+    cy.get('[data-test="dhis2-uicore-fileinput"] input').then($input => {
+        // name, lastModified, size and type are equal to the file created in the story
+        const duplicate = new File([], 'duplicate.md', {
+            lastModified: 1581348175357,
+            type: 'text/markdown',
+        })
```
Mohammer5: Duplicating these values can be tricky. If the test falsy succeeds when `lastModified` is accidentally different, it'd be very hard to spot.

Alternatively you can wrap and alias the previously added file with `cy.wrap(...).alias`, access it hiere and use the same `lastModified` value.

I haven't tried async/await in cypress yet, but I'd assume it to work, then it'd be a bit more readable than having nested `cy.get`.

cypress/integration/FileInputFieldWithList/Disables_button_when_full/index.js
```diff
@@ -0,0 +1,17 @@
+import { Given, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a FileInputFieldWithList without multiple is rendered', () => {
+    cy.visitStory('FileInputFieldWithList', 'Single file - with file')
+})
+
+Given('the FileInputFieldWithList holds a file', () => {
+    cy.get('[data-test="dhis2-uicore-filelistitem"]').then($fileListItem => {
+        expect($fileListItem).to.have.lengthOf(1)
+    })
```
Mohammer5: You could use the dom assertions here as well, no need for `then`

cypress/integration/FileInputFieldWithList/Disables_button_when_full/index.js
```diff
@@ -0,0 +1,17 @@
...
+})
+
+Then('the button is disabled', () => {
+    cy.get('[data-test="dhis2-uicore-button"]').then($btn => {
+        expect($btn).to.have.attr('disabled')
+    })
```
Mohammer5: Same as above

cypress/integration/FileInputFieldWithList/Displays_files_in_a_list/index.js
```diff
@@ -0,0 +1,15 @@
...
+    cy.get('[data-test="dhis2-uicore-filelistitem"] .label').then(
+        $fileListLabels => {
+            const expectedFileNames = ['test1.md', 'duplicate.md', 'test2.md']
+            const fileNames = $fileListLabels
+                .map((_, el) => el.innerHTML)
+                .toArray()
+
+            expect(fileNames).to.deep.equal(expectedFileNames)
+        }
+    )
```
Mohammer5: Same as above

cypress/integration/FileInputFieldWithList/Files_can_be_removed/index.js
```diff
@@ -0,0 +1,23 @@
+import '../common'
+import { When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+When('the remove handle behind a file is clicked', () => {
+    cy.contains('test1.md').then($label => {
+        $label
+            .siblings('[data-test="dhis2-uicore-filelistitem-remove"]')
+            .click()
+    })
```
Mohammer5: Same as above.
If cypress doesn't offer `cy.get(...).siblings(...)`, you can always use [`cy.get(...).invoke(...)`](https://docs.cypress.io/api/commands/invoke.html#Syntax)

cypress/integration/FileInputFieldWithList/Files_can_be_removed/index.js
```diff
@@ -0,0 +1,23 @@
...
+    "the onChange handler's payload does not contain an entry for that file",
+    () => {
+        cy.get('@payload').then(payload => {
+            const files = payload.files
+            expect(files).to.have.lengthOf(2)
+
+            const filesWithNameFileMd = files.filter(f => f.name === 'test1.md')
+            expect(filesWithNameFileMd).to.have.lengthOf(0)
+        })
+    }
```
Mohammer5: You can make use of cypress' `its` method:

```js
cy.get('@payload').its('files')
  .should('have.lengthOf', 2)
  .then(files => files.filter(f => f.name === 'test1.md'))
  .should('have.lengthOf', 0)
```

cypress/integration/FileInputFieldWithList/common/index.js
```diff
@@ -0,0 +1,26 @@
...
+        expect(calls).to.have.lengthOf(1)
+
+        const callArgs = calls[0].args
+        expect(callArgs).to.have.lengthOf(2)
+
+        const payload = callArgs[0]
+        expect(Object.keys(payload)).to.include.members(['files', 'name'])
+
+        cy.wrap(payload).as('payload')
+    })
```
Mohammer5: I'm sure there's a "neat" cypress way for this as well

src/__e2e__/FileInputFieldWithList.stories.js
```diff
@@ -0,0 +1,57 @@
...
+const singleFileArray = [new File([], 'dummy.txt')]
+
+const multipleFileArray = [
+    new File([], 'test1.md'),
+    new File([], 'duplicate.md', {
+        lastModified: 1581348175357,
+        type: 'text/markdown',
+    }),
+    new File([], 'test2.md'),
+]
```
Mohammer5: I see now why you've duplicated the `lastModified`. You could attach this one to the `window` object to have access in cypress.

cypress/integration/FileInputFieldWithList/common/index.js
HendrikThePendric: This can be addressed in another PR. I have simply taken this from the FileInput code.

cypress/integration/FileInputFieldWithList/Deduplicates_the_file_list/index.js
HendrikThePendric: I've attached sth to the window from the story as suggested in https://github.com/dhis2/ui-core/pull/787#discussion_r377589620

src/__e2e__/FileInputFieldWithList.stories.js
HendrikThePendric: Good suggestion 👍 I've attached `duplicateFileConstructorArgs` to the window. This means the pre-populated File instance is different from the new File instance, but the files are constructed with the same arguments, so the reference file appears the same. This as close as we can get to reality IMO.

PR #790, opened 2020-02-07
test: improve cypress test coverage

ismay:
This PR aims to increase the test coverage. I'm focusing mostly on testing the public api of our exported components where it's not style-related, so we can be reasonably confident that if the tests succeed, we haven't broken anything there.

To be completely confident, and since a lot of our components' functionality is also styling, we should eventually probably also have proper screenshot tests. That way it'd also be easier to test whether a suite of components looks and works as expected (TabBar + Tabs, AlertStack + AlertBar, etc.), which is difficult to do purely in code, as so much of what the components are supposed to do is visual.

edit: As Jan-Gerke mentioned previous week, about the Cypress team seeing their tool as allowing for visual unit tests (if I remember correctly), the tests here are mostly asserting that the public api for our components hasn't regressed, in isolation. So more unit-tests. As mentioned above, I think we'll still need to add more proper integration tests, that test the components composed.

Commits will be squashed to `test: add cypress tests` (feel free to suggest otherwise)

See: https://jira.dhis2.org/browse/TECH-280.


Mohammer5:
These all look quite good to me, but I'm wondering why you didn't use jest for most of them, especially the "visible" and "children" ones. They execute much faster than cypress tests.

ismay:
> These all look quite good to me, but I'm wondering why you didn't use jest for most of them, especially the "visible" and "children" ones. They execute much faster than cypress tests.
> 
> My preference is to use cypress for behavior only and keep everything else in jest, as that should be testable independently of the browser

Mostly because with cypress we'll be able to catch regressions that have to do with the browser or styling. Say someone adds some styles that don't prevent the children from rendering, but prevent them from being visible. With cypress we'd catch that.

If it'd be purely a test of internal logic, without having anything to do with the actual display and rendering, I'd also prefer jest.

ismay:
I think that once we have screenshot tests in place (or another type of integration test to reliably assert that our suites of components behave as expected), we can probably remove some tests, like whether the children render. Or instead of deleting we could move them to jest.

We could also do that already, but I guess then the tradeoff is that it's a slightly less realistic test (but it will run faster).

cypress/integration/AlertBar/Accepts_icon.feature
```diff
@@ -0,0 +1,13 @@
+Feature: The AlertBars accepts an icon prop
```
HendrikThePendric: Perhaps this test could be extended. The internal logic is like this (in order of priority):
- If `icon` is `false`, don't show an icon
- If `icon` is an element, show that element
- Otherwise show a status icon

So I think we should either add a test for the StatusIcon component, or we should check if the correct status-icon is rendered for the corresponding statusProp.

HendrikThePendric: Actually, looking at the `AlertBar` tests in general, I think it'd be good to add a section about statuses. Each status has a very different appearance...

cypress/integration/AlertStack/Shows_alertbars.feature
```diff
@@ -0,0 +1,5 @@
+Feature: Shows AlertBars
```
HendrikThePendric: One point that is important (see design specs) is that the AlertBars are displayed in reversed order (i.e. this will mean the most recent one is at the top). Perhaps we could just check that the div has these CSS rules?
```css
display: flex;
flex-direction: column-reverse;
```

cypress/integration/ButtonStrip/Accepts_children.feature
```diff
@@ -0,0 +1,5 @@
+Feature: The ButtonStrip renders children
```
HendrikThePendric: What the ButtonStrip does, is provide some spacing to its children. So I'd like to see a test that asserts that. Having said that, I know this component is due to be removed and replaced by some generic spacing component. So you can probably keep this one as is.

cypress/integration/Chip/Accepts_children.feature
```diff
@@ -0,0 +1,5 @@
+Feature: The Chip renders children
```
HendrikThePendric: For the chip it's important to test the `onRemove` prop. I.e. if that prop is supplied a "remove icon" is rendered and `onRemove` should be called when that icon is clicked.

cypress/integration/MenuItem/Is_clickable.feature
```diff
@@ -1,6 +1,6 @@
 Feature: The MenuItem has an onClick api
 
     Scenario: The user clicks on the MenuItem
-        Given a MenuItem with onClick handler is rendered
+        Given a MenuItem with onClick handler and value is rendered
         When the MenuItem is clicked
-        Then the onClick handler is called
+        Then the onClick handler is called with value
```
HendrikThePendric: IMO, the most complex part of the Menu is it's ability to include submenus that pop-out. This isn't included yet. And I believe it's currently broken functionality too. So I guess you can ignore this, and I will add tests for that when I work on the Menu (make it powered by the Popper)

cypress/integration/Tab/Is_clickable/index.js
```diff
@@ -19,3 +29,9 @@ Then('the onClick handler is called', () => {
         expect(win.onClick).to.be.calledWith({})
     })
 })
+
+Then('the onClick handler is not called', () => {
+    cy.window().then(win => {
+        expect(win.onClick).not.to.be.called
+    })
+})
```
HendrikThePendric: Perhaps add something to assert the `selected` prop. The selected tab needs to have a big blue underline.

cypress/integration/TabBar/Accepts_children/index.js
```diff
@@ -0,0 +1,17 @@
...
+Given('a scrollable TabBar with children is rendered', () => {
+    cy.visitStory('TabBar', 'Scrollable with children')
+    cy.get('[data-test="dhis2-uicore-tabbar"]').should('be.visible')
+})
+
+Then('the children are visible', () => {
+    cy.get('[data-test="dhis2-uicore-tabbar"]')
+        .contains('I am a child')
+        .should('be.visible')
+})
```
HendrikThePendric: We should definitely assert that a `scrollable`  `TabBar` is in fact scrollable.

ismay:
> In general, it would be good to use the new approach implemented here: #773

Definitely. I will rebase on that once it's merged, and update the tests to the new approach.

cypress/integration/AlertBar/Accepts_icon.feature
ismay: Yeah that's what I tried to test here. So this: `AlertBar icon shows by default` tests the standard icon, this one ` Scenario: AlertBar icon can be hidden` tests whether it can be disabled with `false` and this one: `Scenario: Custom AlertBar icon` tests whether a custom element will be rendered.

ismay: > Actually, looking at the AlertBar tests in general, I think it'd be good to add a section about statuses. Each status has a very different appearance...

Yeah, I would also like to test that, but I think visual regression tests would be most appropriate for that. Since that's in my opinion essentially what we want to confirm.

cypress/integration/AlertStack/Shows_alertbars.feature
ismay: I think we could probably also assert the order of the bars via their contents, seems more reliable. I'll see what I can do.

cypress/integration/ButtonStrip/Accepts_children.feature
ismay: Yeah here I've also avoided testing the styles (see earlier comment). Would be nice to test things like that though, for sure.

cypress/integration/Chip/Accepts_children.feature
ismay: Definitely. That test already existed though: https://github.com/dhis2/ui-core/blob/3a8db74bc13f30cbe1b5f58e02c494de5cc363ea/cypress/integration/Chip/Is_removable.feature. Let me know if you think it's sufficient.

cypress/integration/MenuItem/Is_clickable.feature
ismay: Yeah. I tried testing this, but you're right, it seems to be broken currently.

cypress/integration/AlertBar/Accepts_icon.feature
HendrikThePendric: Aha gotcha! I think the same will apply to a few of my other comments too then

cypress/integration/Tab/Is_clickable/index.js
ismay: Agreed, though also here I'd prefer visual tests.

cypress/integration/Chip/Accepts_children.feature
HendrikThePendric: Aha, I didn't realise that. I'll mark this as resolved.

cypress/integration/TabBar/Accepts_children/index.js
ismay: Yeah you're right. That's one I've put on the backburner. I can create a list of all the things I haven't yet addressed, which can be done in a follow up PR. Create a Jira issue for it for example.

cypress/integration/Tab/Is_clickable/index.js
HendrikThePendric: This is again style related, so not part of the PR scope. I'll resolve this one.

cypress/integration/MenuItem/Is_clickable.feature
HendrikThePendric: I'll resolve this one too. I'll deal with this later

cypress/integration/AlertStack/Shows_alertbars.feature
HendrikThePendric: As discussed this will also be done via a visual test.

cypress/integration/TabBar/Accepts_children/index.js
HendrikThePendric: OK

PR #792, opened 2020-02-10
ci: port to cli-utils-cypress

varl:


Mohammer5:
It's really nice to see this in effect!

I'm wondering if we should move the `concurrently`/`wait-for` setup to the cypress repo as well, because that's the really annoying process when setting up this stuff in new repositories. Or is that added by the cli helper scripts?

varl:
BTW there is some rudimentary docs here: https://cli-utils-cypress.dhis2.nu/#/getting-started

Main repo is here: https://github.com/dhis2/cli-utils-cypress/

They will need to be filled out as we go.

varl:
> It's really nice to see this in effect!

:+1: 
 
> I'm wondering if we should move the `concurrently`/`wait-for` setup to the cypress repo as well, because that's the really annoying process when setting up this stuff in new repositories. Or is that added by the cli helper scripts?

Yeah, it is annoying. Will need to think about how that is going to work, maybe could wrap it in a single command like:

```
d2 utils cypress run --wait-for "yarn start:testing"
```

And wire up concurrently and wait-for internally.

Issue #500, opened 2019-10-11
Menu: update/extend from specs


cooper-joe:
I've updated the `menu` component with some extended functionality and cleaned up visuals. 
[The updated documentation](https://github.com/dhis2/design-system/blob/master/molecules/menu.md) (design spec linked at top)

The required changes, summarised:

- [ ] set up min/max width/height, detailed in spec notes
- [ ] adjust font size and padding for dense
- [ ] dense divider
- [ ] section headers (regular/dense)
- [ ] destructive type menu item
- [ ] disabled state menu item
- [ ] icon size standardization (regular/dense)

The design spec is much more detailed now, (the last one came straight out of MUI), the menu will now be more polished and extendable with the new features here 👍

PR #803, opened 2020-02-18
fix(menu): style adjustments, dense divider, multiline items

cooper-joe:
Fixes some of the issues from https://github.com/dhis2/ui-core/issues/500, I've not included things that could be considered features.

Fixed/Added to menu and/or menu-item
- min-width
- max-width, can be customised
- dense divider style
- destructive menu-item style
- disabled menu-item style
- allow multi-line menu items
- removed menu-item heights, now set via padding on content
- adjusted dense menu-item padding

I'm not sure if any of these would classify as being added as a `feat` instead, let me know if so.

Note: I've added and removed some props, I think I've done this in the correct way by documenting them with ` * @prop {boolean} [destructive]` etc. but an extra set of eyes here is much appreciated.

### DIFF
--- a/src/Divider.js
+++ b/src/Divider.js
@@ -10,7 +10,7 @@ import { colors, spacers } from './theme.js'
  * @returns {React.Component}
  * @example import { Divider } from @dhis2/ui-core
  */
-const Divider = ({ margin, className, dataTest }) => (
+const Divider = ({ className, dataTest, dense, margin }) => (
     <div className={className} data-test={dataTest}>
         <style jsx>{`
             div {
@@ -22,27 +22,29 @@ const Divider = ({ margin, className, dataTest }) => (
         `}</style>
         <style jsx>{`
             div {
-                margin: ${margin};
+                margin: ${dense ? `${spacers.dp4} 0` : margin};
             }
         `}</style>
     </div>
 )
 
 Divider.defaultProps = {
-    margin: `${spacers.dp8} 0`,
     dataTest: 'dhis2-uicore-divider',
+    margin: `${spacers.dp8} 0`,
 }
 
 /**
  * @typedef {Object} PropTypes
  * @static
  * @prop {string} [className]
- * @prop {string} [margin="${spacers.dp8} 0"] - A CSS shorthand declaration for margin
  * @prop {string} [dataTest]
+ * @prop {bool} [dense]
+ * @prop {string} [margin="${spacers.dp8} 0"] - DEPRECATED: A CSS shorthand declaration for margin. If margin and dense are used at the same time, dense has precedence.
  */
 Divider.propTypes = {
     className: propTypes.string,
     dataTest: propTypes.string,
+    dense: propTypes.bool,
     margin: propTypes.string,
 }
 
--- a/src/Menu.js
+++ b/src/Menu.js
@@ -16,7 +16,7 @@ import { spacers } from './theme.js'
  * @see Specification: {@link https://github.com/dhis2/design-system/blob/master/molecules/menu.md|Design system}
  * @see Live demo: {@link /demo/?path=/story/menu--default|Storybook}
  */
-const Menu = ({ children, className, dataTest }) => (
+const Menu = ({ children, className, dataTest, maxWidth }) => (
     <div className={className} data-test={dataTest}>
         <Card>
             <div className="menu-list-wrapper">
@@ -28,12 +28,22 @@ const Menu = ({ children, className, dataTest }) => (
             .menu-list-wrapper {
                 padding: ${spacers.dp4} 0;
             }
+            div {
+                display: inline-block;
+                min-width: 128px;
+            }
+        `}</style>
+        <style jsx>{`
+            div {
+                max-width: ${maxWidth};
+            }
         `}</style>
     </div>
 )
 
 Menu.defaultProps = {
     dataTest: 'dhis2-uicore-menu',
+    maxWidth: '380px',
 }
 
 /**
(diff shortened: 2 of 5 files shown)


src/Menu.js
```diff
@@ -16,7 +16,7 @@ import { spacers } from './theme.js'
  * @see Specification: {@link https://github.com/dhis2/design-system/blob/master/molecules/menu.md|Design system}
  * @see Live demo: {@link /demo/?path=/story/menu--default|Storybook}
  */
-const Menu = ({ children, className, dataTest }) => (
+const Menu = ({ children, className, dataTest, width }) => (
```
varl: I would prefer `maxWidth` here as that reflects the result more accurately.

src/Menu.js
```diff
@@ -28,12 +28,18 @@ const Menu = ({ children, className, dataTest }) => (
             .menu-list-wrapper {
                 padding: ${spacers.dp4} 0;
             }
+            div {
+                display: inline-block;
+                min-width: 128px;
+                max-width: ${width};
```
varl: This line should be in its own `<styled jsx>` tag because of how styled-jsx handles static vs. dynamic CSS[1]:

```
        <style jsx>{`
            .menu-list-wrapper {
                padding: ${spacers.dp4} 0;
            }
            div {
                display: inline-block;
                min-width: 128px;
            }
        `}</style>
        <style jsx>{`
            div {
                max-width: ${width};
            }
        `}</style>
```

[1] https://github.com/zeit/styled-jsx#via-interpolated-dynamic-props

src/MenuItem/styles.js
```diff
@@ -60,13 +74,8 @@ export default css`
     .label {
         color: ${colors.grey900};
         font-size: 15px;
-
-        white-space: nowrap;
-        overflow: hidden;
-        text-overflow: ellipsis;
-
+        line-height: 1.2;
```
varl: Could we use a `px` unit for the `line-height`?

src/Divider.js
```diff
@@ -10,7 +10,7 @@ import { colors, spacers } from './theme.js'
  * @returns {React.Component}
  * @example import { Divider } from @dhis2/ui-core
  */
-const Divider = ({ margin, className, dataTest }) => (
```
varl: Removing the margin prop is technically a breaking change. Could we simply deprecate it for now and remove it in the next major that is coming up?

E.g.:

```
 * @prop {string} [margin="${spacers.dp8} 0"] - DEPRECATED: A CSS shorthand declaration for margin
```

cooper-joe: Yes, but is there a way I can bypass the `no-unused-vars` lint error? I was thinking that if the prop is deprecated I could simply not use it, or am I mistaken?

In theory, I could keep the default value for the `margin` prop, but override it with the new `dense` prop. E.g. on line 25 of `Divider.js`:
`div {margin: ${dense ? spacers.dp4 : margin} 0;`?

src/Menu.js
cooper-joe: Oh, that makes sense. Thanks for the reference.

src/Divider.js
varl: Not using it would also be a breaking change, so that option is a no-go. We need to support margin _and_ dense in a transition period.

The problem with this `margin: ${dense ? spacers.dp4 : margin} 0;` is that the `margin` prop accepts a short-hand CSS declaration string, i.e. this is valid for `margin`: `margin="10px 5px 0px 15px"` but that would be wrong in the example you gave.

It would need to be more along the lines of:

```
margin: ${ dense ? `${ spacers.dp4 } 0` : margin };
```

This means that `dense` and `margin` are mutually exclusive though, so that propType validation needs to be added too.

cooper-joe: Ah yes, I understand. The only issue I see now is that, looking at [@dhis2/prop-types](https://github.com/dhis2/prop-types#mutuallyexclusiveexlusivepropnames-proptype--error--null), it is not possible to set multiple propTypes.

dense is `bool` and margin is `string`. It doesn't seem to be possible to do:
```
propTypes.mutuallyExclusive(
    ['margin', 'dense'],
    [dense: propTypes.bool, margin: propTypes.string]
)
```
This is not a critical change, so if this causes issues then it can be left out until the more thorough rework on the menu.

varl: Haha, I did not think about that. Yeah, that probably won't fly.

OK -- then we can forgo the `mutuallyExclusive` propType validation and note it in the jsdoc string:

`* @prop {string} [margin="${spacers.dp8} 0"] - DEPRECATED: A CSS shorthand declaration for margin. If dense is used, it has precedence.`

cooper-joe: Cool 👍 I've made those changes.

HendrikThePendric: I actually think you could get this to work in the current form of `mutuallyExclusive`:
```javascript
const denseNotWithMargin = propTypes.mutuallyExclusive(
    ['margin', 'dense'],
    propTypes.string
)

const marginNotWithDense = propTypes.mutuallyExclusive(
    ['margin', 'dense'],
    propTypes.bool
)
```

I fully agree this is awkward 😬 
But it might be able to help you our here....

src/Menu.js
```diff
@@ -16,7 +16,7 @@ import { spacers } from './theme.js'
  * @see Specification: {@link https://github.com/dhis2/design-system/blob/master/molecules/menu.md|Design system}
  * @see Live demo: {@link /demo/?path=/story/menu--default|Storybook}
  */
-const Menu = ({ children, className, dataTest }) => (
+const Menu = ({ children, className, dataTest, maxWidth }) => (
```
varl: Adding the `maxWidth` is a new feature (`feat(menu):`)

src/MenuItem.js
```diff
@@ -64,6 +64,7 @@ const MenuItem = ({
     icon,
     children,
     active,
+    destructive,
```
varl: `destructive` is also a new feature of `feat(menuitem)`.

src/Divider.js
```diff
@@ -10,7 +10,7 @@ import { colors, spacers } from './theme.js'
...
+const Divider = ({ className, dataTest, dense, margin }) => (
```
varl: `dense` is a new feature of `feat(divider)`.

cooper-joe:
Thanks @varl, I've rebased and reworded those commits and squashed the others.

PR #809, opened 2020-02-19
feat: add Tag component

cooper-joe:
Adds Tag atom, documented here: [Design Spec](https://github.com/dhis2/design-system/blob/master/atoms/tag.md).

There is one issue I need assistance with: the 'With Clipped Long Text' example. The text should truncate with an ellipsis. For some reason this doesn't happen here, even though I have added all the css it _should_ need. 

This is my first component PR, so I would appreciate some extra eyes from @dhis2/team-apps  on this to make sure I've included everything correctly 😄, thanks!


varl:
Good work @cooper-joe, it's awesome that you are starting to produce components!

I know we talked about e2e tests AFK before, and now I've thought about it some more and there are two behaviors we could define in `.feature` files, and subsequently test:

1. Clipping long text
2. The icon variation

src/__demo__/Tag.stories.js
```diff
@@ -0,0 +1,56 @@
...
+        viewBox="0 0 24 24"
+        width="24"
+        xmlns="http://www.w3.org/2000/svg"
+    >
+        <path
+            d="m12 2c-5.52 0-10 4.48-10 10s4.48 10 10 10 10-4.48 10-10-4.48-10-10-10zm0 14.5c-2.49 0-4.5-2.01-4.5-4.5s2.01-4.5 4.5-4.5 4.5 2.01 4.5 4.5-2.01 4.5-4.5 4.5zm0-5.5c-.55 0-1 .45-1 1s.45 1 1 1 1-.45 1-1-.45-1-1-1z"
+            fill="inherit"
+        />
+    </svg>
+)
```
HendrikThePendric: When I looked at this, I didn't initially recognise this as a story, because I not familiar with this way of writing it. There is absolutely nothing wrong with this from a technical perspective, but I was going to request you to change this to the syntax we currently use all over the place, which is using `storiesOf` and `add`, so we keep things consistent.

However, I've just read that the way you've done things is the [recommended way since v5.2](https://storybook.js.org/docs/formats/component-story-format/). So, I've changed my mind and actually think that we should probably all start writing stories this way.

However, for this pattern I think it would be nice if we move all the exports to the bottom of the file, so it is easy to see the structure of the story in one go, a bit like this:
```javascript
export default { title: 'Tag', component: Tag }

export {
    Default,
    WithIcon,
    Neutral,
    Positive,
    Negative,
    Bold,
    WithClippedOversizedIcon,
    WithClippedLongText,
}
```

src/Tag.js
```diff
@@ -0,0 +1,147 @@
...
+                    fill: ${colors.green800};
+                    color: ${colors.green900};
+                }
+
+                .bold {
+                    font-weight: 700;
+                    background-color: ${colors.grey700};
+                    color: ${colors.white};
+                    fill: ${colors.white};
+                }
```
HendrikThePendric: I'm not 100% sold on the `bold` prop-name, because this is not only influencing the font-weight, but also the colours. This might be unexpected behaviour. Should we perhaps use a different name here, like `emphasized` which is more annoying to type but has a slightly more generic meaning than bold?

HendrikThePendric:
> There is one issue I need assistance with: the 'With Clipped Long Text' example. The text should truncate with an ellipsis. For some reason this doesn't happen here, even though I have added all the css it should need.

The problem is `display: inline-flex;` which doesn't support text-ellipsis. `display: inline-block;` would fix it, but I've understood `inline-flex` would be preferred. In this case I'd just add an extra span into the div to take care of the text ellipsis.

src/__demo__/Tag.stories.js
varl: I started a full rewrite of the stories in ui-core to the SCF-format, which I didn't finish because I was doing it at the same time @ismay touched basically all stories to fix another issue. The merge conflict was hellish, so I dropped the work.

I did do around 50 stories and the way @cooper-joe has written these stories is the way I settled on as well, so I recommended him to write stories this way. I tried the structure you lined out as well, but I did not like it in the end.

This structure [recommended by Storybook](https://storybook.js.org/docs/formats/component-story-format/) works very well, so since that is the documented approach I don't see a reason to go against that.

```
// << meta data >>
export default {...}

// << utilities >>
function () {...}

// << stories >>
export const Story = () => {...}
```

It works well for SCF since it allows all the metadata at the top when you open the story. This is usually simple (`title` and `tag`) but all sorts of Storybook extensions go there, so it is important enough to have at the top.

The `export const Story` scheme vs. `export { Story }` is a toss-up for me, but the former leads to less typing and conforms to the way the SCF format is defined, so I don't see overwhelming reason to not use `export const Story`.

src/common-prop-types.js
```diff
@@ -1,5 +1,10 @@
 import propTypes from '@dhis2/prop-types'
 
+export const tagVariantPropType = propTypes.mutuallyExclusive(
+    ['neutral', 'positive', 'negTive'],
+    propTypes.bool
+)
```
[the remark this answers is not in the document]
cooper-joe: I'm not sure these prop-types will be reused, I can't say for definite, so I've moved them to the file instead.

src/Tag.js
cooper-joe: My thinking is that `bold` is also a descriptor of the colour too. The color is bold, where the regular tag colours are muted/pastel. `Bold` as a prop-name is intended to refer to the usage (use this when you need to make a bold statement), rather than a descriptor of appearance. I'm not sure if this is frowned upon for prop names in general?

src/__demo__/Tag.stories.js
HendrikThePendric: OK, thanks for explaining this @varl. Makes sense. So I guess you can just leave the story that way @cooper-joe :-)

src/Tag.js
HendrikThePendric: Interesting... When doing web development and reading `bold` I immediately think of _a bold font_ (thicker characters) probably because of `font-weight: bold;`. But you are right, the actual definition is much broader: _having a strong, vivid, or clear appearance._ 

It is also an easy and short word to type, which is always good.

And thinking about it some more, my suggestion `emphasized` suffers from a similar problem, due to the `<em/>` tag.

Taken the above into account, I think that `bold` is probably fine.

ismay:
I've added integration tests for the parts of the Tag that were testable. Since the truncation of the text is done with CSS I'm not able to test for it in cypress (this would be a good usecase for screenshot tests).

I've also added the public export of the Tag, and moved the two sub Components to a `Tag` folder, like we've done with other components as well.

PR #814, opened 2020-02-25
chore(storybook jsx addon): add missing decorator to show jsx again

Mohammer5:
This will add the jsx-addon decorator which is missing since we upgraded to the newer setup-style.


varl:
Thanks, @Mohammer5. I think I removed it by accident.

varl:
Thanks, @Mohammer5. I think I removed it by accident.

PR #820, opened 2020-03-02
feat: introduce popover, expose popper, refactor split/dropdown-button

HendrikThePendric:
~This PR is still in draft status. There are still quite a few things to do, see this list:~ I've removed the draft status from this PR because I've addressed all the points below.
- [x] Remove `DropMenu`
- [x] Power `DropdownButton` by `Popper`
- [x] Power `SplitButton` by `Popper`
- [x] Introduce `Popover`
- [x] Tweak `Popover` feature-files so it doesn't do re-sizing but shifts into viewport instead
- [x] Add e2e tests and e2e-stories 
- [x] Add demo stories
- [x] Fix build
- [x] Add docs to `Popper` and `Popover`
- [x] Export `Popover` and `Popper` from `index.js`

However, looking at this PR now does make sense, because we see a pattern emerging:
- The current implementation of the `Popper` is very simple. We are essentially fully delegating things to popper.js and ask React to take the back-seat.
- If we need more functionality, this means adding modifiers, i.e. fixing things the popper.js way...

I've studied the [implementation in MUI](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Popper/Popper.js). This is a bit more complex than our approach but the principle is the same. But I've also looked and [`react-popper`](https://github.com/popperjs/react-popper/blob/master/src/Popper.js) (made by the same org as popper.js) and this approach is quite different. Here we see that `setState` is being called from within a component's class method, and this class method is also a modifier on the popper.js instance. This way seems to introduce a lot more complexities, but at least React "is more in control".

My guess is that the second approach is probably more flexible in the long run, but I can't really think of any scenarios that wouldn't be covered by our current simple approach. **I'd love to get some input on that**...

---

Another "pattern" that we see emerging is the repetition of the `createPortal` + `Backdrop` + `Popper` to make "poppable" components. 

I believe @cooper-joe is facing a similar problem in his design work. There are quite a few things that have a lot of aspects in common with the `Popover`, but the `Popover` is a bit too opinionated to use...

Joe and I were discussing this issue and we considered three courses of action:
- Keep the repetition (status quo)
- Introduce a component that is "in between the `Popper` and the `Popover`". We found it hard to come up with something with a unique descriptive name that could be exposed. But perhaps we could introduce an (internal) component and call it `BasePopover` and use that?
- Change the design specs for the `Popover`, so it:
    - Refrains from setting styles to the content
    - Makes the arrow optional

**What do you guys think?**


varl:
> Joe and I were discussing this issue and we considered three courses of action:
> 
>     * Keep the repetition (status quo)
> 
>     * Introduce a component that is "in between the `Popper` and the `Popover`". We found it hard to come up with something with a unique descriptive name that could be exposed. But perhaps we could introduce an (internal) component and call it `BasePopover` and use that?
> 
>     * Change the design specs for the `Popover`, so it:
>       
>       * Refrains from setting styles to the content
>       * Makes the arrow optional

Did you land on a conclusion?

Mohammer5:
Ah, I wanted to imply that I like the status quo, I don't really mind the repetition. Looks quite compos-able to me?

cooper-joe:
> > Joe and I were discussing this issue and we considered three courses of action:
> > ```
> > * Keep the repetition (status quo)
> > 
> > * Introduce a component that is "in between the `Popper` and the `Popover`". We found it hard to come up with something with a unique descriptive name that could be exposed. But perhaps we could introduce an (internal) component and call it `BasePopover` and use that?
> > 
> > * Change the design specs for the `Popover`, so it:
> >   
> >   * Refrains from setting styles to the content
> >   * Makes the arrow optional
> > ```
> 
> Did you land on a conclusion?

I don't have an opinion about repeating the composition code. But conversations that @varl and I have had lately led me to conclude that the best option is the third option, slightly modified:
- Do not set styles on the content of the `Popover`
- Make the arrow optional
- Make the [elevation](https://github.com/dhis2/design-system/blob/master/atoms/elevation.md) a prop that can be changed.
- allow `position` to be set as a prop (based on positions available in our positioning engine.

This component would then cover the majority of the use cases for all 'popping' elements. Guidelines for content padding would be set per-component in the design system, not enforced in the generic component.

HendrikThePendric:
@cooper-joe I have implemented your suggestion, but need to check a few things:
> Do not set styles on the content of the Popover

I've removed `padding`, `color`, `font-size`, and `line-height`. And I've kept the `background-color` and `border-radius` fixed. For `box-shadow` I am using the `elevation` prop as per your suggestion. I thought that the same would also make sense for maxWidth, so that's now controlled by a prop too (similar as Tooltip)... Makes sense?

> Make the arrow optional

I've done this now, there is an `arrow` prop with default value `true`. One thing you need to confirm: When removing the arrow, I also removed the `offset`, so the popover shows right next to the anchor element. Is that OK?

I'll also add an extra e2e test for the no-arrow option. Will see what your reply is on Monday and finish things up then.

cooper-joe:
> I've removed `padding`, `color`, `font-size`, and `line-height`. And I've kept the `background-color` and `border-radius` fixed. For `box-shadow` I am using the `elevation` prop as per your suggestion. I thought that the same would also make sense for `maxWidth`, so that's now controlled by a prop too (similar as Tooltip)... Makes sense?

Yeah, that makes sense 👍 

> > Make the arrow optional
> 
> I've done this now, there is an `arrow` prop with default value `true`. One thing you need to confirm: When removing the arrow, I also removed the `offset`, so the popover shows right next to the anchor element. Is that OK?

Yes, perfect! 😄

Issue #794, opened 2020-02-11
Where to put default translated texts?


HendrikThePendric:
## Intro
We decided a long time ago that `ui-core` should not use `@dhis2/i18n`. Everyone was on board with that decision but it turns out the situation that landed us in is far from ideal. We either need to find a better solution for adding translated default texts, or reconsider putting i18n into `ui-core`.

## Current approach
Currently we have identified three components in `ui-core` that need a quite a few translated default/placeholder texts:
- `SingleSelect`
- `MultiSelect`
- `FileInputFieldWithList`

The approach we tried was to export counter-parts for these components from `ui-widgets` that do contain translations. See https://github.com/dhis2/ui-widgets/pull/309 for more info.

### Problems
As I see it, the main problem with this approach is that it makes the distinction between `ui-widgets` and `ui-core` very blurry. My current metal model is like this:
> ui-core has component-primitives and ui-widgets has components that interact with the dhis2-core webapi, or contain business logic.

That distinction has proved to be fairly hard to grasp, but with the `i18n` we'll need to add another criteria to that distinction:
> [...] or contain translations

A new developer in the ecosystem can't really be expected to know he/she should work like this when wanting to use a `MultiSelect` with the default translations:
```javascript
import { MultiSelectField } from '@dhis2/ui-widgets'
import { MultiSelectOption } from '@dhis2/ui-core'
```

There are some other factors that make the current approach problematic, but I think we should only focus on the one above, because this relates directly to our "end-product".

## Alternative approaches
I've already started discussing this on Slack with @ismay and @Mohammer5 and some alternatives were suggested:

### A library with translated components
Probably this would just re-export all ui-core components but just add translated defaultProps. Personally I wouldn't like this.

### A translation library / a translations helper in an existing library
There are many ways we could implement something like this:
1. Just exporting the translated text as a const and apply it to each individual prop (explicit, but not very convenient)
2. A HoC (might be interesting, but will create an additional component-layer)
3. A helper that just appends the relevant defaultProps to the components that need it (could be convenient, but probably quite quirky)

### Adding `@dhis2/i18n` to `ui-core`
Yes.... I know.... We weren't going to do that... But the other solutions all seem a bit off as well....

amcgee:
I think I'd be in favor of adding translations to `ui-core` for consistency and to keep it self-contained (otherwise you basically duplicate the "helper text" and all its translations everywhere it's used).

[...]

PR #822, opened 2020-03-04
refactor(monorepo): convert ui-core to ui monorepo

varl:
This is based on the proposal in https://github.com/varl/ui-merge-proposal, but it does the bare minimum as a first pass.

Right now this only includes two packages:

- `@dhis2/ui`
- `@dhis2/ui-core`

Required next steps:

- [x] Fix Jest
- [x] Fix Cypress
- [ ] Fix docs
- [x] Pull in `@dhis2/ui-widgets`
- [ ] Pull in `@dhis2/ui-forms`
- [ ] Move molecules out of core to widgets (enables translations, etc.)
- [ ] Move dependencies out from core to widgets (enables dependencies)

Optional steps:

- [ ] Reorganize structure of stories
- [ ] Rewrite stories in SCF

# Caveat

- Yarn 1.19.2 has a [bug in the Workspace feature](https://github.com/yarnpkg/yarn/issues/7807) causing `yarn install` to fail, so it is recommended to run `yarn policies set-version 1.19.1`.

- The 300+ commits is not a mistake, it is done that way to preserve the entire Git history for the merged repository (e.g. all of the history for ui-widgets is preserved when merging it into ui-core).
  _Addedum: I squashed the history for now as it makes it easier to rebase on master_


Mohammer5:
Copied from slack:

> and it's worse because an organism is also a molecule
> it's just a large molecule
> (by @varl)

I def. do not agree with that.. I just like a molecule is not an atom. I prefer to think in terms of chemistry/biology. Then this whole differentiation makes a lot of sense.

<hr />

But I also don't see the benefit of calling it ui-core over ui-atoms?

> this is final, lthe benefit is not breaking the world by renaming ui-core to something else again
> (response by @varl)

ismay:
The discussion we had about this on slack is here: https://dhis2.slack.com/archives/CBM8LNEQM/p1583309236163900

I mentioned the following:

> @varl To me it feels like the core of the issue was unclarity about what to put where in our ecosystem. I.e., where do components with translations go, where does this go, where does that go. I think in the original issue, @Jan-Gerke Salomon's proposal seemed like a good solution to me because it attempted to clarify the distinction between our libs.With the monorepo I don't really see how this would help us distinguish better between our different libs and figure out what to put where. Maybe I'm overlooking something, let me know if I am.
>
> Not saying that JG's proposal would be the only or exact way I'd advocate, but I did like the attempt at setting up clear boundaries between the different libs we have. That felt like the main problem to me.

So the reason I mentioned that I feel like we have a lot of discussions about what to put where was based on our discussing these things in the following issues:

- https://github.com/dhis2/ui-core/issues/794
- https://github.com/dhis2/notes/issues/60

There is quite a bit of discussion on component organisation there. @varl mentioned:

> - atoms => core
> - molecules/organisms => widgets

And that is a clear definition. Yet somehow we still end up discussing these things a lot. So my question is, with this definition why do we still discuss where to put our components so much? And would this PR resolve that?

varl:
#### Problem 1: What goes where (core vs. widgets) 

>Yet somehow we still end up discussing these things a lot. So my question is, with this definition why do we still discuss where to put our components so much? And would this PR resolve that?

I do not see a lot of discussions about what an Atom is, everyone seems to have an OK grasp about what an atom is, so that is not a problem we are having.

Molecules and Organisms have the definitions:

> [...] **Molecules** are relatively simple groups of UI elements functioning together as a unit.
> [...] **Organisms** are relatively complex UI components composed of groups of molecules and/or atoms and/or other organisms.

_Relatively simple_ vs. _relatively complex_ is hard to define. Right now we have "relatively simple" components in ui-core and "relatively complex" components in ui-widgets, and this creates a lot of back and forth.

The idea is to move them all to widgets and removing the "relatively" from the mix and make it clear that if it's not an atom, it's a widget.

#### Problem 2: Components that depend on data from DHIS2 API

The Header Bar (at least) has a prop-driven version and a connected version. I think @Mohammer5's suggestion about having a separate DHIS2-layer lib (let's call it `ui-data` for now) that has the connected variant is interesting. It does open up for discussion about "what goes where" so I'd prefer that it just goes in widgets.

#### Problem 3: Dependency management

I still maintain that it is imperative for the long-term health of the UI system to have zero dependencies for the core components.

I realize that we have a real need to be able to use dependencies, so the compromise is to allow dependencies in widgets (i18n, popper, etc.) and ban then in core.

#### Problem 4: Developer convenience

We can move components from core to widgets and maintain a proper split between all the UI libraries, and the drawback is when developing a molecule that needs new atoms in core, then they first have be merged to core or the developer has to link the libraries locally. Matters are made worse if the dev is working on a form component that needs something in widgets and that triggers work in core.

In a monorepo we can have all the code organized properly in a single PR, and the developer is not "held back" by either local development tricks like links nor waiting for a PR.

HendrikThePendric:
I personally think heading this direction makes sense, and fully agree with the motivation given by @varl.

One note: When pulling in ui-forms, the components need to be renamed. I believe we agreed on the `Control` suffix. So that'd mean `Input` should be renamed to `InputControl` etc.

varl:
Still work to be done, but I've reorganized the stories as an illustration: https://deploy-preview-822--dhis2-ui-core.netlify.com/demo/?path=/story/component-core-card--default

varl:
OK, this way of merging the repos is hard to maintain when it's WIP. Any change in ui-core is nigh impossible to rebase.

I found a different strategy with subtree, that allows us to pull in changes in ui-core and ui-widgets into a new tree and rebase new changes on that.

Superseded by: https://github.com/dhis2/ui/pull/2

PR #823, opened 2020-03-05
fix: rewrite transfer feature files from user perspective, add missing details

cooper-joe:
adjusted feature files to be written from the context, rather than the implications.

also took the opportunity to add some more detail to features that were lacking it.

### DIFF
--- a/cypress/integration/Transfer/drag-and-drop-items.feature
+++ b/cypress/integration/Transfer/drag-and-drop-items.feature
@@ -4,22 +4,36 @@ Feature: Drag and drop selected items
     Given reordering is enabled
 
   # I have excluded any design information from this scenario: opacity, how the items look when dragging etc.
-  Scenario: Drag and drop a single selected item
+  Scenario: The user clicks, drags and drops a single item
     Given the selected list has two or more items
-    When an item is clicked and dragged
+    When the user clicks and drags a single item
     Then the item is removed from its original position in the list
     And the item is placed in the list at the cursor position when the click is released
 
-  Scenario: Drag and drop two or more selected items
+  Scenario: The user clicks, drags and drops multiple items
     Given the selected list has three or more items
     And two or more highlighted items
-    When one of the highlighted items is clicked and dragged
+    When the user clicks and drags one of the highlighted items to a new position
     Then the items are removed from their original positions
     And the items are placed in the list at the cursor position when the click is released
     And the items are ordered based on their previous ordering
 
-  Scenario: Disable reordering with a single item
+  Scenario: The user clicks, drags and drops a single item in a list of one
     Given the selected list has one item
     And one highlighted item
-    When the highlighted item is clicked and dragged
+    When the user clicks, drags and drops the highlighted item
     Then the highlighted item remains in position and does not move
+
+
+  # an attempt an outline of how multiple items reorder on drag and drop
+  Scenario Outline: The user clicks and drags multiple items
+    Given the selected list has items: <items>
+    And the <highlight> items are highlighted
+    When the user clicks and drags on one of the highlighted items
+    And releases the click after the <drop> position of the list
+    Then the new order of the items is <newOrder>
+
+    Examples:
+    | items               | highlight | drop   | newOrder               |
+    | 1, 2, 3 , 4, 5      | 2, 4,     | fourth | 1, 3, 2, 4 , 5         |
+    | 1, 2, 3, 4, 5, 6, 7 | 1, 8      | third  | 2, 3, 1, 8, 4, 5, 6, 7 |
--- a/cypress/integration/Transfer/highlight-items.feature
+++ b/cypress/integration/Transfer/highlight-items.feature
@@ -1,20 +1,43 @@
 Feature: Highlight item in lists
 
-  Scenario Outline: Highlight a single item on mouse click
+  Scenario Outline: The user clicks an item that is not already highlighted
     Given the <type> list has one or more items
-    When an item is clicked
+    When the user clicks an item in the list
     Then it is highlighted
+    And any other highlighted items are no longer highlighted
 
     Examples:
         | type      |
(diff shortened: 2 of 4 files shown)


varl:
The most important structure to me is the same principle we use in commit messages, that is, don't explain what the commit _did_, explain what it _does_, so instead of explaining what happened:

>When one of the highlighted items is **clicked and dragged**

I prefer that this explains what is happening: 

>When the user **clicks and drags** one of the highlighted items to a new position

Which also makes sense in the test case, where the statement matches what the code _does_, not what it _did_.

```
When('the user clicks and drags one of the highlighted items to a new position', () => {
  // code that clicks and drags a highlighted item from x to y
})
```

If we use "the user" as an actor or not is not that important to me, but the present/past tense form matters a lot for readability and understanding.

Mohammer5:
@varl the following sentence is present:

> When one of the highlighted items is clicked and dragged

It's passive, not past. 

These are virtually identical in meaning:

> When one of the highlighted items is clicked and dragged by the user
> When the user clicks and drags one of the highlighted items

I like having passive sentences without the acting entity as that describes pure behavior regardless of who is acting (User, accessibility software, e2e test engine, etc)

varl:
> @varl the following sentence is present:
> 
> > When one of the highlighted items is clicked and dragged
> 
> It's passive, not past.

OK, I may be using the wrong grammar terms here, that is not the heart of the point however.

> 
> These are virtually identical in meaning:
> 
> > When one of the highlighted items is clicked and dragged by the user
> > When the user clicks and drags one of the highlighted items

Agreed, there are many ways to write a sentence so that the meaning is identical. This is also besides the point.

> I like having passive sentences without the acting entity as that describes pure behavior regardless of who is acting (User, accessibility software, e2e test engine, etc)

Passive sentences for what part of Given-When-Then? Or all of them? A common trick that works pretty well in practice is to use past tense for *Given*, present tense for *When*, and future tense for *Then*.

In addition, *Given* and *Then* should be passive, but *When*, should definitely be active.

```
Given the user is logged in as an admin

When the user clicks on delete user "foo"

Then the foo user should be deleted
```

I do not agree that the *When* clause should be passive nor past tense, but I agree with you on *Given* and *Then*.

varl:
Ref. avoiding naming the actor:

The Actor (Cypress, Selenium, Accessibility software) acts on the system _on behalf_ of someone, and that someone is referred to as a generic Role). In UI, we only have a single role (right now at least) and that is a User.

So by not naming the role in the feature files, we just imply it. If in the future we have additional roles, then we will have to make it explicit anyway. To me it doesn't matter that much if we do it now or later.

What is important is to be aware that in applications that have multiple roles _we have to be explicit_ about what role the actor is acting on behalf of. E.g. in the Maintenance app features we have different behaviors depending on if the actor (Cypress) is acting as a restricted user or a super user. 

The actor is constant, the role is not, and that is why it needs to be explicitly mentioned in the feature files.

The question is if we should allow ourselves to have different ways of writing Given-When-Then statements for libraries and applications or be consistent across the board.

Mohammer5:
## Tense

I definitely agree that it makes sense to have `Given` in the past tense. They describe the current situation, so if they're using an action to get there it has to have happened in the past.

I don't think the `Then` step should be future but rather affirmative present (with `should` just like your example) and passive as we're describing a state rather than an action.

In terms of tense, it's probably the easiest to keep them all in the present (you did that in your example as well):

> Given the user is logged in as an admin

## Passive/active

I think you're right here. The `When` step should be active. I've read some articles about what's the standard practice and it seems everyone else is using active here as well (with some good arguments for it).

Another argument for why it's totally fine to include the term "user" is because the entity interacting with the environment is a user, regardless of what is is, even cypress uses the interface to do something. It should suffice to mention a role or specify a particular type of user when necessary

I think we could add a `howto` or `guide` to our notes repo, so we have something to point to and remember this?

cooper-joe:
> I think we could add a `howto` or `guide` to our notes repo, so we have something to point to and remember this?

+1 for this. It would be very useful to be able to refer to this document while writing features and while discussing them in PRs.

cypress/integration/Transfer/disabled-transfer-buttons.feature
```diff
@@ -0,0 +1,41 @@
+Feature: Disable transfer buttons when actions are not possible
+
+  Scenario: None of the available options is highlighted
+      Given the options list has items
+      When no option items are highlighted
```
Mohammer5: Should be `And` (`Given`)

cypress/integration/Transfer/disabled-transfer-buttons.feature
```diff
@@ -0,0 +1,41 @@
...
+      Then the 'move to selected list' button is disabled
+
+  Scenario: Some of the available options is highlighted
+      Given the options list has items
+      When some option items are highlighted
```
Mohammer5: Same as above

cypress/integration/Transfer/disabled-transfer-buttons.feature
```diff
@@ -0,0 +1,41 @@
...
+      Then the 'move to selected list' button is enabled
+
+  Scenario: None of the selected options is highlighted
+      Given the selected list has items
+      When no selected items are highlighted
```
Mohammer5: Same as above

cypress/integration/Transfer/disabled-transfer-buttons.feature
```diff
@@ -0,0 +1,41 @@
...
+      Then the 'move to options list' button is disabled
+
+  Scenario: Some of the selected options is highlighted
+      Given the selected list has items
+      When some selected items are highlighted
```
Mohammer5: Same as above

Mohammer5:
I think it should be fine to merge this, then I can reword the scenarios to match our newly defined rules

PR #828, opened 2020-03-13
fix: typography adjustments

cooper-joe:
Typography adjustments to many components, mainly letter spacing and some line heights.

Changes are made based on a type system, noted below. Note however that some components have been optically adjusted and may not exactly follow the system.

System:
```
caption @ 12px: 0.3px
labels @ 14px: 0.2px
labels @ 16px: 0.3px;
inputs/user-content @ 14px: 0.25px;
```

### DIFF
--- a/src/AlertBar/styles.js
+++ b/src/AlertBar/styles.js
@@ -22,6 +22,8 @@ export default css`
         max-width: 600px;
 
         font-size: 14px;
+        line-height: 17px;
+        letter-spacing: 0.2px;
 
         transform: translateY(1000px);
         transition: transform ${ANIMATION_TIME}ms cubic-bezier(0.4, 0, 0.6, 1);
--- a/src/Checkbox.js
+++ b/src/Checkbox.js
@@ -133,11 +133,13 @@ class Checkbox extends Component {
                         user-select: none;
                         color: ${colors.grey900};
                         font-size: 16px;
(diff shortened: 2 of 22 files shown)


cooper-joe:
> UI Core doesn't force developers to use Roboto, so those letter space values apply to any font the developer chooses.

@varl, Good point, I overlooked that important detail.

> Are we sure that the letter space values herein are better in those cases than the default letter space embedded in the fonts?

Nope, they are roboto specific letter spacing and could/would be inappropriate for other fonts.

> I think that using the font defaults is the most robust solution in most cases, but I cannot read any of the languages I'm concerned about so I have little in terms of evidence to back it up. I do know that we recently had a massive influx of Chinese translations for almost all our apps contributed, so it is a language that we support in DHIS2 now.

Font defaults are usually the best. I led myself down this rabbit hole by reading more about Roboto. I don't know why but the Roboto defaults are not ideal for UI work. The [official Material Design typography guidelines](https://material.io/design/typography/#type-scale) even recommend letter-spacing values. (I suppose it's true, different UI elements require different type characteristics).

In any case, I don't see a solution to this without building a typography system for every available language/script type. That is out of scope, so this PR is null and void.

HendrikThePendric:
> so this PR is null and void.

I'm closing it now

varl:
@cooper-joe We could use something other than Roboto. :wink:

PR #835, opened 2020-03-16
feat: introduce tooltip and e2e tests

HendrikThePendric:


cooper-joe:
@HendrikThePendric Sorry, I miscommunicated the spec for the `OPEN_DELAY`.

My intention was that the mouse cursor would need to be inside the trigger element for a minimum time to trigger the tooltip. This is to prevent tooltips being triggered on unintentionally elements when moving the cursor quickly across a page.

The way the `OPEN_DELAY` seems to work now is that the tooltip always opens, but with a delay on the opening.

`Delay` was the incorrect term, sorry. I should have specified better that this is a `minimum_hover_time` or similar.

HendrikThePendric:
@cooper-joe you found a bug, the show-timeout should be cancelled then the mouse leaves the anchor. I'll have a look why this isn't happening.

HendrikThePendric:
@cooper-joe https://github.com/dhis2/ui-core/pull/835/commits/0408f850d97e341f2494152b73714b9f4d87276b fixes the bug you found as well

cooper-joe:
> @cooper-joe [0408f85](https://github.com/dhis2/ui-core/commit/0408f850d97e341f2494152b73714b9f4d87276b) fixes the bug you found as well

Thanks @HendrikThePendric, I see the fix now. LGTM! I think the delay timings are good 👍 .

src/Tooltip.js
```diff
@@ -0,0 +1,174 @@
...
+        }, OPEN_DELAY)
+    }
+
+    onMouseOut = () => {
+        clearTimeout(this.openTimeout)
+
+        this.closeTimeout = setTimeout(() => {
+            this.setState({ open: false })
+        }, CLOSE_DELAY)
+    }
```
Mohammer5: @cooper-joe only talked about a delay when entering the element (open delay), not when leaving the element (which would be a "close delay"). I also don't see a reason for delaying the hiding process.

src/Tooltip.js
```diff
@@ -0,0 +1,174 @@
...
+            content,
+            dataTest,
+            display,
+            maxWidth,
+            placement,
+        } = this.props
+        const { open } = this.state
+
+        return (
+            <>
```
Mohammer5: No need for using a fragment. Both popper and the styles can be put into the `span`.

src/Tooltip.js
```diff
@@ -0,0 +1,174 @@
...
+                                {content}
+                            </div>
+                        </Popper>,
+                        document.body
+                    )}
+                {popperStyle.styles}
+                <style jsx>{`
+                    span {
+                        display: ${display};
+                    }
```
Mohammer5: I get the intention, not sure I agree with the solution.
Instead, we could expose an `element` prop. 

Span is a generic element for inline content, while div is the generic element for block content, so using a `span` with `display: block` is kind of wrong (technically it's correct though). Using an `element` prop could add more meaning though and would allow to automatically use styles defined by the app (i. e. when the app defined styles for paragraphs and the anchor is a `p` as well with `element="p"`)

Mohammer5: The massive list of allowed values for this prop would also disappear

Mohammer5: Using `element` or `component` would allow custom components as well as long as they support the events passed to the component.

cypress/integration/Tooltip/visibility_toggling.feature
```diff
@@ -0,0 +1,54 @@
...
+        Given there is not enough space above or below the anchor to fit the Tooltip
+        When the mouse cursor enters the anchor
+        Then the Tooltip is rendered on top of the anchor
+        And the Tooltip stays visible
+
+    Scenario: Hiding when Tooltip is on top of element
+        Given there is not enough space above or below the anchor to fit the Tooltip
+        When the mouse cursor enters the anchor
+        And the mouse cursor leaves the anchor
+        Then the Tooltip is not rendered
```
Mohammer5: lacks a newline at the end of the file

src/Tooltip.js
HendrikThePendric: This should be in the updated design-specs now, and @cooper-joe already commented on this: https://github.com/dhis2/ui-core/pull/835#issuecomment-599560591. Would you mind resolving this comment?

cooper-joe: We are using an open delay and a close delay, so this is correct.

Close delay is used to stop tooltip flickering with unintentional mouse movements. A user moving their mouse outside of the tooltip trigger will rarely notice the delay. The delay may prevent a user with motor issues from seeing a flashing tooltip if their mouse cursor is moving quickly in/out of the trigger boundary.

src/Tooltip.js
HendrikThePendric: I'm not too sure about this one, for the following reasons:
- Even though I realise that technically we are nesting a `div` in a `span`, which would result in invalid HTML, the JSX still looks a lot like that. So I think that using a fragment provides better semantics
- We are not nesting a portal within another element/component in any of our other components either. I think using this markup makes it very clear what is happening, the thing in the Portal is a separate element, not a child.

I think both arguments above are a valid reason to keep the fragment. The only valid reason to remove it imo would be if there was a substantial overhead associated with `React.Fragment` but I am not aware of any.

Looking forward to finding out what you think.

Mohammer5: I'm convinced! Keep it as it is

src/Tooltip.js
HendrikThePendric: I wasn't super stoked about my current solution either, but I see some limitations to both of the options you are suggesting.

**An `element` prop**
This basically only lets us differentiate between `inline` and `block`. Other flavours of display would be unsupported, because this cannot be linked to a specific element. I think I'd prefer to just keep the current `display` prop and just limit it to `inline` and `block` for now, and add allowed values when the need presents itself.

**Passing an `element` / `component`**
I considered this option, but it has a major downside:
- If we'd want to accept this, we wouldn't want to wrap the passed `element` / `component`,  but just render it directly (if we'd keep wrapping that component the situation would be identical to what we have now)
- But the popper needs a ref to its reference/anchor element. So that means that every component that is being passed to the Tooltip needs to implement a `ref`. We can't really enforce this, so it'd be up to the docs...

HendrikThePendric:
In reply to your general comments:

> We can use hooks now, so could make this a fully functional component to keep things consistent in `ui-core` (it won't reduce the amount of code I suppose).

I know we are allowed to use hooks now in ui-core. Personally, I think this component implementation is fairly clear and it's definitely in-line with most of the components we have. Is this comment part of the "required changes"? Or just a suggestion?

----

> I've started using 1 story-file for 1 feature file in the `Transfer` branch.

Yeah that makes sense. I'll work on that.

----

> I think we should have stories for all the positioning possibilities.

I disagree with that. The Popper has test for all positions.

src/Tooltip.js
Mohammer5: > If we'd want to accept this, we wouldn't want to wrap the passed element / component, but just render it directly (if we'd keep wrapping that component the situation would be identical to what we have now)

We can do it this way:

```jsx
const AComponent = ({ component: Component }) => (
  <Component>
  </Component>
)
```

> But the popper needs a ref to its reference/anchor element. So that means that every component that is being passed to the Tooltip needs to implement a ref. We can't really enforce this, so it'd be up to the docs...

That's true, which just means that a custom component would need to use `forwardRef`, which is available, don't see the issue here? We can't enforce certain props either (we've tried that and agreed that it's impossible), so we already have this "pattern" of "do it right and read the docs or it won't work"

Mohammer5:
> Is this comment part of the "required changes"? Or just a suggestion?

It's just a suggestion until we decide to make it consistent.

> Personally, I think this component implementation is fairly clear

Do you think hooks would be less clear?

<hr />

> I disagree with that. The Popper has test for all positions.

Yes, but they're not part of the docs for the users of `ui-core`. I like to see the behavior of a component before I use it (otherwise I'll have to try it out myself)

HendrikThePendric:
> Yes, but they're not part of the docs for the users of ui-core. I like to see the behavior of a component before I use it (otherwise I'll have to try it out myself)

I guess you mean we need more demo-stories? In that case I agree.

> Do you think hooks would be less clear?

TBH, I wouldn't even know off the top of my head how to write it. I'd have to look into the most correct implementation. What I do know is that we need the following:
- We'd need to use `useState`, `useEffect` and I think `useCallback` too.
- All these hooks would need to somehow interact with one another.
- When the one timeout is being created, the other one needs to be cleared... So I guess we'd have to declare variables for the timeouts outside of the `useEffect` function scope and create all these hook-functions inline before the return statement of the function

I've worked with hooks a bit more now, but I am certainly not an expert. Perhaps there is an easier way to rewrite this component with hooks.

src/Tooltip.js
HendrikThePendric: OK... I've implemented this in https://github.com/dhis2/ui-core/pull/835/commits/90eba6a331a89fe7ce0591e87d765ea5dd0147af

HendrikThePendric:
I've checked with @cooper-joe and since we expect to show tooltips on any given type of content (block-level, inline, etc), my original `display` prop is not flexible enough. As such, I've implemented things according to the recommendations that @Mohammer5 did.

It is very flexible but there are also elements I don't like...

src/__demo__/Tooltip.stories.js
```diff
@@ -0,0 +1,94 @@
...
+
+const Content = React.forwardRef(({ children, ...rest }, ref) => (
+    <em style={{ display: 'inline-block', color: 'red' }} ref={ref} {...rest}>
+        {children}
+    </em>
+))
+Content.displayName = 'Content'
+Content.propTypes = {
+    children: propTypes.node,
+}
```
HendrikThePendric: I personally think this is quite involved / not really user-friendly. But it is very flexible ....

varl: Yeah, it moves the complexity from inside the component to the consumer.

But on the other hand, if you just need to swap the `span` for a `div`, this approach does work quite well since it is just `component="div"` that is required. The main criticism I have is that it is a bit obscure as it is not immediately obvious how the Tooltip wraps children internally.

It would be nice to handle this all transparently (as MUI does), but their Tooltip implementation is as usual more complex than what we need. They use `React.forwardRef(Tooltip)` and `React.cloneElement(children)` to accomplish it:

- https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Tooltip/Tooltip.js#L178
- https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Tooltip/Tooltip.js#L505

HendrikThePendric: I've looked at the MUI implementation today and there is currently something stopping us from implementing things that way. I'll explain:
In the MUI docs it says this about the `children` prop:
`children * | element |  Tooltip reference element.⚠️ *Needs to be able to hold a ref*.`
And in their examples they use a variety of MUI components that are wrapped directly in a MUI `Tooltip`. When inspecting these components you will see that *each one* is actually wrapped in `React.forwardRef`. That is the sole reason that line 505, where the element is being cloned is working correctly.

So if we want to go that way, which we might want to do, then one of the consequences would be that all of our components need to be able to hold a ref. That would be a pretty big change. Perhaps quite a good one. But in any case very substantial.

HendrikThePendric: Thinking about this some more, I'm not sure how we'd work with these event handlers either... I don't think we'd want every component to have implemented `onMouseOver` and `onMouseOut` too, on top of being able to hold a ref... I suppose we could just use the ref and add some native event listeners to that... But it'd be better to use React's synthetic events....

Quite an interesting problem 🤔

HendrikThePendric: The most basic approach would just be to wrap the children in some element (or perhaps allow switching between `span`, `p`, etc.), and allow some form of styling to make sure that element behaves properly. Shame this approach would be very inconsistent with the rest of the codebase.

varl: Would it be easier to build the Tooltip using the `react-popper` library?

https://github.com/popperjs/react-popper

Maybe we found a reason to use react-popper over popper.

HendrikThePendric:
```jsx
    <Reference>
      {({ ref }) => (
        <button type="button" ref={ref}>
          Reference element
        </button>
      )}
    </Reference>
```
I don't think we should switch to `react-popper` (motivation available upon request), but perhaps we could use this pattern as a source of inspiration... Maybe we can implement a `TooltipReference` component that also does renderProps.

I think the most important thing to realise is that even if we were to implement this helper Component, or even switch to `react-popper`, the core problem remains:
- The popper needs a ref to a reference-element
- So whatever is being used as a Tooltip child needs to be able to hold a `ref`

So, thinking of a `Button` (ui-core component) with a `Tooltip` the following example would NOT work:
```jsx
    <Reference>
      {({ ref }) => (
        <Button ref={ref}>
          Reference element
        </Button>
      )}
    </Reference>
```

The reason is that our `Button` can't hold a ref. So the app would either need to implement a wrapper around the `Button` that is able hold a ref, or ui-core needs to make sure all/most of its components can hold a ref.

I'm not really in favour of reworking all of our component, just to make the Tooltip work[*], so letting the app wrap whatever they need seems to make more sense.

[*] Perhaps letting all of our components accept a ref isn't such a bad idea in general, but we shouldn't start adding this just because we need to fix the Tooltip.

PR #836, opened 2020-03-16
refactor(select): power select by popper

HendrikThePendric:
This PR replaces the custom menu-positioning logic from to Select by the Popper. I have also added tests for the positioning strategies we use (default > flip > shift).  I remember we had some issue with page scrolling... Not sure if a test for that was already added when the issue was solved, otherwise perhaps I should add one...


cypress/integration/Select/position.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Position of select menu dropdown
```
Mohammer5: We're not exposing the select. Currently this is only a helper to reduce shared code between the `SingleSelect` and the `MultiSelect`. I think we should have two (probably identical) feature files for those two, then the underlying architecture doesn't matter

PR #843, opened 2020-03-24
feat: tooltip with support for renderprops

HendrikThePendric:
The purpose of this PR is mainly to demonstrate a generic approach to dealing with components that need to be able to render any content, but do require this content to implement certain props. 

This is an example for the `Tooltip`, but we see the same/a very similar problem appear in  the `Menu`/`SubMenu` too.

The main idea is that we offer two methods of rendering:
- For default scenarios we accept children in the normal way (CONVENIENT)
- And if further customization is required we switch to the renderProps approach (FLEXIBLE)

So for the Tooltip this means:
- Normally children are simply wrapped in a `span` and the ref, OnMouseOver, onMouseOut props are attached to that in the Tooltip. If the app wants to use a different tag, this can be customised via the `tag` prop (see demo CustomElementViaTagProp)
- If further customisation is needed, the app can switch to renderProps (see CustomBuiltInComponent and CustomComponent)

An important note for that last example, CustomComponent: 
In the example you see that I am adding a wrapper around the Button. This is for two reasons:
1. The Button is currently not able to hold a ref. But we can see in @dhis2/ui that this is going to change...
2. The Tooltip anchor needs to implement not only a ref, but also a `onMouseOver` and `onMouseOut`, and our Button has not implemented those methods either.

There's nothing wrong with a wrapper element per se, but if we look into the future, i.e. `@dhis2/ui` , where the `*Base` components [can all hold a ref](https://github.com/dhis2/ui/blob/alpha/packages/core/src/ButtonBase/ButtonBase.js#L69) and [forward props](https://github.com/dhis2/ui/blob/alpha/packages/core/src/ButtonBase/ButtonBase.js#L29), we could also solve this in a different way, using the `ButtonBase`:

_quick and dirty:_
```jsx
<p>
    I am a{' '}
    <Tooltip content="Some extra info">
        {tooltipProps => <ButtonBase {...tooltipProps}>paragraph</ButtonBase>}
    </Tooltip>{' '}
    that contains a Button with a tooltip.
</p>
```

_dedicated component:_
```jsx
const TooltipButton = ({ children, content, placement, onClick }) => (
    <Tooltip content={content} placement={placement}>
        {tooltipRenderProps => (
            <ButtonBase {...tooltipRenderProps} onClick={onClick}>
                {children}
            </ButtonBase>
        )}
    </Tooltip>
)

<p>I am a paragrapg that contains a <TooltipButton content="Some extra info">tooltip button</TooltipButton>.</p>
```

**If I've understood the feedback I've received recently correctly, then one of the main problems that has been mentioned is the lack of transparency. Does the current approach with renderProps solve that in your opinions?**

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -65,7 +65,7 @@ class Tooltip extends Component {
         const {
             children,
             className,
-            component: Component,
+            tag: Element,
             content,
             dataTest,
             maxWidth,
@@ -75,14 +75,22 @@ class Tooltip extends Component {
 
         return (
             <>
-                <Component
-                    onMouseOver={this.onMouseOver}
-                    onMouseOut={this.onMouseOut}
-                    ref={this.ref}
-                    data-test={`${dataTest}-reference`}
-                >
-                    {children}
-                </Component>
+                {typeof children === 'function' ? (
+                    children({
+                        onMouseOver: this.onMouseOver,
+                        onMouseOut: this.onMouseOut,
+                        ref: this.ref,
+                    })
+                ) : (
+                    <Element
+                        onMouseOver={this.onMouseOver}
+                        onMouseOut={this.onMouseOut}
+                        ref={this.ref}
+                        data-test={`${dataTest}-reference`}
+                    >
+                        {children}
+                    </Element>
+                )}
 
                 {open &&
                     createPortal(
@@ -122,9 +130,9 @@ class Tooltip extends Component {
 
 Tooltip.defaultProps = {
     dataTest: 'dhis2-uicore-tooltip',
-    component: 'span',
     maxWidth: 300,
     placement: 'top',
+    tag: 'span',
 }
 
 /**
@@ -132,41 +140,20 @@ Tooltip.defaultProps = {
  * @static
  * @prop {Node} [children]
  * @prop {string} [className]
- * @prop {string|function} [component] Use a custom component to wrap the Tooltip children or pass a string (i.e. 'p') to use a custom built-in component
- * @example
- * const Content = React.forwardRef(
- *     ({ children, onMouseOver, onMouseOut, dataTest }, ref) => (
- *         <em
- *             style={{ display: 'inline-block', color: 'red' }}
- *             ref={ref}
- *             onMouseOver={onMouseOver}
- *             onMouseOut={onMouseOut}
- *             dataTest={`${dataTest}-reference`}
- *         >
- *             {children}
- *         </em>
- *     )
- * )
- * Content.displayName = 'Content'
- * Content.propTypes = {
- *     children: propTypes.node,
- *     dataTest: propTypes.string,
- *     onMouseOut: propTypes.func,
- *     onMouseOver: propTypes.func,
- * }
  * @prop {Node} [content]
- * @prop {string} [dataTest]
- * @prop {number} [maxWidth]
+ * @prop {string} [dataTest=dhis2-uicore-tooltip]
+ * @prop {number} [maxWidth=300]
  * @prop {('top'|'bottom'|'right'|'left')} [placement=top]
+ * @prop {string} [tag=span] The HTML tag to render
  */
 Tooltip.propTypes = {
     children: propTypes.node,
     className: propTypes.string,
-    component: propTypes.oneOfType([propTypes.string, propTypes.func]),
     content: propTypes.node,
     dataTest: propTypes.string,
     maxWidth: propTypes.number,
     placement: propTypes.oneOf(['top', 'right', 'bottom', 'left']),
+    tag: propTypes.string,
 }
 
 export { Tooltip, TOOLTIP_OFFSET }
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -65,7 +65,6 @@ class Tooltip extends Component {
         const {
             children,
             className,
-            tag: Element,
             content,
             dataTest,
             maxWidth,
@@ -82,14 +81,14 @@ class Tooltip extends Component {
                         ref: this.ref,
                     })
                 ) : (
-                    <Element
+                    <span
                         onMouseOver={this.onMouseOver}
                         onMouseOut={this.onMouseOut}
                         ref={this.ref}
                         data-test={`${dataTest}-reference`}
                     >
                         {children}
-                    </Element>
+                    </span>
                 )}
 
                 {open &&
@@ -144,7 +143,6 @@ Tooltip.defaultProps = {
  * @prop {string} [dataTest=dhis2-uicore-tooltip]
  * @prop {number} [maxWidth=300]
  * @prop {('top'|'bottom'|'right'|'left')} [placement=top]
- * @prop {string} [tag=span] The HTML tag to render
  */
 Tooltip.propTypes = {
     children: propTypes.node,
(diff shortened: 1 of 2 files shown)


HendrikThePendric:
That's quite an interesting idea.... The biggest advantage would be that both the Tooltip and the anchor element could just render children. To me the downside is that you sort of loose the visual connection/relationship between the tooltip and the anchor...

I'd be interested to find out what @ismay and @varl think.

One minor detail that I'd like to change from your suggestion is how we actually toggle the Tooltip visibility. I'd just use conditional rendering:
```jsx
{showTooltip && <Tooltip anchor={ref}>Some extra info</Tooltip>}
```

HendrikThePendric:
🤔 after reviewing a number of other react tooltip implementations, it does seem to be that the approach I went with initially, i.e. wrapping the anchor in the tooltip seems to be the most common pattern. So, perhaps that'd be an argument for sticking to the current approach.

varl:
If the hook-based approach allows us to use any component (HTML or React) as an anchor, and we can ignore the display mode (inline vs. block) then I like that approach.

Mohammer5:
@varl yeah it'd work with any element (built-in or custom).

I'd like to slightly change what I've proposed though:

```jsx
export const CustomComponent = () => {
    const { ref, onMouseOver, onMouseOut, tooltip } = useTooltip()

    return (
        <p>
            I am a{' '}
            <span ref={ref} onMouseOver={onMouseOver} onMouseOut={onMouseOut}>
                <Button>paragraph</Button>
            </span>
            that contains a Button with a tooltip.

            <Tooltip {...tooltip}>Some extra info</Tooltip>

            <style jsx>{`
                span {
                   display: inline-flex;
                }
            `}</style>
        </p>
    )
}
```

This way we can add/remove props we want/need to pass to the tooltip without breaking apps.
Apps also don't have to care about open/close states anymore.

<hr />

> To me the downside is that you sort of loose the visual connection/relationship between the tooltip and the anchor...

The tooltip could be placed inside the anchor. It's rendered in a portal anyways, isn't it?

HendrikThePendric:
> If the hook-based approach allows us to use any component (HTML or React) as an anchor, and we can ignore the display mode (inline vs. block) then I like that approach.

@varl but so would the renderprops approach

HendrikThePendric:
Let's make the choice fully explicit. The following examples are functionally identical and just use slightly different syntax:

OPTION A:
```jsx
export const CustomComponent = () => (
    <p>
        I am a{' '}
        <Tooltip content="Some extra info">
            {tooltipAnchorProps => (
                <span {...tooltipAnchorProps}>
                    <Button>paragraph</Button>
                    <style jsx>{`
                        span {
                            display: inline-flex;
                        }
                    `}</style>
                </span>
            )}
        </Tooltip>{' '}
        that contains a Button with a tooltip.
    </p>
)
```

OPTION B:
```jsx
export const CustomComponent = () => {
    const { ref, onMouseOver, onMouseOut, tooltip } = useTooltip()

    return (
        <p>
            I am a{' '}
            <span ref={ref} onMouseOver={onMouseOver} onMouseOut={onMouseOut}>
                <Button>paragraph</Button>
                <Tooltip {...tooltip}>Some extra info</Tooltip>
            </span>{' '}
            that contains a Button with a tooltip.

            <style jsx>{`
                span {
                   display: inline-flex;
                }
            `}</style>
        </p>
    )
}
```

ismay:
Maybe this discussion is also relevant to this PR? https://github.com/dhis2/design-system/issues/23. It might be that the tooltip will become a minor, escape hatch, type of component in our ecosystem (though nothing has been decided yet). So maybe that also means that it's ok if the tooltip doesn't do too much, and won't be applicable for all usecases.

But to reply to the discussion above. From my perspective a combination of what @Mohammer5 suggested and what @HendrikThePendric suggested [below that](https://github.com/dhis2/ui-core/pull/843#issuecomment-603344126) seems pragmatic and relatively easy to understand:

```jsx
export const CustomComponent = () => {
    const { ref, onMouseOver, onMouseOut, showTooltip } = useTooltip()

    return (
        <p>
            I am a{' '}
            <span ref={ref} onMouseOver={onMouseOver} onMouseOut={onMouseOut}>
                <Button>paragraph</Button>
            </span>
            that contains a Button with a tooltip.

            {showTooltip && <Tooltip anchor={ref}>Some extra info</Tooltip>}

            <style jsx>{`
                span {
                   display: inline-flex;
                }
            `}</style>
        </p>
    )
}
```

Seems simple and explicit to me, plus it doesn't make the pattern _too_ convenient, so that hopefully it won't be overused and we can nudge people towards a toggletip.

I'm not that sure about the revised example @Mohammer5 suggested after that [here](https://github.com/dhis2/ui-core/pull/843#issuecomment-603766318). It seems slightly less explicit to me, so it's a tad harder to see at a glance what's going on. I see the benefit of being able to pass anything to the tooltip without requiring devs to interfere, but I don't think that's that bad and prefer the explicit route personally.

ismay:
So I minimized the above comment after talking a bit about the implementation with @HendrikThePendric since I see things a bit differently now. So, the main difference between the two proposals seems to be if we use renderProps or hooks. Functionally they basically allow the same thing. But that's not the crux of the matter maybe, I'll elaborate below.

I looked around and found this https://kentcdodds.com/blog/react-hooks-whats-going-to-happen-to-render-props, and that led us to talking about whether here the renderProp pattern fits with the usecase where Kent still recommends renderProps.

And I agree with @HendrikThePendric, seems like renderProps are a nice fit here. If I understand correctly, we'll have a convenient option for users that just need a default Tooltip. They can pass elements as children and just use the defaults.

For those who need more customization, they can pass a function and have more freedom in what gets rendered exactly. So an escape hatch for more complicated situations. That seems quite user friendly to me.

With the hooks they'll have to always use the hook, even though they might just want a regular tooltip. That would probably lead to them creating a default tooltip locally for the regular tooltips, which seems like duplicated effort. Plus, none of our other components are currently hooks, so it would be a bit uncommon for our lib atm.

So long story short, I see @HendrikThePendric's point. I think that his suggestion would be what I'd prefer from the perspective of a dev using this component.

varl:
I think the `style jsx` tag needs to be moved to top-level in **Option A**, other than that good examples.

> * A convenient option for users that just need a default Tooltip. They can pass elements as children and just use the defaults.
> 
> * For those who need more customization, they can pass a function and have more freedom in what gets rendered exactly. So an escape hatch for more complicated situations. That seems quite user friendly to me.

Given that these two are correct, I am swayed as well.

All examples here have omitted the imports which for **Option A** would be slightly simpler than **Option B**.

Will this decision affect other components as well? Seems like we are establishing a pattern here.

Mohammer5:
> Given that these two are correct, I am swayed as well.

Agreed. Good points @ismay !

HendrikThePendric:
> Will this decision affect other components as well? Seems like we are establishing a pattern here.

I was thinking about doing something similar for a nested Menu, so we end up with something as flexible as @Mohammer5 [suggested on Slack](https://dhis2.slack.com/archives/CBM8LNEQM/p1585043814115900?thread_ts=1585037324.086900&cid=CBM8LNEQM), but a bit more convenient. I still have to work out what the best structure would be, but the problem we need to solve there is identical to the Tooltip:
- A MenuItem can be a custom component
- A MenuItem can also be the anchor element for a sub-menu
- We'll need a single ref that we can a) apply to the MenuItem, and b) pass to the Popper

So I'd say here renderProps will help too because it would offer an easy way to provide the ref to both of these...

Ideally I'd also like to end up with something similar to the Tooltip, so we have a convenient way to render a default MenuItem plus an escape hatch via renderProps for custom stuff.

PR #845, opened 2020-03-25
test(layer-context): demonstrate incorrect behaviour in e2e test

HendrikThePendric:
The purpose of this PR is to demonstrate the limitations of the current layering system (i.e. the `Layer` from `./src/LayerContext.js`). There is a story and an e2e test to show what's wrong. The example might be a bit contrived, but I hope I managed to capture the essence of the problem.

Perhaps for that extra dosage of clarity:
- The test is failing
- But if our layering solution was correct, it should be passing
- As such this is _evidence_ that our current solution is not good enough

And my suggestion is that I prepare a similar test for the revised layering system, to prove that this one does work. We can then focus on moving forward with the new layering system and deprecate the old one.


HendrikThePendric:
UPDATE:
In c912fa8 I've made the following changes:
- Changed the original e2e test to just try clicking the element that we expect on top. This still fails because cypress detects that the element is being covered by something else
- I've introduced an identical story and e2e test for the new Layer component. This passes, so demonstrates the new approach does stacking better.

HendrikThePendric:
This was superseded by https://github.com/dhis2/ui/pull/10

Issue #760, opened 2020-01-23
Reorganise overlay components


HendrikThePendric:
Currently we have several components with overlapping responsibility and some duplication. Some things I noticed:
-`ScreenCover` uses a `Backdrop`, but it’s a private one, not the one that we expose..
- Suppose the (public) `Backdrop` is meant to be placed into a `ScreenCover` or `ComponentCover` (I think it should), then I think the styling it has is way too opinionated, it should just cover the entire parent, not work with a position: fixed.
- The `ScreenCover` is rendering its children into a `Content` component, which vertically and horizontally centers its content. Again, I think this is too opinionated.
- The `ComponentCover` does something similar

So I’d like to propose, for the next major release to change things around a bit:
- `ScreenCover` covers the window
- `ComponentCover` covers the component, provided the parent has a non-static position
- `Backdrop`: can accept an `onClick`, has a semi-transparent background by default and a `transparent` prop
- `CenteredContent`: aligns its content vertically and horizontally with it’s children

If we structure the components like this, it's pretty to build a variety of blocking/non-blocking overlays with/without loaders (or other centered content)

PR #846, opened 2020-03-25
refactor: rearrange overlay components

HendrikThePendric:
This PR is not done yet, but does illustrate how I would like to rearrange the overlay components for v5. When done, this PR would solve https://github.com/dhis2/ui-core/issues/760. I've tackled things slightly differently than I wrote in that issue, see below:
- `Layer` covers the window or body and takes care of stacking (z-index). You can also give this component a semi-transparent background by adding the `translucent` prop. It's also possible to create a "non-blocking" layer by setting `pointerEvents="none"`. And finally, you can attach an `onClick` handler.
- `ComponentCover` covers the component, provided the parent has a non-static position. The other props are identical to the `Layer` (i.e. `onClick`, `translucent`, `pointerEvents`) 
- `CenteredContent`: aligns its content vertically and horizontally.

The following components would be deprecated and replaced by the new ones above:
- `Backdrop`
- `Layer` (from `LayerContext.js`)

### DIFF
--- a/src/CenteredContent.js
+++ b/src/CenteredContent.js
@@ -1,0 +1,44 @@
+import React from 'react'
+import propTypes from '@dhis2/prop-types'
+;('') // TODO: https://github.com/jsdoc/jsdoc/issues/1718
+
+/**
+ * @module
+ * @param {CenteredContent.PropTypes} props
+ * @returns {React.Component}
+ * @example import { CenteredContent } from @dhis2/ui-core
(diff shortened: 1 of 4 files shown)


HendrikThePendric:
Small note: it looks like there are a lot of changes in the stories, but this is just formatting, the only new story is "'With centered content and click".

HendrikThePendric:
Closing this in favour of https://github.com/dhis2/ui/pull/10

PR #859, opened 2020-04-02
feat: chip adjustments

cooper-joe:
This PR makes additions and adjustments to the `Chip` component:
- added `dense` style
- adjusted remove icon to be less visually dominating

### DIFF
--- a/src/Chip/Remove.js
+++ b/src/Chip/Remove.js
@@ -2,30 +2,33 @@ import React from 'react'
 import propTypes from '@dhis2/prop-types'
 import { css, resolve } from 'styled-jsx/css'
 
-import { Cancel } from '../icons/Cancel'
+import { CancelOutline } from '../icons/Cancel--outline.js'
 import { colors } from '../theme.js'
 
 const containerStyle = css`
     span {
-        display: block;
-        height: 18px;
-        width: 18px;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        height: 20px;
+        width: 20px;
         margin-right: 4px;
+        border-radius: 12px;
+        margin-left: -8px;
+    }
+    span:hover {
+        background: ${colors.grey400};
     }
 `
 
 const removeIcon = resolve`
     svg {
-        color: ${colors.grey700};
-		height: 18px;
-		width: 18px;
+        fill: ${colors.grey600};
+		    height: 16px;
+		    width: 16px;
         cursor: pointer;
         opacity: 1;
         pointer-events: all;
-    }
-
-    svg:hover {
-        opacity: 0.82;
     }
 `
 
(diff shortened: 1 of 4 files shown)


src/Chip/Remove.js
```diff
@@ -2,31 +2,34 @@ import React from 'react'
...
 const containerStyle = css`
     span {
-        display: block;
-        height: 18px;
-        width: 18px;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        height: 20px;
+        width: 20px;
```
edoardo, line 14: One question about style rules like this one.
Are we suppose to use `spacers` variables from the theme?
This is also valid for app level style files, I'm trying to use variables from the theme as much as possible but in a few occasions I didn't have the variable I needed so I hardcoded the value.

ismay: > Are we suppose to use spacers variables from the theme? This is also valid for app level style files, I'm trying to use variables from the theme as much as possible but in a few occasions I didn't have the variable I needed so I hardcoded the value.

That has been my approach as well.

cooper-joe: Where possible `spacer` should be used, yeah. However, a spacer doesn't need to be used unless it is doing 'spacing'. In this example the icon is `16`, but I've set the w/h to `20` to give a larger click area for the user. The unit of `20` here is not working as a generic spacer, it's specific to this element.

For actual spacing between elements (list items, for example) it's always good to use the right `spacer` variable. Sounds like you guys are using it in the intended way 👍

ismay: Speaking of using the spacer vars, is there a reason that the pixel values are mentioned in the css var names? To me, using relative names (like spacer-sm, spacer-md, spacer-lg, spacer-xl) seem like a nice way to have variables that you can update without invalidating the names.

cooper-joe: @ismay please excuse my delayed response.

Thinking back, I think the reason the values are mentioned in the spacer variables was because I could not think of a non-confusing alternative. There are not enough size abbreviations to do `spacer-m`, `spacer-l` etc. I solved this with colors by doing `grey100`, `grey200` etc. I thought it would be very confusing to do `spacer100` that was actually `24px`. Mixing values was not nice.

In the end, I justified the choice by saying that the spacer variables were not intended to change, but rather to help achieve a consistent design across apps on an 8pt grid. If we ever want to change those spacers it will be a mess, but, on balance, it seemed like the quickest way to get the spacers up and running.

varl: We did a two-step scale with colors and we could have the same strategy for spacers.

With colors we map e.g. `primary900` to `blue900`: https://github.com/dhis2/ui-core/blob/master/src/theme.js#L84

This allows us to redefine the `primary900` if we need to, without changing the definition of `blue900`. We could could say that `spacer-xl` or `spacer900` is `dp24` or what have you.

ismay: Yeah that seems practical to me 👍

PR #875, opened 2020-04-14
fix(MultiSelect): use dense chips for selections in input

cooper-joe:
This PR changes the `Chip` type used in a `MultiSelect`. It adds a `dense` prop to the `Chip`.

Note: is this the correct conventional commit message type? It's not a `fix:` because this was not a bug, but is there something before `feat:`, more like...`adjustment: `? 😉

### DIFF
--- a/src/MultiSelect/SelectionList.js
+++ b/src/MultiSelect/SelectionList.js
@@ -47,6 +47,7 @@ const SelectionList = ({ selected, onChange, disabled, options }) => (
                     onRemove={onRemove}
                     disabled={isDisabled}
                     overflow
+                    dense
                 >
                     {label}
                 </Chip>


cooper-joe:
@edoardo This is the change you are waiting on for Data Visualizer YoY controls.

Mohammer5:
@varl @ismay should this go into `ui-core` or `ui`?

ismay:
I'd personally say ui@alpha, the NoticeBox is also there for example. Otherwise we have changes in ui-core that we'd have to pull into ui. But I don't know what @varl prefers. Plus it depends on the timeline for this change (it will take a bit until the ui stable release), ui-core could be released r.n.

varl:
I'd say that this is a `fix:, as a fix doesn't have to be a straight up bug fix, it can be a design fix or something else that fixes something without introducing a new feature.

As for where it goes, merging from ui-core to ui is relatively straight forward, so this can go here as it is being waited on.

cooper-joe:
> I'd say that this is a `fix:`, as a fix doesn't have to be a straight up bug fix, it can be a design fix or something else that fixes something without introducing a new feature.

Thanks for the clarification @varl, that makes sense to me now. I've changed the commit message.

PR #877, opened 2020-04-14
docs: adds prefix to jsdocs link since they were returning 404

paschalidi:
👋 hello! 

This is my first PR without opening an issue. Maybe I should have opened one first. Feel free to give me your feedback both on the code and the procedure anyways! 

## The issue description:
While I was navigating at the [ui-core docs](https://ui-core.dhis2.nu/#/api) I realised that clicking on the Live demo link would return a 404 page instead of the story. 
image

The problem is that we are having this `/#/` on the url and therefore the link is broken. I added the full link to fix this issue however I can imagine that there might be a better solution to this! Le me know your thoughts and feedback!

### DIFF
--- a/src/AlertBar.js
+++ b/src/AlertBar.js
@@ -18,7 +18,7 @@ import styles, { ANIMATION_TIME } from './AlertBar/styles.js'
  * @example import { AlertBar } from @dhis2/ui-core
  *
  * @see Specification: {@link https://github.com/dhis2/design-system/blob/master/molecules/alertbar.md|Design system}
- * @see Live demo: {@link /demo/?path=/story/alertbar--default|Storybook}
+ * @see Live demo: {@link https://ui-core.dhis2.nu/demo/?path=/story/alertbar--default|Storybook}
  */
 class AlertBar extends Component {
     // visible is used to control the show/hide animation
--- a/src/AlertStack.js
+++ b/src/AlertStack.js
@@ -11,7 +11,7 @@ import { layers } from './theme.js'
  * @param {AlertStack.PropTypes} props
  * @returns {React.Component}
  * @example import { AlertStack } from '@dhis2/ui-core'
- * @see Live demo: {@link /demo/?path=/story/alertstack--default|Storybook}
+ * @see Live demo: {@link https://ui-core.dhis2.nu/demo/?path=/story/alertstack--default|Storybook}
  */
 const AlertStack = ({ className, children, dataTest }) =>
     createPortal(
--- a/src/Button.js
+++ b/src/Button.js
@@ -13,7 +13,7 @@ import styles from './Button/styles.js'
  *
  * @example import { Button } from @dhis2/ui-core
  * @see Specification: {@link https://github.com/dhis2/design-system/blob/master/atoms/button.md|Design system}
- * @see Live demo: {@link /demo/?path=/story/button-basic--default|Storybook}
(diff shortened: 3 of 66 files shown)


ismay:
Thanks! I guess the hash is considered part of the root, and then it appends the root relative url to that. The fix does mean that the links in the api docs will now always point to the live production docs. So the links in PR builds will now also point to the 'production' docs, whereas previously they'd link to the docs from the PR. So:

* now it's: https://ui-core.dhis2.nu/demo/?path=%2fstory%2falertbar--default
* but this PR's build is at: https://deploy-preview-877--dhis2-ui-core.netlify.com/demo/?path=%2fstory%2falertbar--default

I haven't looked into what's going on with the jsdocs. Maybe the routing there could be changed from hash routing? That'd allow us to get rid of the hash.

ismay:
The build is done by https://github.com/dhis2/cli-utils-docsite, so if we want to change anything I assume we'd have to change it there. I'm sure of the history of our doc builds though, and what the plans are. I think @amcgee and @varl know more about that.

paschalidi:
> Thanks! I guess the hash is considered part of the root, and then it appends the root relative url to that? The fix does mean that the links in the api docs will now always point to the live production docs. So the links in PR builds will now also point to the 'production' docs, whereas previously they'd link to the docs from the PR. So:
> 
> * With the fix it points to: https://ui-core.dhis2.nu/demo/?path=%2fstory%2falertbar--default
> * but this PR's build is actually at: https://deploy-preview-877--dhis2-ui-core.netlify.com/demo/?path=%2fstory%2falertbar--default
> 
> I haven't looked into what's going on with the jsdocs. Maybe the routing there could be changed from hash routing? That'd allow us to get rid of the hash.

Yeah you right havent thought of that. Apologies. 

Best would be as you say to remove the hash. Do know @ismay what is the use of the hash?

ismay:
Actually, my assumption was wrong. The problem is with the generated html. It's generated by https://docsify.js.org, which prepends the hash, so it turns this: 

```
/demo/?path=/story/alertbar--default|Storybook
```

into this:

```
<a href="#/demo/?path=%2fstory%2falertbar--default">Storybook</a>
```

The hash is used for docsify's routing, but it shouldn't be prepended for these links. I'll look into if it can be omitted somehow. Will report back.

paschalidi:
Thanks for taking the initiative! Will be willing to help dont hesitate to ask

ismay:
I think that this is what we're looking for https://docsify.js.org/#/configuration?id=nocompilelinks. And our docsify configuration can be found here: https://github.com/dhis2/cli-utils-docsite/blob/master/template/js/initDocsify.js

I assume that setting:

```
noCompileLinks: ['/demo/.*']
```

would solve our issue. (Or something of that nature, I don't know precisely what glob syntax they expect)

We could also change the [router mode](https://docsify.js.org/#/configuration?id=routermode), but that would require us to configure redirects, so the above might be the simplest option.

ismay:
@paschalidi I'm hoping that this'll do the trick: https://github.com/dhis2/cli-utils-docsite/pull/47

paschalidi:
cool! I am closing this! Thanks for taking care of it!!

ismay:
> cool! I am closing this! Thanks for taking care of it!!

N.p., thanks for picking this up (I'm not sure whether my proposed fix is how we want to fix it btw., but we'll see).

Thanks again for the PR, it's easy to miss these kinds of hiccups with documentation, much appreciated 👍!

PR #886, opened 2020-05-18
feat: let popper and popover accept react-refs or dom nodes directly

HendrikThePendric:
This PR is essentially identical to https://github.com/dhis2/ui/pull/71 (apart from some import paths). Since @edoardo needs the popper fixed for patch release `2.34.1`, this needs to be implemented in ui-core too.

I guess the next steps should be something like this:
- Merge this into master once approved so a new minor version is released
- When we pull in the upstream (`@dhis2/ui-core`) changes to `@dhis2/ui`, we should fix the import paths and documentation links
- We can then close https://github.com/dhis2/ui/pull/71 because that will have become redundant/superseded...
- I'd also like to add some additional e2e test for arrow positions, I guess I can just add them later (in `@dhis2/ui`)

### DIFF
--- a/src/Popover.js
+++ b/src/Popover.js
@@ -1,20 +1,16 @@
-import React from 'react'
-import { createPortal } from 'react-dom'
-import propTypes from 'prop-types'
+import React, { useState, useMemo } from 'react'
+import propTypes from '@dhis2/prop-types'
+import { usePopper } from 'react-popper'
 
+import { colors, elevations } from './theme.js'
+import { Layer } from './Layer.js'
+import { combineModifiers } from './Popover/modifiers.js'
+import { Arrow } from './Popover/Arrow.js'
 import {
     elementRefPropType,
     referencePlacementPropType,
 } from './common-prop-types.js'
 
-import { Popper } from './Popper'
-import { Backdrop } from './Backdrop'
-
-import { Arrow } from './Popover/Arrow'
-import { arrow, offset, hideArrowWhenDisplaced } from './Popover/modifiers.js'
-import { colors, elevations } from './theme.js'
-
-const arroModifiers = [arrow, offset, hideArrowWhenDisplaced]
 /**
  * @module
  * @param {Popover.PropTypes} props
@@ -34,34 +30,66 @@ const Popover = ({
     dataTest,
     elevation,
     maxWidth,
+    observePopperResize,
+    observeReferenceResize,
     placement,
     onBackdropClick,
-}) =>
-    createPortal(
-        <Backdrop onClick={onBackdropClick} transparent>
-            <Popper
-                dataTest={`${dataTest}-popper`}
-                placement={placement}
-                reference={reference}
-                modifiers={arrow ? arroModifiers : []}
-                className={className}
-            >
-                <div>
-                    {children}
-                    {arrow && <Arrow />}
-                    <style jsx>{`
-                        div {
-                            max-width: ${maxWidth}px;
-                            box-shadow: ${elevation};
-                            background-color: ${colors.white};
-                            border-radius: 4px;
-                        }
-                    `}</style>
-                </div>
-            </Popper>
-        </Backdrop>,
-        document.body
+}) => {
+    const referenceElement =
+        reference instanceof Element
+            ? reference
+            : reference && reference.current
+    const [popperElement, setPopperElement] = useState(null)
+    const [arrowElement, setArrowElement] = useState(null)
+    const modifiers = useMemo(
+        () =>
+            combineModifiers(arrow, arrowElement, {
+                observePopperResize,
+                observeReferenceResize,
+            }),
+        [arrow, arrowElement, observePopperResize, observeReferenceResize]
     )
+    const { styles, attributes } = usePopper(referenceElement, popperElement, {
+        placement,
+        modifiers,
+    })
+
+    return (
+        <Layer onClick={onBackdropClick}>
+            <div
+                data-test={dataTest}
+                className={className}
+                ref={setPopperElement}
+                style={styles.popper}
+                {...attributes.popper}
+            >
+                {children}
+                {arrow && (
+                    <Arrow
+                        hidden={
+                            attributes.arrow &&
+                            attributes.arrow['data-arrow-hidden']
+                        }
+                        popperPlacement={
+                            attributes.popper &&
+                            attributes.popper['data-popper-placement']
+                        }
+                        ref={setArrowElement}
+                        styles={styles.arrow}
+                    />
+                )}
+                <style jsx>{`
+                    div {
+                        max-width: ${maxWidth}px;
+                        box-shadow: ${elevation};
+                        background-color: ${colors.white};
+                        border-radius: 4px;
+                    }
+                `}</style>
+            </div>
+        </Layer>
+    )
+}
 
 Popover.defaultProps = {
     arrow: true,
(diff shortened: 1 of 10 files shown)


src/Popover/Arrow.js
```diff
@@ -1,79 +1,91 @@
...
+                }
+            `}</style>
+        </div>
+    )
+})
+Arrow.displayName = 'Arrow'
 
 Arrow.propTypes = {
-    className: propTypes.string,
+    attributes: propTypes.object,
```
amcgee, line 82: should we give this a `shape` with `arrow` and `popper` keys and an optional `arrow.data-arrow-hidden`?

src/Popover/modifiers.js
```diff
@@ -11,35 +12,47 @@ const computeArrowPadding = () => {
...
-}
+const hideArrowWhenDisplaced = ({ state }) => {
+    const halfArrow = ARROW_SIZE / 2
+    const displacement = state.modifiersData.preventOverflow
+    const referenceRect = state.rects.reference
+    const shouldHideArrow =
+        Math.abs(displacement.x) >= referenceRect.width + halfArrow ||
+        Math.abs(displacement.y) >= referenceRect.height + halfArrow
+
+    state.attributes.arrow = { ['data-arrow-hidden']: shouldHideArrow }
```
amcgee, line 23: Is there a reason this is passed as a data- prop under `arrow` rather than just as a normal `hidden` prop to the `Arrow` component?

amcgee: Seems like maybe I'm missing something about the `react-popper` API here?

src/Popover/Arrow.js
amcgee: (possibly without `arrow`, see below)

src/Popover.js
```diff
@@ -80,18 +97,23 @@ Popover.defaultProps = {
...
 Popover.propTypes = {
     children: propTypes.node.isRequired,
     reference: elementRefPropType.isRequired,
     arrow: propTypes.bool,
     className: propTypes.string,
     dataTest: propTypes.string,
+    elevation: propTypes.string,
     maxWidth: propTypes.number,
+    observePopperResize: propTypes.bool,
+    observeReferenceResize: propTypes.bool,
```
amcgee, line 114: Should these default to `true`?

HendrikThePendric: No, I think they should default to something falsy. The app-developer can add these props if there are reasons to expect the element size will change. Also note that we do include the `eventListeners` modifier by default, which will make sure the popper updates on `window.resize` (and scroll).

amcgee: Hmmm.... this could maybe be problematic?  We use `div` scrolling rather than `window` scrolling in a few places, including the platform layout...

src/Popover/Arrow.js
HendrikThePendric: Ultimately the arrow needs these two:
- `attributes.popper['data-popper-placement']`
- `attributes.arrow['data-arrow-hidden']`

Both `attributes.popper` and `attributes.arrow` can be undefined (during the first render cycle).

So currently I decided to just pass the `attributes` to the arrow and process them there.

Instead I could also use these two props:
- `popperPlacement`
- `hidden`

And effectively this means moving some complexity from the Arrow to the Popover.

Would you prefer this?
```jsx
<Arrow
    hidden={attributes.arrow && attributes.arrow['data-arrow-hidden']}
    popperPlacement={attributes.popper && attributes.popper['data-popper-placement']}
    ref={setArrowElement}
    styles={styles.arrow}
/>
```

amcgee: OK, but don't we control `attributes.arrow.data-arrow-hidden` [here (line 23)](https://github.com/dhis2/ui-core/blob/76ffa937f3a2211e9f4c4daa3c40311d7fa3b3a0/src/Popover/modifiers.js#L23) and only use it [here](https://github.com/dhis2/ui-core/blob/76ffa937f3a2211e9f4c4daa3c40311d7fa3b3a0/src/Popover/Arrow.js#L11) ?  Maybe it can also be set by `react-popper` but that's not really clear to me here, so I think I'd advocate for colocating the parsing of the hook response with the hook call, and a simple primitive props to our components.

src/Popover.js
HendrikThePendric: > [...] updates on window.resize (and scroll)

Sorry this part of my comment is a bit ambiguous... This modifier listens to `window.resize` and _various_ scroll events, [see here](https://github.com/popperjs/popper-core/blob/master/src/modifiers/eventListeners.js#L17-L26).

But in any case, if/when we encounter problems, we can add an additional modifier to fix things.

src/Popover/modifiers.js
HendrikThePendric: Yeah this is the way of popper.js, you add behaviour via modifiers. And, when used directly, popper.js will attach data-attributes to elements, hence the kebab-cased property names. We're mainly using modifiers from popper.js itself, which are already using that kebab-cased approach, so I decided to just stick to that in our custom modifiers.

src/Popover/Arrow.js
HendrikThePendric: Implemented with primitives in https://github.com/dhis2/ui-core/pull/886/commits/4eaafb66220bf396f2feb1c8da06c101dad60742. Can this now be resolved?

HendrikThePendric:
One thing that came up regularly during conversations about these components, is what the `reference` prop should actually be. People might be adding a DOM node directly, and in some cases that is actually the most convenient way to work. So I've added support for that in https://github.com/dhis2/ui-core/pull/886/commits/7e5bf142f4e5defd5c4118186f2a3ea383d0a9f2

src/Popover.js
```diff
@@ -34,34 +30,66 @@ const Popover = ({
     dataTest,
     elevation,
     maxWidth,
+    observePopperResize,
+    observeReferenceResize,
     placement,
-    onBackdropClick,
```
amcgee, line 38: This makes this a breaking change, can we keep it `onBackdropClick` since this is going into `4.x`?

src/Popover.js
```diff
@@ -34,34 +30,66 @@ const Popover = ({
...
-                dataTest={`${dataTest}-popper`}
-                placement={placement}
-                reference={reference}
-                modifiers={arrow ? arroModifiers : []}
+    onClickOutside,
+}) => {
+    const referenceElement =
+        reference instanceof Element
+            ? reference
+            : reference && reference.current
```
amcgee, line 41: Minor : this can be written `reference?.current` with [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), which we support in our babel config.  Optional if you prefer `&&` here though.

src/Popover.js
```diff
@@ -34,34 +30,66 @@ const Popover = ({
...
+    const [popperElement, setPopperElement] = useState(null)
+    const [arrowElement, setArrowElement] = useState(null)
+    const modifiers = useMemo(
+        () =>
+            combineModifiers(arrow, arrowElement, {
+                observePopperResize,
+                observeReferenceResize,
+            }),
+        [arrow, arrowElement]
```
amcgee, line 50: This is missing `observePopperResize` and `observeReferenceResize` as dependencies, is that intentional?  I think this should also fail the `eslint` `react-hooks/exhaustive-deps` check, is that not running here for some reason?

src/Popover.js
HendrikThePendric: great find

src/Popover/modifiers.js
amcgee, line 23: Should this be `state.attributes.arrow = { ...state.attributes.arrow, ['data-arrow-hidden']: shouldHideArrow }` or will there never be any other attributes?

amcgee: Ah I see we're just using this mechanism ourselves, so it should always be only that property.

src/Popover.js
HendrikThePendric: It was simply an oversight.

src/Popover.js
HendrikThePendric: I'd like to start using this syntax but currently this project or my editor doesn't seem to like it. As discussed on Slack, let's fix this later in `dhis2/ui`.

src/Popper/modifiers.js
```diff
@@ -1,16 +1,63 @@
...
-    },
+import ResizeObserver from 'resize-observer-polyfill'
+
+const attachResizeObservers = ({
+    state: { elements },
+    options,
+    instance: { update },
+}) => {
+    const observers = Object.keys(options).reduce((acc, elementKey) => {
+        if (options[elementKey]) {
```
amcgee, line 9: Should this also check that `elements[elementKey]` exists?

amcgee:
> One thing that came up regularly during conversations about these components, is what the reference prop should actually be. People might be adding a DOM node directly, and in some cases that is actually the most convenient way to work. So I've added support for that in 7e5bf14

Makes sense, let's merge this as a `feat` though for this reason?

src/Popover/modifiers.js
HendrikThePendric: Yes, but still a very valid point. I've fixed this in https://github.com/dhis2/ui-core/pull/886/commits/ac0529302c54c51fb7c8c9274cf50534aa06baf1

src/Popper/modifiers.js
HendrikThePendric: No, the popper.js state object will always contain `elements.popper` and `elements.reference`.

HendrikThePendric:
> Makes sense, should we merge this as a feat for this reason?
I'll squash my commits and make sure we end up with a feat type commit

PR #887, opened 2020-05-19
feat: add support for virtual elements in popper and popover

HendrikThePendric:
This feature is needed to unblock some work that @edoardo is doing. I've included some stories that are demoing the functionality, but no e2e tests. I'd like to pick that up later, once this work has gone into `@dhis2/ui`.

### DIFF
--- a/src/Popover.js
+++ b/src/Popover.js
@@ -7,9 +7,10 @@ import { Layer } from './Layer.js'
 import { combineModifiers } from './Popover/modifiers.js'
 import { Arrow } from './Popover/Arrow.js'
 import {
-    elementRefPropType,
-    referencePlacementPropType,
+    popperReferencePropType,
+    popperPlacementPropType,
 } from './common-prop-types.js'
+import { getReferenceElement } from './Popper/getReferenceElement.js'
 
 /**
  * @module
@@ -35,10 +36,7 @@ const Popover = ({
     placement,
     onBackdropClick,
 }) => {
-    const referenceElement =
-        reference instanceof Element
-            ? reference
-            : reference && reference.current
+    const referenceElement = getReferenceElement(reference)
     const [popperElement, setPopperElement] = useState(null)
     const [arrowElement, setArrowElement] = useState(null)
     const modifiers = useMemo(
@@ -115,10 +113,6 @@ Popover.defaultProps = {
  */
 Popover.propTypes = {
     children: propTypes.node.isRequired,
-    reference: propTypes.oneOfType([
-        elementRefPropType,
-        propTypes.instanceOf(Element),
-    ]).isRequired,
     arrow: propTypes.bool,
     className: propTypes.string,
     dataTest: propTypes.string,
@@ -126,7 +120,8 @@ Popover.propTypes = {
     maxWidth: propTypes.number,
     observePopperResize: propTypes.bool,
     observeReferenceResize: propTypes.bool,
-    placement: referencePlacementPropType,
+    placement: popperPlacementPropType,
+    reference: popperReferencePropType,
     onBackdropClick: propTypes.func,
 }
 
--- a/src/Popover.js
+++ b/src/Popover.js
@@ -100,7 +100,7 @@ Popover.defaultProps = {
  * @typedef {Object} PropTypes
  * @static
  *
- * @prop {React.Ref|Element} reference A React ref or DOM element to the Popover should position against
+ * @prop {React.Ref|Element|VirtualElement} reference A React ref, DOM node, or {@link https://popper.js.org/docs/v2/virtual-elements/|popper.js virtual element} for the Popover to position itself against.
  * @prop {Node} children
  * @prop {boolean} [arrow=true] Show or hide the arrow
  * @prop {string} [className]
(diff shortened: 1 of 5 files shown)


ismay:
So if I understand correctly, the usecase here is to allow placement of a popper without a reference element?

src/Popover.js
```diff
@@ -115,18 +113,15 @@ Popover.defaultProps = {
  */
 Popover.propTypes = {
     children: propTypes.node.isRequired,
-    reference: propTypes.oneOfType([
-        elementRefPropType,
-        propTypes.instanceOf(Element),
-    ]).isRequired,
```
ismay, line 121: Should the reference no longer be required?

HendrikThePendric:
Yeah, it's nice to support this anyway for example if you want a popper to follow a mouse cursor or something.

The specific issue @edoardo encountered was as follows:
- When clicking in a table-cell, a popper with a menu should show, positioned against that table cell.
- However, on click the table re-renders, so the ref becomes null
- But before that re-render, we would still be able to access the DOM node and call `.getBoundingClientRect()` on it.
- This way we'd still be able to pass the right position to the Popper.

I guess this isn't something that will happen a lot, but nice to have support for it anyway.

I also quite like the fact that our Popper now has support for all types of references that are relevant:
- The objects created by `useRef` or `createRef`
- Plain DOM nodes
- Virtual elements (which is supported in popper.js)

src/Popper/getReferenceElement.js
```diff
@@ -0,0 +1,18 @@
+const getReferenceElement = reference => {
+    // Elements or virtualElements
```
ismay, line 2: Maybe a comment linking to: https://popper.js.org/docs/v2/virtual-elements would be nice here. I had no idea that this was an officially supported usecase and the term virtual element was new to me in this context.

src/Popover.js
HendrikThePendric: > Should the reference no longer be required? 

No, for the following reasons:
- Most importantly, because in the new implementation using `usePopper` from `react-popper`, this doesn't cause a runtime error anymore.
- But also because we are now also allowing DOM nodes directly, and it's a fairly likely scenario that these are undefined at some point in the render cycle

> By the way (sidenote), weren't we moving away from common prop-types?

I don't really know about that... In `@dhis2/ui-core` they are still there but in a separate package. In any case, the ones we had needed to be corrected/updated.

src/Popper/getReferenceElement.js
HendrikThePendric: Good idea, I'll update the jsDocs

ismay:
> Yeah, it's nice to support this anyway for example if you want a popper to follow a mouse cursor or something.
> 
> The specific issue @edoardo encountered was as follows:
> 
>     * When clicking in a table-cell, a popper with a menu should show, positioned against that table cell.
> 
>     * However, on click the table re-renders, so the ref becomes null
> 
>     * But before that re-render, we would still be able to access the DOM node and call `.getBoundingClientRect()` on it.
> 
>     * This way we'd still be able to pass the right position to the Popper.
> 
> 
> I guess this isn't something that will happen a lot, but nice to have support for it anyway.

Hmm yeah, it sounds like we need better support for attaching a popper in this case, using a virtual element seems like it could still go wrong if the table decides to render differently for some reason. Would there not be a more robust solution, something that utilizes the actual ref somehow?

HendrikThePendric:
> Hmm yeah, it sounds like we need better support for attaching a popper in this case, using a virtual element seems like it could still go wrong if the table decides to render differently for some reason and then the measurement would be wrong. Would there not be a more robust solution, something that utilizes the actual ref somehow?

I think I disagree with that, at least from the position of the ui-core library... The popper needs some sort of reference point to position against. It's up to the app to provide one. Passing a "virtual element" proved to be a good solution in this case, and for the popper to accept these as `reference` makes sense because the positioning engine supports that too. However, if the situation you were describing would occur, the table re-rendering into a different layout, then it's up to the app to determine the position of the popper.

ismay:
The code changes themselves look good to me. I do wonder about the usecase it's solving though. If this is a common thing we should have a more robust solution I think, unless that's just not possible (I don't know). I would also prefer to add some tests before merging, personally. It's not a huge change and I understand that there's time pressure, so I'll leave the decision to @HendrikThePendric as you know more than me about what @edoardo is working on and the deadlines involved.

ismay:
> I think I disagree with that, at least from the position of the ui-core library... The popper needs some sort of reference point to position against. It's up to the app to provide one. Passing a "virtual element" proved to be a good solution in this case, and for the popper to accept these as reference makes sense because the positioning engine supports that too. However, if the situation you were describing would occur, the table re-rendering into a different layout, then it's up to the app to determine the position of the popper.

Yeah the thing is, I can see this being used as a shortcut for these kinds of problems. And it seems like a shortcut that could break easily. If this is a problem that people are encountering I think we should investigate if it's possible to solve this in a robust way. Otherwise I can imagine this pattern would proliferate and that might not be a good thing.

HendrikThePendric:
> so I'll leave the decision to @HendrikThePendric as you know more than me about what @edoardo is working on and the deadlines involved.

It's still being discussed in the analytics-team, but it could well be that these two ui-core fixes (the current one and #888) will enable some of @edoardo work to make it into the 2.34.1 release, provided this is released today (or I guess first thing tomorrow).

I'd also like to add tests for this work, but would probably do so after we've pulled all of this into `@dhis2/ui`.

